PR #11, opened 2018-12-08
Add travis config

wjsi:
appveyor.yml not included temporarily, as we need to test Mars in Windows without plasma_store.

Altered cases:

* test_gevent_pool.py::Test::testRemoteConnections

  Under MacOS in Travis CI, the number of connections exceeds the system limitation.

* scheduler/test_main.py::Test::testMain

  Sometimes KVStoreActor is not ready when the worker is about to register.

* test_promise::Test::testAll

  Orders of latter ```gen_promise``` and prior ```thread_body``` may change randomly, which is consistent with the expectation.

Fixed cases:

* test_api::Test::testApi

  Web session does not handle timeout properly.

* test_operand.py::Test::testOperandActorWithCancel

  stop_graph sometimes get stuck due to sync operation.

Changed dependency:

* numpy >= 1.14, as pyarrow depends on the new API.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,12 +1,14 @@
 [run]
 branch = True
 concurrency = multiprocessing,gevent
-plugins = Cython.Coverage
-
-[report]
+include =
+    mars/*
 omit =
     mars/lib/functools32/*
     mars/lib/futures/*
-    mars/protos/*
     mars/lib/enum.py
     mars/lib/six.py
+    mars/lib/tblib/*
+    mars/lib/uhashring/*
+    mars/serialize/protos/*
+    */tests/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,45 @@
+language: python
+python:
+  - "2.7"
+  - "3.5"
+  - "3.6"
+
+matrix:
+  include:
+    - os: linux
+      python: "3.7"
+      dist: xenial
+      sudo: true
+    - os: osx
+      language: generic
+      env: PYTHON=2.7.12
+    - os: osx
+      language: generic
+      env: PYTHON=3.5.3
+    - os: osx
+      language: generic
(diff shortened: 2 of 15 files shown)


.coveragerc
```diff
@@ -1,7 +1,6 @@
 [run]
 branch = True
 concurrency = multiprocessing,gevent
-plugins = Cython.Coverage
```
chineking: Why remove this plugin?

mars/actors/pool/tests/test_gevent_pool.py
```diff
@@ -516,6 +518,8 @@ def testProcessUnpickled(self):
             with self.assertRaises(pickle.PickleError):
                 ref1.send(('create_unpickled',))
 
+    @unittest.skipIf(sys.platform == 'darwin' and 'TRAVIS' in os.environ,
+                     'Skip in Travis CI under MacOS due to open file number limitations')
```
chineking: Is it possible to modify the limitation of open file number for MacOS?

.coveragerc
wjsi: Cython.Coverage has a bug that blocks collecting coverage data from *.py (https://github.com/cython/cython/issues/1985)

mars/actors/pool/tests/test_gevent_pool.py
wjsi: I don't know where is the limitation.

wjsi: @chineking I've already tested that this problem can be simply solved by adding a ulimit -n 1024 in the preparation stage.

chineking: Great, so we don't have to skip this test, right?

wjsi: Yeah.

PR #14, opened 2018-12-10
Raise TypeError if argument is a shape-tuple in mt.random.rand & mt.random.randn

### LINKED ISSUE #1: [BUG] mt.random.rand got an unexpected result
example:
```
In [1]: import mars.tensor as mt

In [2]: r = mt.random.rand((2, 3))

In [3]: r.execute()
Out[3]: array([0.86173081, 0.0367415 ])
```
this interface should raise TypeError if argument is a shape-tuple, like numpy:
```
In [1]: import numpy as np

In [2]: np.random.rand((2, 3))
TypeError                                 Traceback (most recent call last)
<ipython-input-2-e49eb55bb286> in <module>()
----> 1 np.random.rand((2, 3))

mtrand.pyx in mtrand.RandomState.rand()

mtrand.pyx in mtrand.RandomState.random_sample()

mtrand.pyx in mtrand.cont0_array()

TypeError: 'tuple' object cannot be interpreted as an integer
```

hekaisheng:
raise TypeError if argument is a shape-tuple in mt.random.rand & mt.random.randn
fix #1

### DIFF
--- a/mars/tensor/expressions/random/rand.py
+++ b/mars/tensor/expressions/random/rand.py
@@ -69,7 +69,7 @@ def rand(random_state, *dn, **kw):
            [ 0.49313049,  0.94909878]]) #random
     """
     if len(dn) == 1 and isinstance(dn[0], (tuple, list)):
-        raise TypeError('this function cannot take tuple as argument')
+        raise TypeError("'tuple' object cannot be interpreted as an integer")
     if 'dtype' not in kw:
         kw['dtype'] = np.dtype('f8')
     chunks = kw.pop('chunks', None)
--- a/mars/tensor/expressions/random/randn.py
+++ b/mars/tensor/expressions/random/randn.py
@@ -82,7 +82,7 @@ def randn(random_state, *dn, **kw):
            [ 0.39924804,  4.68456316,  4.99394529,  4.84057254]]) #random
     """
     if len(dn) == 1 and isinstance(dn[0], (tuple,  list)):
-        raise TypeError('this function cannot take tuple as argument')
+        raise TypeError("'tuple' object cannot be interpreted as an integer")
     if 'dtype' not in kw:
         kw['dtype'] = np.dtype('f8')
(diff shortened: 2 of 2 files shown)


mars/tensor/expressions/random/rand.py
```diff
@@ -68,6 +68,8 @@ def rand(random_state, *dn, **kw):
            [ 0.37601032,  0.25528411],  #random
            [ 0.49313049,  0.94909878]]) #random
     """
+    if len(dn) == 1 and isinstance(dn[0], (tuple, list)):
+        raise TypeError('this function cannot take tuple as argument')
```
chineking: I have tried numpy, and I got

```
In [1]: import numpy as np

In [2]: np.random.rand((10, 20))
TypeError                                 Traceback (most recent call last)
<ipython-input-2-0fb309c754b2> in <module>()
----> 1 np.random.rand((10, 20))

mtrand.pyx in mtrand.RandomState.rand()

mtrand.pyx in mtrand.RandomState.random_sample()

mtrand.pyx in mtrand.cont0_array()

TypeError: 'tuple' object cannot be interpreted as an integer
```

What's the reason that we do not follow the exact error message?

mars/tensor/expressions/random/randn.py
```diff
@@ -81,6 +81,8 @@ def randn(random_state, *dn, **kw):
     array([[-4.49401501,  4.00950034, -1.81814867,  7.29718677],  #random
            [ 0.39924804,  4.68456316,  4.99394529,  4.84057254]]) #random
     """
+    if len(dn) == 1 and isinstance(dn[0], (tuple,  list)):
+        raise TypeError('this function cannot take tuple as argument')
```
chineking: The same question as the other one.

mars/tensor/expressions/random/rand.py
NJUPole: OK, i will change it to be consistent with numpy.

PR #24, opened 2018-12-11
Switch the graph into FAILED when preparation failed

### LINKED ISSUE #19: [BUG][DISTRIBUTED] client may hang if something wrong happens in the scheduler
**Describe the bug**

If some errors occur in the distributed, the client may hang, #8 and #17 are examples.

**To Reproduce**

As examples in #8 and #17.

**Expected behavior**

Either we deliver the error to the client, or we hide the real error and deliver an InternalServerError to the client.

**Additional context**

We should separate the system error with the user caused error. We should add some tests.

wjsi:
When there are any errors happened in GraphActor.execute_graph(), the state of the graph is switched to ``GraphState.FAILED`` which can be fetched by queries in the web. This may fix #19.


chineking:
Greate job

PR #26, opened 2018-12-12
add mars common API

NJUPole:
Add common API, it will be used by all user's interaction with cluster.

For example, web server hold a `MarsAPI` and all the handler get cluster information via it.

resolve #26


mars/api.py
```diff
@@ -0,0 +1,93 @@
...
+    def get_schedulers_info(self):
+        schedulers = self.cluster_info.get_schedulers()
+        infos = []
+        for scheduler in schedulers:
+            info_ref = actor_client.actor_ref(NodeInfoActor.default_name(),
+                                              address=scheduler)
+            infos.append(info_ref.get_info())
+        return infos
+
+    def list_workers(self):
```
wjsi: list_xxx should not return a number

mars/api.py
```diff
@@ -0,0 +1,93 @@
...
+from .scheduler.session import SessionManagerActor
+from .scheduler.graph import ResultReceiverActor
+
+logger = logging.getLogger(__name__)
+
+actor_client = new_client()
+
+
+class MarsAPI(object):
+    def __init__(self, scheduler_ip):
```
chineking: actor_client can be instantiated each time in the MarsAPI init.

mars/api.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def submit_graph(self, session_id, serialized_graph, graph_key, target):
+        session_uid = SessionActor.gen_name(session_id)
+        session_ref = self.get_actor_ref(session_uid)
+        session_ref.submit_tensor_graph(serialized_graph, graph_key, target, _tell=True)
+
+    def delete_graph(self, session_id, graph_key):
+        graph_uid = GraphActor.gen_name(session_id, graph_key)
+        graph_ref = self.get_actor_ref(graph_uid)
+        graph_ref.stop_graph()
```
chineking: stop graph in the delete seems unreasonable, how about add a stop_graph and do destroy in the delete_graph?

mars/session.py
```diff
@@ -43,10 +43,9 @@ class Session(object):
 
     def __init__(self, endpoint=None):
         if endpoint is not None:
-            from .web import get_client
+            from .web.session import Session
```
chineking: Here must be 3 kinds of sessions, besides the LocalSession, if the endpoint is a http/https url, the web session required, but for the LocalCluster, web is optional, if the endpoint is the scheduler address, the session should be based on the actor client api.

chineking: A LocalClusterSession can be added to do the job.

PR #31, opened 2018-12-14
Fix web, not patch thread to make web debuggable

NJUPole:
Web fix


mars/web/api.py
```diff
@@ -44,11 +45,12 @@ def _is_future(x):
 
 if six.PY2:
     _patch_futures()
-
+#
```
chineking: remove blank comment

PR #35, opened 2018-12-16
Add pseudo-cluster mode support

### LINKED ISSUE #16: Provide a way to start a complete distributed environment in a single machine

Now it's quite painful to start a complete distributed environment in a single machine, we have to start the scheduler, start the worker, then start the web.

I hope there is a simple way to do this by a single line of code. For example

```
from mars.deploy.local import new_cluster

cluster = new_cluster()
```

and then we can create a session by

```
from mars.session import new_session

session = new_session(cluster.endpoint)
```

Of course this is just an example.

I suggest we start the scheduler and worker in an actor pool, and start the web in a separate process.

**Additional context**

It's also useful for using in our unittests.

chineking:
resolve #16 

Add a way to start a distributed cluster in a single machine. This provides the convenient path to debug the distributed runtime on the laptop, and also it's quite easy to benchmark the distributed runtime without starting scheduler, worker and web through bash.

Usage:

Starting the local cluster:

```python
from mars.deploy.local import new_cluster

cluster = new_cluster()
```

By default the cluster will **not** contain the web ui, if users want to use web ui, they can specify `web=True` or `web=127.0.0.1:22334` to start a web ui.

For example

```python
from mars.deploy.local import new_cluster

cluster = new_cluster(web=True)
print(cluster.endpoint)  # the scheduler/worker address
print(cluster.web_endpoint)  # the web address
```

Start a local cluster will create a session and set the session as the default one, so when users call `execute` method, the graph will be submitted to the local cluster.

Users can also use the `new_session` to connect to the local cluster if they want to.


mars/deploy/local/tests/test_distributor.py
```diff
@@ -0,0 +1,41 @@
...
+from mars.worker.dispatcher import DispatchActor
+
+
+class Test(unittest.TestCase):
+    def testDistribute(self):
+        distributor = gen_distributor(2, 4)
+
+        idx = distributor.distribute(SessionActor.default_name())
+        self.assertLess(idx, 2)
+        self.assertGreaterEqual(idx, 1)
```
wjsi: why not assert idx == 1 ?

PR #36, opened 2018-12-17
add multiple scheduler support & process restart support

wjsi:
1. Add process-level restart mechanism
2. Allow running multiple schedulers
3. Allow relying on multiple etcds
4. Simplify in-process tests


wjsi:
After this PR is merged, #37 could be ready for merge.

PR #54, opened 2018-12-20
Support set endpoint for web session to reuse the session but with a different web

NJUPole:
## What do these changes do?

Add endpoint setter in web session so that it still works if web endpoint changed.

### DIFF
--- a/mars/session.py
+++ b/mars/session.py
@@ -14,14 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import uuid
-import json
-import time
 import numpy as np
-
-from .api import MarsAPI
-from .scheduler.graph import GraphState
-from .serialize import dataserializer
 
 
 class LocalSession(object):
@@ -37,6 +30,8 @@ class LocalSession(object):
 
     @endpoint.setter
     def endpoint(self, endpoint):
+        if endpoint is not None:
+            raise ValueError('Local session cannot set endpoint')
         self._endpoint = endpoint
 
     def run(self, *tensors, **kw):
@@ -103,6 +98,12 @@ class Session(object):
             return ret
         return ret[0]
 
+    def endpoint(self):
+        return self._sess.endpoint
+
+    def set_endpoint(self, endpoint):
+        self._sess.endpoint = endpoint
+
     def decref(self, *keys):
         if hasattr(self._sess, 'decref'):
             self._sess.decref(*keys)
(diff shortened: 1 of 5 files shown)


mars/web/tests/test_api.py
```diff
@@ -204,6 +205,21 @@ def setUp(self):
 
         self.start_web(scheduler_address)
 
+    def tearDown(self):
+        self._web.stop()
+        self._pool.stop()
+
+    def start_web(self, scheduler_address):
+        import gevent.monkey
```
chineking: The travis test test_api with process, so the monkey patch will not effect other tests, but in appveyor, we didn't run in process for this particular test. Please make sure this monkey patch will not effect other ut for now and the coming future.

mars/web/tests/test_api.py
```diff
@@ -204,6 +205,21 @@ def setUp(self):
...
+        gevent.monkey.patch_all(thread=False)
+
+        web_port = str(get_next_port())
+        mars_web = MarsWeb(port=int(web_port), scheduler_ip=scheduler_address)
+        mars_web.start()
```
chineking: `start` method supports `event`  parameter. I suggest to start the web by using the event, and do `event.wait(timeout=30` and then do timeout check in stead of keeping on loop which looks quite embarrassing.

mars/web/tests/test_api.py
```diff
@@ -252,3 +253,12 @@ def testApi(self, mock_fetch_tensor, _):
             data_url = graph_url + '/data/' + c.key
             data = loads(requests.get(data_url).content)
             assert_array_equal(data, np.ones((100, 100)) * 100)
+
+            # test web session endpoint setter
+            self._web.stop()
+            self.start_web(self._scheduler_address)
```
chineking: The same suggestion to use event, better to wrap the event and check logic to a class function to prevent from code copy.

mars/session.py
```diff
@@ -105,6 +105,15 @@ def __getattr__(self, attr):
         except AttributeError:
             raise
 
+    def __setattr__(self, key, value):
```
chineking: I suggest to make `endpoint` the common property of all kinds of sessions.

mars/web/tests/test_api.py
NJUPole: Now use mock server to test api, so the problem may be ignored.

mars/web/tests/test_api.py
NJUPole: the same as the above comment.

mars/web/tests/test_api.py
NJUPole: the same as the above comment.

PR #100, opened 2019-01-04
Check parameters for rand and randn

### LINKED ISSUE #91: [BUG][TENSOR] Check chunks parameter for mt.random.rand and mt.random.randn
**Describe the bug**

`chunks` parameter has been renamed to `chunk_size`, but for `mt.random.rand` and `mt.random.randn`, we did not check the parameter. If `chunks` is specified, no exception raised.

**Expected behavior**

Raise ValueError if `chunks` is specified for `mt.random.rand` and `mt.random.randn`

NJUPole:
## What do these changes do?

Check `mt.random.rand` and `mt.random.randn` parameters and raises `ValueError` if argument key not starts with '_' except `chunk_size`.

## Related issue number

fixes #91

### DIFF
--- a/mars/tensor/expressions/random/random_integers.py
+++ b/mars/tensor/expressions/random/random_integers.py
@@ -34,7 +34,7 @@ class TensorRandomIntegers(operands.RandomIntegers, TensorRandomOperandMixin):
         return self.new_tensor(None, None, raw_chunk_size=chunk_size)
 
 
-def random_integers(random_state, low, high=None, size=None, chunk_size=None, gpu=None, **kw):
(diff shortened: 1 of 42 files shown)


mars/tensor/expressions/random/random_integers.py
```diff
@@ -63,6 +63,8 @@ def random_integers(random_state, low, high=None, size=None, chunk_size=None, gp
         Desired chunk size on each dimension
     gpu : bool, optional
         Allocate the tensor on GPU if True, False as default
+    dtype : data-type, optional
```
chineking: No need to add docs here

PR #101, opened 2019-01-04
Fix setitem and getitem generate same operand key

### LINKED ISSUE #98: [BUG][TENSOR] Different slice on the same tensor generate same operand key
**Describe the bug**

Different slice on the same tensor generate same operand key which must be different.

**To Reproduce**

```python
In [22]: base_arr = np.random.random((4000, 2))                                 

In [23]: a = mt.array(base_arr)                                                 

In [24]: a[:400].op.key                                                         
Out[24]: '52efbfe8840294880d82da453025296b'

In [25]: a[400: 800].op.key                                                     
Out[25]: '52efbfe8840294880d82da453025296b'
```

NJUPole:
## What do these changes do?

Set `indexes` or `value` before generate operand key in `TensorIndexSetValue` and `TensorIndex` if they are provided.

## Related issue number

fixes #98

### DIFF
--- a/mars/deploy/local/tests/test_cluster.py
+++ b/mars/deploy/local/tests/test_cluster.py
@@ -197,13 +197,14 @@ class Test(unittest.TestCase):
                 np.testing.assert_array_equal(r, np.ones((5, 5)) * 2)
 
             with new_session(cluster.endpoint) as session3:
-                a = mt.ones((100, 5))
+                a = mt.random.rand(100, 5)
 
                 slice1 = a[:10]
                 slice2 = a[10:20]
-                r1, r2 = session3.run(slice1, slice2)
+                r1, r2, expected = session3.run(slice1, slice2, a)
 
-                np.testing.assert_array_equal(r1, r2)
+                np.testing.assert_array_equal(r1, expected[:10])
+                np.testing.assert_array_equal(r2, expected[10:20])
 
     def testExecutableTuple(self):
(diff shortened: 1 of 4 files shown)


mars/deploy/local/tests/test_cluster.py
```diff
@@ -196,6 +196,15 @@ def testIndexTensorExecute(self):
 
                 np.testing.assert_array_equal(r, np.ones((5, 5)) * 2)
 
+            with new_session(cluster.endpoint) as session3:
+                a = mt.ones((100, 5))
+
+                slice1 = a[:10]
+                slice2 = a[10:20]
+                r1, r2 = session3.run(slice1, slice2)
```
chineking: This test is meaningless, should test on random data, if test on ones, r1 will always be equal to r2 whatever the generated key is same or not.

mars/tensor/expressions/tests/test_indexing.py
```diff
@@ -215,3 +215,10 @@ def testNonzero(self):
         y = nonzero(x)
 
         self.assertEqual(len(y), 2)
+
+    def testOperandKey(self):
+        t = ones((10, 2), chunk_size=5)
+        t_slice1 = t[:5]
+        t_slice2 = t[5:]
+
+        self.assertNotEqual(t_slice1.op.key, t_slice2.op.key,)
```
chineking: Remove useless  comma.

Issue #106, opened 2019-01-06
Optimize preparation of huge graphs


wjsi:
**Is your feature request related to a problem? Please describe.**
When running huge graphs (~80k nodes) in Mars, runners have to wait for a long time (~20min) before the graph is tiled into chunks and actually start running. Therefore optimization is needed.

**Describe the solution you'd like**
All time-consuming steps and possible optimizations:

* [ ] graph tiling (~10min)
* [ ] initial placement (~6min)
* [ ] creating and distributing operands (~2min)

These steps can be optimized one by one.

chineking:
Can we give a isolated example to produce this process? So that we can dive into to profile and optimize.

wjsi:
Now after #107, the tiling time has been reduced to 1.5min.

PR #107, opened 2019-01-07
Use property instead of weak-key dict to store cix

wjsi:
## What do these changes do?

Use property instead of weak-key dict to store cix in order to accelerate. This avoids __eq__() calls in WeakKeyDictionary when seeking items and can reduce the time of tiling by at least 50%.

## Related issue number

N/A


chineking:
Is this about the first item in #106

PR #113, opened 2019-01-07
Support QR decomposition on short-and-fat tensor(sfqr method)

### LINKED ISSUE #114: [TENSOR] QR decomposition for Short-and-Fat matrix
`mt.linalg.qr` now only works for tall and skinny matrix, short and fat matrix is easy to implement in theory, something should be metioned is that 'sfqr' method cannot be used for `mt.linalg.svd`.

Short-and-Fat matrix A can be split like [A1, A2, A3, ...], we can choose A1 to be decomposed as a product A1 = Q1R1 of an orthogonal matrix Q1 and an upper triangular matrix R1, matrix Q1 as A's orthogonal matrix, then compute R ([R1, R2, R3, ...]) matrix by matrix Q1 to make sure A2=Q1R2, A3=Q1R3...

NJUPole:
## What do these changes do?

Add 'sfqr' method for `mt.linalg.qr`.

## Related issue number

resolve #114

### DIFF
--- a/mars/tensor/expressions/linalg/core.py
+++ b/mars/tensor/expressions/linalg/core.py
@@ -39,12 +39,18 @@ class SFQR(object):
         tinyq, tinyr = np.linalg.qr(np.ones((1, 1), dtype=a.dtype))
         q_dtype, r_dtype = tinyq.dtype, tinyr.dtype
 
+        need_rechunk = False
+        rechunk_size = dict()
         if a.chunk_shape[0] != 1:
-            new_chunks = decide_chunk_sizes(a.shape, {1: a.shape[0]}, a.dtype.itemsize)
-            a = a.rechunk(new_chunks).single_tiles()
+            need_rechunk = True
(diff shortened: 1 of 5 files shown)


chineking:
Better to open a issue first.

mars/tensor/expressions/linalg/core.py
```diff
@@ -21,7 +21,7 @@
 from ..core import TensorOperandMixin
 
 
-class TSQR(TensorOperandMixin):
+class QRBase(TensorOperandMixin):
```
wjsi: It is better to create separate classes for different QR methods and dispatch in ``qr(a, method='method')``.

NJUPole: `TensorQR` is a subclass of `QRBase` now, if create separate classes, each class may provide a staticmethod `tile` ?

chineking: `tile` is a classmethod, this is not a issue

PR #121, opened 2019-01-11
Add cython coverage

wjsi:
## What do these changes do?

Add coverage for *.pyx. Given that https://github.com/cython/cython/pull/2784 has not been merged by Cython, we use a temporarily solution by removing *.c under gevent path (fix-gevent-cov.sh). What's more, ``binding`` is configured to increase the coverage of ``cpdef``s, as is referred to in https://github.com/cython/cython/issues/1461. These shall be removed once Cython resolves the issues.

## Related issue number

N/A.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,6 +3,7 @@ branch = True
 cover_pylib = False
 concurrency = multiprocessing,gevent
 parallel = True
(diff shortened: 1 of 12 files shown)


wjsi:
fix-gevent-cov.sh removed in #159

PR #139, opened 2019-01-17
Fix missing import in install.rst

mmm311:
## What do these changes do?
Mistake in  install.rst:  "import mars.tensor as mt" should be added in local cluster example.


mmm311:
Thank you for your suggestion.

Issue #140, opened 2019-01-17
[BUG] Worker could not be started as described in the README


pingrunhuang:
**Describe the bug**
As described in the README.md file, worker process should be started properly after I started the `scheduler` and `web` process. However, after the `scheduler` and `web` are spawned which went well for both, the worker process fall into the following issue:

![screen shot 2019-01-17 at 4 14 46 pm](https://user-images.githubusercontent.com/13326067/51305020-a481a880-1a74-11e9-8bf0-7e78a742ae9b.png)

One thing to note is that I run it on a docker container as I want to separate the environment from the system environment. And I also strongly recommend adding Dockerfile and docker-compose.yml file to this project. I have already created these 2 files which is ready to be tested. If you want I can create a pull request.

1. Your Python version: `python 3.7.1`
2. The version of Mars you use: `'0.1.0b1'`
3. Versions of crucial packages, such as numpy, scipy and protobuf:
```
bokeh            1.0.4     
certifi          2018.11.29
chardet          3.0.4     
dnspython        1.16.0    
etcd-gevent      0.1.1     
gevent           1.4.0     
geventhttpclient 1.3.1     
greenlet         0.4.15    
idna             2.8       
Jinja2           2.10      
lz4              2.1.6     
MarkupSafe       1.1.0     
numexpr          2.6.9     
numpy            1.16.0    
packaging        18.0      
pandas           0.23.4    
Pillow           5.4.1     
pip              18.1      
protobuf         3.6.1     
psutil           5.4.8     
pyarrow          0.11.1    
pymars           0.1.0b1   
pyparsing        2.3.1     
python-dateutil  2.7.5     
pytz             2018.9    
PyYAML           3.13      
requests         2.21.0    
scipy            1.2.0     
setuptools       40.6.2    
six              1.12.0    
tornado          5.1.1     
urllib3          1.24.1    
wheel            0.32.2  
```
4. Full stack of the error.
5. Minimized code to reproduce the error.

**Expected behavior**
Start properly

pingrunhuang:
I imagine either the `--endpoint` should be specified or somehow I did it wrong.

NJUPole:
> I imagine either the `--endpoint` should be specified or somehow I did it wrong.

It's a bug, you can specified `--cache-mem 45%` as a workaround.

NJUPole:
[...]

PR #141, opened 2019-01-17
Fix worker start fail on machines with insufficient memory (also some doc fixes)

wjsi:
## What do these changes do?

Previously Mars workers are started on machines with sufficient memory, and the default value of --cache-mem, 48%, works well. When it is started on a machine with insufficient memory, however, the worker fails when doing memory test. Now we change the strategy by using half of free memory to ensure that it can start under most of the circumstances.

What's more, we add fixes on docs to add memory tuning section and errors on endpoints.

## Related issue number

#140 (Fix not added, just for verification)

### DIFF
--- a/mars/worker/__main__.py
+++ b/mars/worker/__main__.py
@@ -59,15 +59,19 @@ class WorkerApplication(BaseApplication, WorkerService):
 
         options.worker.cpu_process_count = int(self.args.cpu_procs or resource.cpu_count())
         options.worker.io_process_count = int(self.args.io_procs or '1')
+
+        if self.args.phy_mem:
+            mem_total = self._calc_size_limit(self.args.phy_mem, mem_stats.total)
+        else:
+            mem_total = mem_stats.total
+
         options.worker.physical_memory_limit_hard = self._calc_size_limit(
-            self.args.phy_mem or options.worker.physical_memory_limit_hard, mem_stats.total
+            options.worker.physical_memory_limit_hard, mem_total
         )
         options.worker.physical_memory_limit_soft = self._calc_size_limit(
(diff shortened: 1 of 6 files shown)


mars/worker/__main__.py
```diff
@@ -59,15 +59,19 @@ def create_pool(self, *args, **kwargs):
...
         options.worker.io_process_count = int(self.args.io_procs or '1')
+
+        if self.args.phy_mem:
+            mem_total = self._calc_size_limit(self.args.phy_mem, mem_stats.total)
+        else:
+            mem_total = mem_stats.total
+
         options.worker.physical_memory_limit_hard = self._calc_size_limit(
-            self.args.phy_mem or options.worker.physical_memory_limit_hard, mem_stats.total
+            options.worker.physical_memory_limit_hard, mem_total
```
NJUPole: those memory calculations may be move to `WorkerService`

chineking: > those memory calculations may be moved to `WorkerService`

Currently local cluster use 20% of shared memory.

chineking: This calculation is in main instead of `WorkerService.start` before.

chineking:
Seems break the local cluster, please try to fix this.

mars/worker/__main__.py
wjsi: Already moved and unified.

wjsi: Currently options should be set before actor pool starts. Thus these configurations CANNOT be moved inside start_service. Further refactoring are welcomed.

wjsi:
I completely remove codes dealing with service close as these codes lead to errors in local cluster cases, and closing problem for workers is not relevant with the topic of this pr.

PR #143, opened 2019-01-18
[BACKPORT]  Fix worker start fail on machines with insufficient memory, docs fixed also

wjsi:
## What do these changes do?

Back-port of #133, #136, #137, #139 and #141

## Related issue number

NA


wjsi:
Temporarily stop back-porting as new, potential doc fix may enter.

Issue #146, opened 2019-01-18
Add DOCKERFILE for building image


pingrunhuang:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I want to quick start the cluster in order to deploy my code to run on the cluster. However, I don't want to mess up the system dependencies and or that, I figure it is better to create a dockerfile for quick start.

**Describe the solution you'd like**
Based on the python:3.7.1 image, pip install the dependent lib into the container.

PR #148, opened 2019-01-18
Add Dockerfile

pingrunhuang:
## What do these changes do?

Add dockerfile for building up Mars project runnable environment

## Related issue number
#146


wjsi:
I wonder whether a Dockerfile like this is necessary, as any user with certain knowledge about docker can build one in a minute, and before you need to use it, you have to remember a lot of arguments in ``docker run``, which is still not convenient for common users. Maybe you need to looking at Dockerfiles written by other projects and rethink how docker can be used to facilitate Mars deployment. What's more, you also need to write documents about how users can use this file. Shall we make automated scripts to build images and upload to DockerHub, or we just add a script for people to build their own image themselves? Adding a Dockerfile is far from enough.

pingrunhuang:
@wjsi As I mentioned in my description, this Dockerfile is just a very basic one. More features should absolutely be added in the future. But this Dockerfile enable us as user to use docker-compose for starting a cluster in a more standard way since in the description of your README file differ from different environment. Therefore, I think it is necessary.

wjsi:
Your comments still cannot convince me as there are multiple ways to run Mars in separate environments, such as virtualenvs and conda envs. Installing in those environments are at least as easy as installing in a docker container.
> ...since in the description of your README file differ from different environment. Therefore, I think it is necessary
I wonder what is the true problem you are encountering, and can you share that with us and let us think how to solve it? As stated above, simply wrapping files in Dockerfile does not solve the problem, but make it more complicated.

pingrunhuang:
@wjsi yeah, sure. I installed it with virtualenv. And after that, running `mars-worker` require me to use `sudo` for higher privilege. Therefore I created one for myself and think might be also helpful for others. But it is totally understandable if you don't accept my solution. Personally I like the light way of creating a cluster for testing and development with docker swarm.

wjsi:
@pingrunhuang Can you open a new issue and paste the error log when you are not run mars-worker with ``sudo``? We can try to solve it.

chineking:
In my opinion, dockerfile is necessary, but how it should be needs further discussion.

wjsi:
Close this as we won't introduce premature changes hacking on fixed issues.

Issue #152, opened 2019-01-18
Win7 32-bit, when I install mars, it reports file not found


eagleLiu82:
Environment：`win7 32-bit+python3.7.0`
Execute：`pip install   pymars`
Error Message：
```
Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "C:\Users\ADMINI~1\AppData\Local\Temp\pip-install-h838j__8\pymars\set
up.py", line 42, in <module>
        with open(os.path.join(repo_root, 'requirements-extra.txt'), 'r') as f:
    FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\ADMINI~1
\\AppData\\Local\\Temp\\pip-install-h838j__8\\pymars\\requirements-extra.txt'

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in C:\Users\ADMINI~1
\AppData\Local\Temp\pip-install-h838j__8\pymars\
```

chineking:
Could you please translate your issue into English? Thanks in advance.

wjsi:
Due to limitations by ``pyarrow``, we do not support 32-bit Windows, please use 64-bit version instead. What's more, in Windows, only threaded mode or client mode is supported.

eagleLiu82:
> Due to limitations by `pyarrow`, we do not support 32-bit Windows, please use 64-bit version instead. What's more, only threaded mode or client mode is supported in Windows.

suse 10 64bit, when I use pip install pymars, it reports "Import Error:no module named Cython". I think this project should tell us what should prepare.

chineking:
Mars has pre-compiled wheel files listed at pypi: https://pypi.org/project/pymars/#files, if the Python version are one of 2.7, 3.5, 3.6 and 3.7, usually, `pip install pymars` will work.  In other situations, the source code will be fetched and  Cython would be necessary to do compilation work.

eagleLiu82:
> Mars has pre-compiled wheel files listed at pypi: https://pypi.org/project/pymars/#files, if the Python version are one of 2.7, 3.5, 3.6 and 3.7, usually, `pip install pymars` will work. In other situations, the source code will be fetched and Cython would be necessary to do compilation work.

suse 10 64bit,python3.6
when I use source code to install pymars, it reports "    with open(os.path.join(repo_root, 'requirements-extra.txt'), 'r') as f:
[...]

PR #153, opened 2019-01-18
Include requirements-*.txt, pyx and pxd, exclude c file in MANIFEST.in

chineking:
## What do these changes do?

Include requirements-extra.txt and requirements-dev.txt in MANIFEST.in. Global include pyx and pxd, exclude c file in MANIFEST.in.

## Related issue number

Fix file not found mentioned in #152

### DIFF
--- a/requirements-extra.txt
+++ b/requirements-extra.txt
@@ -5,6 +5,5 @@ psutil>=4.0.0
 lz4>=1.0.0
 protobuf>=3.6
 gevent>=1.3.0
-etcd-gevent>=0.1.0
 bokeh>=1.0.0
 jinja2>=2.0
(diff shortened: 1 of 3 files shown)


requirements-extra.txt
```diff
@@ -5,6 +5,5 @@ psutil>=4.0.0
 lz4>=1.0.0
 protobuf>=3.6
 gevent>=1.3.0
-etcd-gevent>=0.1.0
```
wjsi: Is there any error messages about etcd-gevent so we can fix?

chineking: I collect the feedback from users that the dependency like `geventhttpclient` is hard to install.

wjsi:
Maybe we need to give separate guidances both in docs and README on building from sources or prepare a development environment.

chineking:
> Maybe we need to give separate guidances both in docs and README on building from sources or preparing a development environment.

Agree, an issue can be created to mention this.

requirements-extra.txt
wjsi: Source packages are always hard to install. Okay, forget that.

wjsi:
etcd-gevent need to be installed in .travis.yml, as now it is removed from extras.

PR #169, opened 2019-01-22
Revert "Use miniconda to test on OSX"

chineking:
Reverts mars-project/mars#168


wjsi:
Cause already spotted and will fix in the next pr.

PR #180, opened 2019-01-24
Replace executed tensor with fetch tensor

### LINKED ISSUE #172: Replace executed tensor by tensor with fetch operand

As a step of eager mode #28 , all the executed tensor should be replaced by tensor whose op is `Fetch`, the logic is different for local executor and distributed cluster.

- Change the logic of `tensor.build_graph`. For local, `build_graph` generates a chunk graph, we can replace executed chunk by fetch chunk. When submit to cluster, it will build tensor graph, we should use fetch tensor to replace executed tensor. At the same time, the predecessors of executed one will not in generated graph.

- Handle the tile logic of fetch tensor. tiling happens in `GraphActor.prepare_graph`, we should get the fetch tensor's chunk meta from the executed `GraphActor`.

- Remove fetch chunk from chunk graph after `prepare_graph`. As fetch chunk has been executed before, `OperandActor` for fetch should not be created, we just ensure that the fetch key is the same as executed chunk's.

**Additional context**

As eager mode will trigger fetch frequently, we must be care of the logic of `decref`.

NJUPole:
## What do these changes do?

As described in #172 , use fetch tensor/chunk to replace executed tensor/chunk when submit graph.

Change the logic of `decref` for local session, local cluster session and web session.

## Related issue number

resolve #172


mars/tensor/expressions/utils.py
```diff
@@ -472,3 +472,26 @@ def calc_rough_shape(tensor):
...
+    if isinstance(tensor, CHUNK_TYPE):
+        new_op = TensorFetch(dtype=tensor.dtype)
+        return new_op.new_chunk(None, tensor.shape, index=tensor.index,
+                                _key=tensor.key, _id=tensor.id)
+    elif isinstance(tensor, TENSOR_TYPE):
+        new_op = TensorFetch(dtype=tensor.dtype)
+        if tensor.is_coarse():
+            return new_op.new_tensor(None, tensor.shape, _key=tensor.key, _id=tensor.id)
+        else:
+            chunks = []
```
wjsi: Branch not covered by ci

mars/session.py
```diff
@@ -101,58 +105,51 @@ def run(self, *tensors, **kw):
...
             return ret[0]
 
-    def fetch(self, tensor):
-        if tensor.key not in self._executed_keys:
-            raise ValueError('Cannot fetch the unexecuted tensor')
-        return self._sess.fetch(tensor)
+    def fetch(self, *tensors, **kw):
+        ret_list = False
+        if len(tensors) == 1 and isinstance(tensors[0], (tuple, list)):
+            ret_list = True
```
wjsi: Not covered

mars/session.py
```diff
@@ -101,58 +105,51 @@ def run(self, *tensors, **kw):
...
+            tensors = tensors[0]
+        elif len(tensors) > 1:
+            ret_list = True
+
+        result = self._sess.fetch(*tensors, **kw)
+
+        ret = []
+        for r, t in zip(result, tensors):
+            if r is None:
+                ret.append(r)
```
wjsi: Not covered

NJUPole: Just copy the logic from `run`, does any tensor's return value is `None`?

NJUPole: It should be removed as `run` will return None when specify `fetch=False` which won't happened in `fetch`

mars/web/tests/test_api.py
```diff
@@ -161,10 +160,6 @@ def testApi(self):
             value2 = sess.run(c)
             assert_array_equal(value, value2)
 
-            # todo this behavior may change when eager mode is introduced
-            with self.assertRaises(ExecutionFailed):
```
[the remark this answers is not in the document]
NJUPole: Now tested in `test_cluster`

mars/scheduler/graph.py
```diff
@@ -799,6 +813,36 @@ def build_tensor_merge_graph(self, tensor_key):
 
         return serialize_graph(graph)
 
+    def tile_fetch_tensor(self, tensor):
```
chineking: We can let the tile done by the GraphActor which owned the original tensor, it would be much simpler.

mars/scheduler/session.py
```diff
@@ -42,6 +43,9 @@ def get_argument(self, key):
     def get_graph_refs(self):
         return self._graph_refs
 
+    def get_graph_ref_by_tensor(self, tensor_key):
```
chineking: Better with name `get_graph_ref_by_tensor_key`.

mars/session.py
```diff
@@ -101,58 +105,48 @@ def run(self, *tensors, **kw):
...
-                for j, result in zip(idx_to_run_tensors, execute_result):
-                    results[j] = result
-            run_tensors = list(idx_to_run_tensors.values())
-            self._executed_keys.update(t.key for t in run_tensors)
-            for t in run_tensors:
-                t._execute_session = self
+        result = self._sess.run(*tensors, **kw)
+
+        for t in tensors:
+            t._execute_session = self
```
chineking: Shall we consider that a tensor could be executed by different sessions?

mars/web/session.py
```diff
@@ -60,7 +60,13 @@ def _main(self):
         self._session_id = content['session_id']
 
     def _get_graph_key(self, tensor_key):
-        return self._tensor_to_graph[tensor_key]
+        return self._executed_tensors[tensor_key][0]
+
+    def _set_graph_key(self, tensor_key, tid, graph_key):
```
chineking: The same problem as I mentioned before.

mars/deploy/local/session.py
```diff
@@ -45,7 +45,13 @@ def endpoint(self, endpoint):
         self._api = MarsAPI(self._endpoint)
 
     def _get_graph_key(self, tensor_key):
```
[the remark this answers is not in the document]
NJUPole: Since `DataFrame` is introduced, most function or variable name contains 'tensor' may convert to 'entity' ?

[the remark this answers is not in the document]
chineking: Yeah, let we do this latter, it's a big change beyond this PR though, suggest to only mark a todo for now.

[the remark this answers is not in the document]
chineking: This place could be `tensor` for now.

mars/deploy/local/tests/test_cluster.py
```diff
@@ -293,3 +294,81 @@ def testGraphFail(self):
...
+    def testFetch(self):
+        with new_cluster(scheduler_n_process=2, worker_n_process=2, web=True) as cluster:
+            session = cluster.session
+
+            a1 = mt.ones((10, 20), chunk_size=8) + 1
+            r1 = session.run(a1)
+            r2 = session.fetch(a1)
+            np.testing.assert_array_equal(r1, r2)
+
+            r3 = session.run(a1 * 2)
```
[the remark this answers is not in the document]
NJUPole: Should we add ut for `build_graph` to check fetch replacement?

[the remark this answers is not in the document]
chineking: This should be necessary. And also session should have some sort of generating graph API, test on this.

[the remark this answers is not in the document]
NJUPole: `build_graph` is now a util function and test separately.

mars/session.py
NJUPole: Yeah, and do decref for all sessions if the tensor is garbage collected

mars/scheduler/graph.py
```diff
@@ -813,35 +814,42 @@ def build_tensor_merge_graph(self, tensor_key):
...
+        """
+        Convert single tensor to graph which contains one tensor node and multiple chunk nodes
+        :param tensor_key: the key of tensor
+        """
+        tiled_tensor = self._get_tensor_by_key(tensor_key)
+        graph = DAG()
+        for c in tiled_tensor.chunks:
+            fetch_op = TensorFetch(dtype=c.dtype)
+            fetch_chunk = fetch_op.new_chunk(None, c.shape, c.index, _key=c.key).data
+            graph.add_node(fetch_chunk)
```
chineking: Wait for #191 finished, don't need to add the chunk any more, just add the tensor is fine.

mars/scheduler/graph.py
```diff
@@ -813,35 +814,42 @@ def build_tensor_merge_graph(self, tensor_key):
...
-
-            # calculate nsplits
-            for i in range(ndim):
-                if all(idx == 0 for j, idx in enumerate(chunk_index) if j != i):
-                    tensor_nsplits[i].append(chunk_shape[i])
-
-        nsplits = tuple(tuple(n) for n in tensor_nsplits)
-        new_op = TensorFetch(dtype=tensor.dtype)
-        return new_op.new_tensor(None, tensor.shape, chunks=chunks, nsplits=nsplits, _key=tensor.key)
+        tensor_node = None
```
chineking: Ditto, wait for #191 completed, just get the sole tensor in the graph.

chineking:
LGTM, @wjsi please check if there is any further suggestion.

PR #193, opened 2019-01-30
Separate graph analyzer from graph and add test cases

### LINKED ISSUE #196: Separate graph analyzer from graph
Initial graph assignment will be too lengthy when incremental assignments or other improvements are added. It is better to separate it to an independent module for maintainability.

wjsi:
## What do these changes do?

Separate graph analytics from GraphActor into a separate GraphAnalyzer (this is not an actor). This facilitates further optimization and also paves way for worker-level failover (#5).

## Related issue number

Fixes #196

### DIFF
--- a/mars/scheduler/operand.py
+++ b/mars/scheduler/operand.py
@@ -551,12 +551,7 @@ class OperandActor(SchedulerActor):
                 self.free_data(OperandState.CANCELLED)
                 return
 
-            if exc and issubclass(exc[0], ResourceInsufficient):
-                # resource insufficient: just set to READY and continue
-                self.worker = None
-                self.state = OperandState.READY
-                self.ref().start_operand(_tell=True)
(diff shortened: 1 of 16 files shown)


mars/scheduler/operand.py
```diff
@@ -551,12 +551,7 @@ def _rejecter(*exc):
                 self.free_data(OperandState.CANCELLED)
                 return
 
-            if exc and issubclass(exc[0], ResourceInsufficient):
```
chineking: Why remove this?

wjsi: Never raised in worker.

wjsi:
New code starts from https://github.com/mars-project/mars/pull/193/files#diff-601556fda73a5008c900c8a9e5d0997eR88. Initial assigning is rewritten to support assigning nodes in an adaptive manner.

PR #198, opened 2019-01-31
Use fixed versions of dependencies to build wheels

### LINKED ISSUE #197: [BUG] Version of numpy should be fixed when building wheels
**Describe the bug**
As numpy may [change their API version](https://github.com/numpy/numpy/blob/master/doc/release/1.16.0-notes.rst#c-api-changes), we should fix the version of numpy as the version used by pyarrow.

wjsi:
## What do these changes do?

Use fixed versions of dependencies to build wheels to reduce potential errors caused by upgraded versions of dependencies.

Tag build see https://travis-ci.org/wjsi/mars/builds/486882560, test packages at testpypi.

## Related issue number

Fixes #197

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ install:
   - pip install -r requirements-dev.txt &&
     pip install -r requirements-extra.txt &&
     pip install virtualenv coveralls flake8 etcd-gevent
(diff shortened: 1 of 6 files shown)


chineking:
@NJUPole please review this for double check.

PR #200, opened 2019-02-02
Support eager mode

### LINKED ISSUE #194: Trigger execution once tensor created when eager mode is on
This is the last step of eager mode(#28 ).

When eager mode is on, tensor will be executed by default session once it is created by operator, user can access the tensor results via `tensor.fetch` or check them by using `print`.

As mentioned in #28 , `new_tensor` is also used in the `tiles`, we can use a decorator to wrap operand's `__call__` function and trigger execution. For example:
```Python


def trigger_execution(func):

    def _wrapped(*args, **kwargs):
        expression = func(*args, **kwargs)
        if options.eager_mode:
            expression.execute(fetch=False)
        return expression

    return _wrapped


class TensorBinOp():
    @trigger_execution
    def __call__(self, x1, x2, out=None, where=None):
        return self._call(x1, x2, out=out, where=where)

```

Trigger the execution in the `new_tensor` and decide to whether do execution by incoming parameter like `executable`, `new_tensor` called in `tiles` should set `executable=False`.

**Additional context**
Tests should cover all the operands to ensure their correct behavior in eager mode.

NJUPole:
## What do these changes do?

Support eager mode for `mars.tensor`. Trigger execution after `new_tensors` when create tensors.

## Related issue number

resolve #194

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -86,57 +86,37 @@ Any contribution from community is sincerely welcomed. The main feature not impl
 Eager Mode
 ```````````
 
-.. Note:: New in version 0.2.0a2
-
 Mars supports eager mode which makes it friendly for developing and easy to debug.
 
 Users can enable the eager mode by options, set options at the beginning of the program or console session.
 
 .. code-block:: python
 
-    from mars.config import options
-
-    options.eager_mode = True
+    >>> from mars.config import options
+    >>> options.eager_mode = True
 
 Or use a context.
 
 .. code-block:: python
 
-    from mars.config import option_context
-
-    with option_context() as options:
-        options.eager_mode = True
-        # the eager mode is on only for the with statement
+    >>> from mars.config import option_context
+    >>> with option_context() as options:
+    >>>     options.eager_mode = True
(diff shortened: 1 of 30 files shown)


chineking:
Briefly scanned the code, a few suggestions.

1. Add `_new_entities` and `_new_chunks` which is for the subclass to inherit, `new_entities` would become a final method.
2. Entering and leaving kernel mode once entering and leaving the Executor, tiles and so forth which belongs to the kernel stuff.

This would make the code more neater  in my opinion.

README.rst
```diff
@@ -83,6 +83,61 @@ Any contribution from community is sincerely welcomed. The main feature not impl
...
+    t = mt.arange(6).reshape((2, 3))
+    print(t)
+
+.. code-block:: python
+
+    Out: Tensor(op=TensorRand, shape=(4, 4), data=
+         [[0 1 2]
+          [3 4 5]])
+
+Use ``fetch`` to obtain numpy value from a tensor:
```
chineking: `fetch` couldn't be mentioned here.

PR #226, opened 2019-02-22
Use random object id in plasma store to avoid potential collision

### LINKED ISSUE #224: Use random keys in plasma_store to avoid collision
Currently in Mars, object keys in plasma store and chunk keys are closely binded. As plasma store does not provide explicit method to delete some keys from the store, it is hard for us to remove chunks not fully transferred.

Use a worker-wide kv store to save key -> object id mappings and store keys in plasma.

wjsi:
## What do these changes do?

Use random object id in plasma store to avoid potential collision. A worker-wide kv store maintains mappings between chunk keys and plasma object ids.

## Related issue number

Fixes #224 #233

### DIFF
--- a/mars/scheduler/tests/test_kvstore.py
+++ b/mars/scheduler/tests/test_kvstore.py
@@ -24,6 +24,10 @@ from mars.utils import get_next_port
 
 
 class Test(unittest.TestCase):
+    def tearDown(self):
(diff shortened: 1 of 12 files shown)


chineking:
For #233 , we can backport a PR individually.

PR #237, opened 2019-02-26
Add TensorInv operand to fix bug in mt.linalg.inv

### LINKED ISSUE #230: [BUG] Failed to calculate inverse matrix
**Describe the bug**
When input matrix is not a `TensorDataSource` type, `mt.linalg.inv` will raise `AttributeError` which is caused by `mt.eye` called in `inv`.

**To Reproduce**

``` Python
In [1]: import mars.tensor as mt                                                                                                                        

In [2]: a = mt.random.randint(1, 10, (6, 6), chunk_size=3)                                                                                              

In [3]: b = a.dot(a)                                                                                                                                    

In [4]: r = mt.linalg.inv(b)     
                                                                                                                       
Traceback (most recent call last):
  File "/Users/hekaisheng/Documents/mars_dev/mars/mars/utils.py", line 65, in __getattr__
    return self[item]
KeyError: 'raw_chunk_size'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/IPython/core/interactiveshell.py", line 3267, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-7-7491bd48bc7b>", line 1, in <module>
    mt.linalg.inv(a.dot(a)).execute()
  File "/Users/hekaisheng/Documents/mars_dev/mars/mars/tensor/expressions/linalg/inv.py", line 61, in inv
    return solve(a, eye(a.shape[0], chunk_size=a.params.raw_chunk_size))
  File "/Users/hekaisheng/Documents/mars_dev/mars/mars/utils.py", line 68, in __getattr__
    "'AttributeDict' object has no attribute {0}".format(item))
AttributeError: 'AttributeDict' object has no attribute raw_chunk_size
```

NJUPole:
## What do these changes do?

Add operand `TensorInv` and generate identity matrix in `tile` so that we don't need to pass `raw_chunk_size` to `mt.eye`. `TensorInv` is only available in coarse-grained graph, it will convert to combination of other operands in `tiles`.

## Related issue number

resolve #230


mars/tensor/expressions/linalg/inv.py
```diff
@@ -31,15 +87,6 @@ def inv(a):
     -------
     ainv : (..., M, M) ndarray or matrix
         (Multiplicative) inverse of the matrix `a`.
-    Raises
```
chineking: Can keep the raises.

PR #254, opened 2019-03-02
Support mt.diag on sparse 2d tensor

### LINKED ISSUE #252: Support execute mt.diag given a sparse input tensor

Currently, `mt.diag` can accept a sparse tensor, but the execution cannot be performed because the internal sparse lib of Mars does not support this.

According to the behavior of `scipy.sparse.csr_matrix.diagonal`, we can just return a dense tensor。

chineking:
## What do these changes do?

Support `mt.diag` on sparse 2d tensor.

## Related issue number

Resolves #252

### DIFF
--- a/mars/lib/sparse/__init__.py
+++ b/mars/lib/sparse/__init__.py
@@ -706,10 +706,7 @@ def ones_like(x):
 
 
 def diag(v, k=0, gpu=False):
-    if v.ndim == 2:
-        raise NotImplementedError
-
-    assert v.ndim == 1
+    assert v.ndim in {1, 2}
 
     from .matrix import diag_sparse_matrix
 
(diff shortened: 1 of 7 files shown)


mars/lib/sparse/__init__.py
```diff
@@ -706,10 +706,7 @@ def ones_like(x):
 
 
 def diag(v, k=0, gpu=False):
-    if v.ndim == 2:
-        raise NotImplementedError
-
-    assert v.ndim == 1
+    assert v.ndim in {1, 2}
```
wjsi: Raise explicitly instead of using assertions

chineking:
The decrease of coveralls is not relevant to this PR.

PR #281, opened 2019-03-12
Add more translation

ChenQuan:
## What do these changes do?

The translation is as follows:
- datasouce.po
- execution.po
- overview.po
- routines.po
- sparse.po
- unfunc.po

## Related issue number
Not.


docs/source/locale/zh_CN/LC_MESSAGES/tensor/ufunc.po
```diff
@@ -140,7 +141,7 @@ msgstr ""
 
 #: ../../<autosummary>:1
 msgid "Return element-wise remainder of division."
-msgstr ""
+msgstr "返回按元素排序的除法余数。"
```
chineking: '返回按元素排序的除法余数。' -> '返回逐元素的除法余数'

ChenQuan:
@chineking Sorry,I  am not particularly skilled at using Github. I also changed the error, do I need to close the current PR? Then re-PR?

ChenQuan:
> Thanks for your work on translation, small tip is that, after translation, call `make gettext` which according to the [building documents part in contributing](http://docs.mars-project.io/en/latest/contributing.html#building-documentations).
Yes,I hava done it.

chineking:
> @chineking Sorry,I am not particularly skilled at using Github. I also changed the error, do I need to close the current PR? Then re-PR?

Since you didn't modify the generated directory, just reset the docs/source/tensor/generated to older version, and keep the other modification.

chineking:
I noticed that seems the `make gettext` cause huge modification, I suggest you just reset to the first commit, and apply the new modification to keep the minimized change.

ChenQuan:
image
I reset to #279, but I still have to pass the test automatically.

ChenQuan:
I think I need remove the current forks,then translate it ,and try PR again. @chineking

PR #292, opened 2019-03-13
Refactor unittests for tensor expressions to ensure full tests on sparse and serialization

### LINKED ISSUE #71: Test tensor execution in distributed mode
Currently, our expressions test cases all run by thread-based executor, however the behavior changes when running in cluster, only a few cases in `test_api` and `test_main` works for end to end.

Several issues(#19  #64  #8 ) are related to this, maybe we need some tests to cover them in the future.

At least, we can add cases in `testGraphActor` to test `GraphActor` with different input tensors.

### LINKED ISSUE #267: Check all operands about sparse, add unittests for both expressions and executions

I'd like to conduct a full scan of current code to make sure the `sparse` property is handled by each operand. Unittests for both expressions and executions could be added either.

This work should include fix about #263 and #264 .

chineking:
## What do these changes do?

I decided to do a full scan on our tensor expressions, so that each operand handles sparse, serialization and so forth correctly.

Also, I'd like to write each operand's ut separately, and make sure each operand for tensor is fully tests for sparse, serialization etc.

## Related issue number

Fixes #267 
Fixes #71


chineking:
Since #299, we decide to unify the sparse behavior, for instance, sparse tensor add 1 will only apply the add to the part with data. So I guess the separated tests on each operands is not necessary any more. We can parametrize the tests on all operands.

PR #330, opened 2019-04-04
[WIP] Try fix actor case failure on windows

### LINKED ISSUE #45: test_gevent_pool should be fixed in Windows
Now, we skip all the cases in `actors.pool.test_gevent_pool` in Windows, it may be used in future, so all gevent test cases should pass.

wjsi:
## What do these changes do?

Try fix gevent cases in Windows

## Related issue number

Fixes #45


wjsi:
As we decide to move into asyncio, this PR shall be deprecated.

PR #228, opened 2019-02-22
Fix create existing operand actor

NJUPole:
## What do these changes do?

Add `append_graph` method in `OperandActor`.

If the existing operand actor's data is stored, add the new successors and predecessors to this operand.

If data is freed or the operand is in other state, reset the state of operand and start running again.

## Related issue number

resolve #179

PR #345, opened 2019-04-15
[BACKPORT] Fix create existing operand actor

NJUPole:
This PR is a backport of #228 and part of #301 . 

Because of the refactor of operand actor and graph actor, the code and logic may be different from the original commit.

### DIFF
--- a/mars/scheduler/graph.py
+++ b/mars/scheduler/graph.py
@@ -658,10 +658,10 @@ class GraphActor(SchedulerActor):
             op_info['io_meta'] = io_meta
 
             if io_meta['predecessors']:
-                state = 'UNSCHEDULED'
+                state = OperandState.UNSCHEDULED
             else:
                 initial_keys.append(op_key)
-                state = 'READY'
+                state = OperandState.READY
             op_info['retries'] = 0
             op_info['state'] = state
 
@@ -673,7 +673,6 @@ class GraphActor(SchedulerActor):
                 uid=op_uid, address=scheduler_addr,
                 wait=False
             )
-            op_info['state'] = getattr(OperandState, state.upper())
             if _clean_io_meta:
                 del op_info['io_meta']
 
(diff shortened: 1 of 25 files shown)


mars/scheduler/graph.py
```diff
@@ -625,33 +653,11 @@ def create_operand_actors(self, _clean_io_meta=True):
...
-                shared_input_chunks=list(shared_input_chunk_keys),
-                chunks=list(chunks),
-            )
+            io_meta = self._collect_operand_io_meta(chunk_graph, self._op_key_to_chunk[op_key])
             op_info['op_name'] = op_name
             op_info['io_meta'] = io_meta
 
-            if predecessor_keys:
+            if io_meta['predecessors']:
                 state = 'UNSCHEDULED'
```
wjsi: This can be changed into real enum value instead of names and values to reduce code complexity.

NJUPole: Done.

mars/scheduler/operand.py
```diff
@@ -118,6 +119,38 @@ def add_finished_predecessor(self, op_key):
...
+            self._pred_keys.update(op_info['io_meta']['predecessors'])
+            self._succ_keys.update(op_info['io_meta']['successors'])
+
+        futures = []
+        for pred_key in op_info['io_meta']['predecessors']:
+            futures.append(self._get_operand_actor(pred_key).remove_finished_succs(
+                self._op_key, _tell=True, _wait=False))
+        for succ_key in op_info['io_meta']['successors']:
+            futures.append(self._get_operand_actor(succ_key).remove_finished_preds(
+                self._op_key, _tell=True, _wait=False))
```
wjsi: It is better to use a batch interface

mars/scheduler/operand.py
```diff
@@ -118,6 +119,38 @@ def add_finished_predecessor(self, op_key):
...
+        [future.result() for future in futures]
+
+    def remove_finished_succs(self, op_key):
```
wjsi: Comments needed to mark these as workarounds.

Issue #134, opened 2019-01-16
Add shuffle support


wjsi:
**Is your feature request related to a problem? Please describe.**
Currently there are no shuffle support in Mars, therefore a lot of functions cannot be added, such as fancy indexing.

**Describe the solution you'd like**
Add shuffle support for Mars. To reduce potential burden on graph processing, a proxy node is added and edge count is reduced from a * b to a + b.

![image](https://user-images.githubusercontent.com/8284922/51223626-7a998a80-197d-11e9-95f6-12f09970995d.png)

Shuffle output operands inherit a base class called ``ShuffleMap`` and input operands ``ShuffleReduce``. ``ShuffleMap`` exports data parts into separate parts by keys. ``ShuffleReduce`` accepts data parts by key and calculate combinations.

Local execution, scheduler and worker needs updates to accept such kind of graph and data organization.

Shuffle will first be applied on irregular reshapes.

wjsi:
The beginning steps can be divided into two steps:

* [ ] Implement shuffle in threaded mode to support reshape.
* [ ] Implement shuffle in distributed mode.

wjsi:
A preliminary version is ready, see https://github.com/mars-project/mars/compare/master...wjsi:feature/shuffle for diff. PR date not decided.

wjsi:
This will be implemented in two steps:

[ ] Add shuffle operands for reshape and capability to run in local mode, and
[ ] Add shuffle support in distributed environment

PR #351, opened 2019-04-19
Experimental `add` support for DataFrame

### LINKED ISSUE #177: Implement DataFrame's arithmetic interface

We should make sure that Mars DataFrame can do arithmetic e.g. add, minus etc with scalar, Mars(pandas) Series and DataFrame,  and tensor(numpy ndarray).

Note that pandas will union the indices, thus this may trigger a shuffle operation when the index is not RangeIndex. We can implement the DataFrame with RangeIndex first, and consider shuffle situation when #134 is ready.

chineking:
## What do these changes do?

This PR aims to supply experimental `add` support for DataFrame. 

The key point is that when adding two dataframes, pandas will automatically align indices and columns. In order to address this problem for Mars under the parallel or distributed setting, we need to try our best to avoid shuffle which introduced in #134 and has high cost for computation.

In current implementation, if columns and index of both sides are monotonic increasing or decreasing, we could avoid shuffle.

Some further improvements could be considered:

1. The DataFrame's and chunk's index and columns are absent for now.
2. If index or columns have only 1 chunk, shuffle could be avoided either. 
3. GPU support.

## Related issue number

Resolves #177 .


mars/dataframe/expressions/utils.py
```diff
@@ -84,22 +91,192 @@ def decide_chunk_sizes(shape, chunk_size, memory_usage):
     return tuple(row_chunk_size), tuple(col_chunk_size)
 
 
-def parse_index(index_value):
+def parse_index(index_value, store_data=False):
     import pandas as pd
 
+    def _parse_property(index, ret_data):
```
wjsi: parse or extract?

mars/dataframe/expressions/arithmetic/core.py
```diff
@@ -0,0 +1,358 @@
...
+            left_chunk_index_min_max = cls._get_chunk_index_min_max(left, index_type, axis)
+            right_chunk_index_min_max = cls._get_chunk_index_min_max(right, index_type, axis)
+            if left_chunk_index_min_max is not None and right_chunk_index_min_max is not None:
+                # no need to do shuffle on this axis
+                left_splits, right_splits = split_monotonic_index_min_max(
+                    *(left_chunk_index_min_max + right_chunk_index_min_max))
+                left_increase = left_chunk_index_min_max[1]
+                right_increase = right_chunk_index_min_max[1]
+                splits[axis] = (left_splits, build_split_idx_to_origin_idx(left_splits, left_increase)), \
+                               (right_splits, build_split_idx_to_origin_idx(right_splits, right_increase))
```
wjsi: The tuple on the right along with the dict introduces three ``__getitem__``s for actual tiling functions, which introduces great difficulty for code readers when considering left and right ``splits`` are three-leveled tuples. Therefore I suggest using a namedtuple to pass these values to make this argument more clear.

mars/dataframe/expressions/arithmetic/core.py
```diff
@@ -0,0 +1,358 @@
...
+        l_filter = operator.ge if column_min_max[1] else operator.gt
+        l = l_filter(dtypes.index, column_min_max[0])
+        r_filter = operator.le if column_min_max[3] else operator.lt
+        r = r_filter(dtypes.index, column_min_max[2])
+        f = l & r
+        return dtypes[f]
+
+    @classmethod
+    def _gen_out_chunks_without_shuffle(cls, op, splits, out_shape, left, right):
+        out_chunks = []
```
wjsi: Please assign splits[axis][left_or_right][splits_or_back_mappings] explicitly to variables before referencing them inside the loop. splits[a][b][c][d] makes it hard for people to understand.

mars/dataframe/expressions/arithmetic/core.py
```diff
@@ -0,0 +1,358 @@
...
+        for out_idx in itertools.product(*(range(s) for s in out_shape)):
+            if splits[0] is not None:
+                # index does not need shuffle
+                left_row_idx, left_row_inner_idx = splits[0][0][1][out_idx[0]]
+                right_row_idx, right_row_inner_idx = splits[0][1][1][out_idx[0]]
+            else:
+                left_row_idx = left_row_inner_idx = \
```
wjsi: Will these ever happen when we have already detected that no shuffle is needed? I also notice that this line is not covered in coverage report.

mars/dataframe/expressions/arithmetic/core.py
```diff
@@ -0,0 +1,358 @@
...
+                    right_row_idx = right_row_inner_idx = None
+            if splits[1] is not None:
+                # column does not need shuffle
+                left_column_idx, left_column_inner_idx = splits[1][0][1][out_idx[1]]
+                right_column_idx, right_column_inner_idx = splits[1][1][1][out_idx[1]]
+            else:
+                left_column_idx = left_column_inner_idx = \
```
wjsi: Will these ever happen when we have already detected that no shuffle is needed? I also notice that this line is not covered in coverage report.

mars/dataframe/execution/tests/test_arithmetic_execution.py
```diff
@@ -0,0 +1,105 @@
...
+
+        graph = df3.build_graph(tiled=True)
+        results = self.executor.execute_graph(graph, keys=[c.key for c in df3.chunks])
+
+        expected = data1 + data2
+        result = self._concat(expected.index, expected.dtypes, results)
+
+        pd.testing.assert_frame_equal(expected, result)
+
+    def testAddWithOneShuffleExecution(self):
```
wjsi: I notice that https://coveralls.io/builds/22915062/source?filename=mars%2Fdataframe%2Fexecution%2Farithmetic.py#L78 is not covered. Do we miss some case?

mars/dataframe/expressions/utils.py
```diff
@@ -84,22 +91,192 @@ def decide_chunk_sizes(shape, chunk_size, memory_usage):
...
+                if right_curr_idx + 1 >= len(right_min_max):
+                    right_terminate = True
+                else:
+                    right_curr_idx += 1
+                    right_curr_min_max = list(right_min_max[right_curr_idx])
+            else:
+                # from left min(left min close) to right min(exclude right min close)
+                right_curr_min_max[:2] = left_curr_min_max[:2]
+        else:
+            # left min == right min
```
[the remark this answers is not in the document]
chineking: Left partition will be `..., a) | [a, a] (a, ...` , `|` indicates that the [a, a] if belong to the second chunk. At the same time, right partition would be `..., a) [a, a] | (a, ...)`, and the `[a, a]` belongs to the first chunk. The behavior should be decisive, anyway, I will add a test to illustrate these.

mars/dataframe/expressions/arithmetic/core.py
chineking: After refine, `splits` is a class instead of endless nested list.

mars/dataframe/expressions/arithmetic/core.py
chineking: ditto

mars/dataframe/expressions/arithmetic/core.py
chineking: This never gotta happen, removed already.

mars/dataframe/expressions/arithmetic/core.py
chineking: ditto

mars/dataframe/execution/tests/test_arithmetic_execution.py
chineking: Previously only test shuffle on columns, now add test on index.

PR #368, opened 2019-05-06
Support **kwargs in new_session

NJUPole:
## What do these changes do?

As introduced in #294 , `WebSession` support passing arguments like `req_session`, `new_session` should support passing `**kwargs` when creating web session.

### DIFF
--- a/mars/deploy/local/session.py
+++ b/mars/deploy/local/session.py
@@ -27,7 +27,7 @@ from ...utils import build_graph
 
 
 class LocalClusterSession(object):
-    def __init__(self, endpoint):
+    def __init__(self, endpoint, **kwargs):
         self._session_id = uuid.uuid4()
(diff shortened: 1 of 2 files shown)


chineking:
This only effects the 0.1 branch without 0.2?

NJUPole:
> This only effects the 0.1 branch without 0.2?

Both, so this PR needs backport.

mars/deploy/local/session.py
```diff
@@ -38,6 +38,9 @@ def __init__(self, endpoint):
         # create session on the cluster side
         self._api.create_session(self._session_id)
 
+        if kwargs:
+            raise TypeError('Local cluster session do not support other arguments now.')
```
wjsi: Maybe remaining keys in kwargs can be printed here. Same issue for WebSession.

PR #375, opened 2019-05-07
Adjust memory estimation for TensorDot & fix defunct hard_limit

### LINKED ISSUE #373: [BUG] hard_limit is not effective in worker
**Describe the bug**
As is implemented in

https://github.com/mars-project/mars/blob/a6077669bf50ebe99aaa286beadb3c2796640e1c/mars/worker/quota.py#L397

``mem_stats.total - self._overall_size`` is always larger than 0, as self._overall_size is equal to a proportion of mem_stats.total. Therefore available_size is always equal to mem_stats.available when nothing is held in MemQuotaActor. Thus hard_limit configuration for a worker takes no effect.

### LINKED ISSUE #374: TensorDot does not deal with sparse inputs when calculating memory cost
**Describe the bug**
When calculating memory cost for dense TensorDots, we only multiply the size of output chunk with 2.5. When testing on a real machine with MKL, actual memory cost is no more than the size of the output chunk. What's more, when inputs are sparse while the output is dense, inputs shall be converted into dense and this will lead to extra memory cost. Memory size estimation of TensorDot should reflect these facts.

wjsi:
## What do these changes do?

1. Adjust memory estimation for TensorDot: fix according to actual memory cost when inputs are dense or sparse.
2. Fix defunct hard_limit in worker by changing typo in code.
3. Simplify logging in QuotaActor.

## Related issue number

Fixes #373 
Fixes #374

### DIFF
--- a/mars/worker/calc.py
+++ b/mars/worker/calc.py
@@ -294,12 +294,7 @@ class CpuCalcActor(WorkerActor):
             logger.debug('Start calculating operand %s.', op_key)
             start_time = time.time()
 
-            try:
-                result_pairs = self._calc_results(graph, context_dict, chunk_targets)
-            finally:
-                # release memory alloc for load keys
-                for k in direct_load_keys:
-                    self._mem_quota_ref.release_quota(k)
+            result_pairs = self._calc_results(graph, context_dict, chunk_targets)
 
             end_time = time.time()
 
--- a/mars/worker/calc.py
+++ b/mars/worker/calc.py
@@ -294,7 +294,12 @@ class CpuCalcActor(WorkerActor):
             logger.debug('Start calculating operand %s.', op_key)
             start_time = time.time()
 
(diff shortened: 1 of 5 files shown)


mars/worker/calc.py
```diff
@@ -294,12 +294,7 @@ def calc(self, session_id, ser_graph, chunk_targets, callback):
             logger.debug('Start calculating operand %s.', op_key)
             start_time = time.time()
 
-            try:
-                result_pairs = self._calc_results(graph, context_dict, chunk_targets)
-            finally:
-                # release memory alloc for load keys
-                for k in direct_load_keys:
-                    self._mem_quota_ref.release_quota(k)
+            result_pairs = self._calc_results(graph, context_dict, chunk_targets)
```
chineking: Remove because the finally part does not work?

PR #395, opened 2019-05-17
Fix defunct compose option in web session

### LINKED ISSUE #394: [BUG] compose=False is not effective in web session
**Describe the bug**
``compose=False`` is not effective in web session, as it passes the string ``False`` into server and the string is treated as True value.

wjsi:
## What do these changes do?

Pass string 0 or 1 instead of boolean string in web sessions to fix mismatch compose arg.

## Related issue number
Fixes #394


wjsi:
Designating whether to compose is not supported in v0.1, thus we do not back-port.

PR #403, opened 2019-05-21
Do consolidation for totiledb

### LINKED ISSUE #402: Consolidation should be triggered after all chunks have been stored to tiledb array
Currently, we just store each chunk into tiledb array. for each store, tiledb will create a directory with timestamp etc, thus this will generate a mount of directories. Consolidation should be triggered at last when all chunks have been processed.

qinxuye:
## What do these changes do?

Previously, we just store the chunk into tiledb array separately, but this will generate a mount of files, now the consolidation will be triggered at last when all chunks have been processed.

## Related issue number

Resolves #402


chineking:
tiledb 1.5.0 and tiledb-py 0.4.1 will cause core dump reported in https://github.com/TileDB-Inc/TileDB-Py/issues/141 . They solved the crash problem, but now I find that the consolidation seems not take effect now. Thus I will leave this PR draft until the problem solved.

qinxuye:
As tiledb-py released 0.4.3, come up with tiledb core lib 1.6, the consolidation issue is solved, so this PR is ready for review.

mars/tensor/expressions/datastore/totiledb.py
```diff
@@ -70,6 +72,18 @@ def _get_out_chunk(cls, op, in_chunk):
         return chunk_op.new_chunk([in_chunk], shape=out_chunk_shape,
                                   index=in_chunk.index)
 
+    @classmethod
+    def _process_out_chunks(cls, op, out_chunks):
+        if len(out_chunks) == 1:
+            return out_chunks
+
+        consolidate_op = TensorTileDBConsolidate(
```
hekaisheng: Should we introduce tree dependency ?

qinxuye: No, I think tree dependency is just a tradeoff, and looks uncomfortable, I hope we could discuss sort of new dependency except for data dependency, until then, we could resolve such operand.

hekaisheng: Agree.

PR #414, opened 2019-05-23
Check reference type in serialization

### LINKED ISSUE #413: [BUG] Reference type is not checked during serialization
**Describe the bug**

`ReferenceField` or `ListField` with reference supports reference other serializable objects. But the type is not checked during serialization.

**To Reproduce**

```
In [1]: from mars.serialize import ReferenceField, Serializable                                                                                                                                             

In [2]: class A(Serializable): 
   ...:     pass 
   ...:                                                                                                                                                                                                     

In [3]: class B(Serializable): 
   ...:     a = ReferenceField('a', A) 
   ...:                                                                                                                                                                                                     

In [4]: class C(Serializable): 
   ...:     pass 
   ...:                                                                                                                                                                                                     

In [5]: b = B()                                                                                                                                                                                             

In [6]: b.a = C()                                                                                                                                                                                           

In [8]: b.to_json()    # should raise error here                                                                                                                                                                                    
Out[8]: {'a': {}}

```

**Expected behavior**

Raise error when reference type is wrong during serialization.

chineking:
## What do these changes do?

Reference type in `ReferenceField` or `ListField` with reference should be checked.

## Related issue number

Fixes #413

### DIFF
--- a/mars/serialize/core.pyx
+++ b/mars/serialize/core.pyx
@@ -16,6 +16,7 @@
 
 import importlib
 import inspect
+import copy
 from collections import Iterable
 
 from ..compat import six, OrderedDict
@@ -121,7 +122,7 @@ cdef class ValueType:
 
 cdef class Field:
     def __init__(self, tag, default=None, bint weak_ref=False, on_serialize=None, on_deserialize=None):
-        self.model = None
+        self._model_cls = None
         self.attr = None
 
         self.tag = tag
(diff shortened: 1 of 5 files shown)


mars/serialize/core.pyx
```diff
@@ -431,6 +452,16 @@ cdef class ReferenceField(Field):
...
             self._model = model
 
+    @property
+    def model(self):
+        return self._model_cls
+
+    @model.setter
+    def model(self, new_model_cls):
+        if getattr(self, '_model', None) == 'self' and self._model_cls is not None:
+            raise SelfReferenceOverwritten('self reference is overwritten')
```
NJUPole: should check `new_model_cls` as in `ListField`

PR #426, opened 2019-05-30
Support creating tensor from list of tensors

### LINKED ISSUE #232: How to perform all tensor on np.array packages
Example：
```
import mars.tensor as mt
import numpy as np

def test_x():
    X1 = mt.tensor([0, 1])
    X2 = mt.tensor([2, 3])
    X = np.array([X1,X2])
    print(X.execute())
```
How to calculate X？

sighingnow:
## What do these changes do?

Support creating tensor from list of tensors by creating a `stack` op on those tensors.

## Related issue number

Fixes #232

### DIFF
--- a/mars/tensor/expressions/datasource/array.py
+++ b/mars/tensor/expressions/datasource/array.py
@@ -129,7 +129,8 @@ def tensor(data, dtype=None, chunk_size=None, gpu=None, sparse=False):
         if dtype is not None and data.dtype != dtype:
             return data.astype(dtype)
         return data
-    elif isinstance(data, tuple) and all(isinstance(d, TENSOR_TYPE) for d in data):
+    elif (isinstance(data, tuple) or isinstance(data, list)) and \
+            all(isinstance(d, TENSOR_TYPE) for d in data):
         from ..merge import stack
 
         data = stack(data)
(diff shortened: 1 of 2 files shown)


chineking:
Thank you for your contribution, our habit is write the `fixes #xxx` in the **Related issue number**, and  not included in the title.

mars/tensor/expressions/datasource/array.py
```diff
@@ -129,7 +129,8 @@ def tensor(data, dtype=None, chunk_size=None, gpu=None, sparse=False):
         if dtype is not None and data.dtype != dtype:
             return data.astype(dtype)
         return data
-    elif isinstance(data, tuple) and all(isinstance(d, TENSOR_TYPE) for d in data):
```
chineking: Could be `isinstance(data, (tuple, list))`

chineking:
The failure on PY35 of appveyor may happen sometimes due to unknown reason, rerun could make it work.

mars/tensor/expressions/datasource/array.py
sighingnow: Done.

sighingnow:
Thanks for the suggestion! I have revised the title and the patch. I have also noticed the non-deterministic failure on appveyor and have no idea about the reason.

Issue #434, opened 2019-06-01
[BUG] Fused operands may cause OOM


wjsi:
**Describe the bug**
When operands are fused and we estimate memory costs, the memory cost of intermediate variables are omitted. For instance, when TensorDot and Triu is fused, we get an executable dag for worker below:

![image](https://user-images.githubusercontent.com/8284922/58746960-0cbaaf00-8497-11e9-93ad-1696856ce83c.png)

The maximal memory cost shall be max{Proc0 + Proc1 + Triu, Proc0 + TensorDot}. However, in current implementation, we calc max{Proc1 + Triu, Proc0 + TensorDot} and Proc0 is omitted for Triu, which causes OOM in practical execution.

**Additional context**
This bug can both be triggered in v0.1 and v0.2. Given differences in executors, bugs may be fixed in separate PRs.

PR #435, opened 2019-06-02
Fix memory size estimation for Fuse nodes in v0.1

wjsi:
## What do these changes do?

Fix execution memory size estimation for fused dags sent to workers. The value is now total stored size in chunk_result minus total size of Fetch nodes.

Note that we assume that Numexpr and Cupy will automatically reuse allocated buffers if the size is sufficient, we does not calculate memory size of inputs when estimating sizes for CpFuse and NeFuse. The arg ``update_local`` indicates the behavior we use when calculating sizes.

## Related issue number

#434 (fix for v0.1)

### DIFF
--- a/mars/tensor/execution/core.py
+++ b/mars/tensor/execution/core.py
@@ -82,13 +82,13 @@ class Executor(object):
             return cls._get_op_runner(chunk, cls._op_size_estimators)(results, chunk)
 
     def execute_graph(self, graph, keys, n_parallel=None, n_thread=None, show_progress=False,
-                      compose=True, fetch_keys=None, mock=False):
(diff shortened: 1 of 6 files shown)


mars/tensor/execution/core.py
```diff
@@ -309,6 +321,12 @@ def execute_graph(graph, keys, executor, n_parallel=None, show_progress=False,
         dict(((t.key, t.id), v) for t, v in graph.iter_successor_items())
     )
 
+    fetch_keys = set(v.key for v in graph if isinstance(v.op, TensorFetchChunk))
```
chineking: `fetch_keys` is only for `mock`, could be triggered only for `mock`is True.

PR #443, opened 2019-06-11
Remove handle_params and calc_shape in operand class

### LINKED ISSUE #438: [ENH] Role of _handle_params and _set_inputs conflict with each other
In ``mars.tensor``, ``_handle_params`` picks inputs for args like ``lhs`` or ``rhs`` when initializing tensors or chunks, while ``_set_inputs`` does the same thing when setting ``inputs`` of tensors or chunks. Therefore one logic is implemented in two different functions. This shall be changed esp. for operands with ``_handle_params`` but without ``_set_inputs``, as when ``inputs`` of these operands are set, arguments are not updated which may lead to errors. 

A recommended solution is to remove ``_handle_params`` and move all logic into ``_set_inputs``.

NJUPole:
## What do these changes do?

The `calc_shape` method was introduced in #128 to calculate `rough_nbytes` for unknown shape tensors, however `rough_nbytes` was removed in #266 , the `calc_shape` is not used by any module, so they are totally removed now. 

Also removed the `handle_params` method in some special operands and move the logic to `set_inputs` to simplified the operand class.

This PR has two commits, the first one just do some deletions, the second one may need detailed review.

## Related issue number

resolve #438

### DIFF
--- a/mars/tensor/expressions/arithmetic/clip.py
+++ b/mars/tensor/expressions/arithmetic/clip.py
@@ -24,7 +24,7 @@ from ....core import Base, Entity
 from ...core import Tensor
 from ..utils import broadcast_shape
 from ..datasource import tensor as astensor
-from .core import TensorOperand, TensorElementWise
+from .core import TensorOperand, TensorElementWise, filter_inputs
 
 
 class TensorClip(TensorOperand, TensorElementWise):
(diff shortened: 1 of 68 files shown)


wjsi:
Shall this PR be backported?

NJUPole:
> Shall this PR be backported?

It will be, I have added the  label `to be backported`.

mars/tensor/expressions/arithmetic/clip.py
```diff
@@ -51,80 +51,26 @@ def a_max(self):
...
-            has_out = False
-            out = next(inputs_iter)
-            if out is not None:
-                has_out = True
-                inps.append(out)
+            self._a = next(inputs_iter)
+            self._a_min = next(inputs_iter) if isinstance(inputs[1], (Base, Entity)) else inputs[1]
+            self._a_max = next(inputs_iter) if isinstance(inputs[2], (Base, Entity)) else inputs[2]
+            if inputs[3] is not None:
+                self._out = next(inputs_iter)
```
wjsi: When either inputs[1] or inputs[2] is neither Base nor Entity, self._out will be inputs[1]. Is this behavior expected?

mars/tensor/expressions/arithmetic/core.py
```diff
@@ -220,6 +176,27 @@ def casting(self):
...
 
+    def _set_inputs(self, inputs):
+        super(TensorBinOp, self)._set_inputs(inputs)
+        inputs_iter = iter(self._inputs)
+
+        if getattr(self, '_lhs', None) is None:
+            # create binop from beginning
+            self._lhs = next(inputs_iter)
+            self._rhs = next(inputs_iter)
+            if inputs[2] is not None:
```
wjsi: Similar question to clip.py. iter() and getitem should not be mixed in case of ambiguity. This may apply to all ``_set_inputs()`` changed in this PR.

mars/tensor/expressions/arithmetic/clip.py
NJUPole: `self._inputs` are all `Base` or `Entity`, if inputs[1] or inputs[2] is neither Base nor Entity, and `self._out` is not `None`, then `self._inputs` has only two elements.

> self._out will be inputs[1] while we just checked inputs[3]

`self._out` will be `self._inputs[1]`

wjsi: So why we need to check inputs[3] while we actually deal with inputs[1]?

NJUPole: it's self._inputs[1], not the inputs[1], they are different after calling the `super(TensorClip, self)._set_inputs(inputs)`.

wjsi: Therefore it is better to reference inputs in a unified manner, for instance, use inputs[idx] instead of mixing self._inputs, inputs and inputs_iter.

wjsi:
``a_max`` arg of ``clip()`` is not tested when it is a Tensor, pls fix.

Issue #415, opened 2019-05-24
[PROPOSAL] Mutable tensor abstraction to expose storage layer of Mars


qinxuye:
Mars is an execution engine dedicated to perform massive tensor or dataframe computation. Sometimes, if users want to just partially or concurrently update or read data in tensor, it's not effective and intuitive enough.

Thus I propose an abstraction called mutable tensor, which is different from non-volatile tensor that the user can create a chunked mutable tensor, and read and write chunk data concurrently. At last, users can transform the mutable tensor to a vanilla tensor to perform upcoming computations.

The API designs are listed below.

# Create mutable tensor

Create a mutable tensor via `session.create`.

```python
import mars.tensor as mt
from mars.session import new_session

sess = new_session('http://***', id='exist_session_id')

# create a 1000x2000 mutable tensor wich chunk size 500 on each dimension
# other clients can ref the same mutable tensor by the `name`
# compression and size of redundant can also be specified
a = mt.mutable_tensor(1000, 2000, chunk_size=500, name='mutable_tensor1',
                      compression='lz4', n_redundant=2)
sess.create(a)
```

# Get mutable tensor

Like I said, other clients can get the mutable tensor by its name.

```python
from mars.session import new_session

sess = new_session('http://***', id='exist_session_id')

a = sess.get('mutable_tensor1')
```

# Delete mutable tensor

```python
from mars.session import new_session

sess = new_session('http://***', id='exist_session_id')

sess.delete('mutable_session')
```

# Alias mutable tensor with another name

```python
from mars.session import new_session

sess = new_session('http://***', id='exist_session_id')
sess.alias(a, 'mutable_tensor2')
sess.get('mutable_tensor2')
```

# Write chunk data into mutable tensor

Write data by slice which is quite intuitive. At first, we can support write into only one chunk data with exact the chunk size, but we can support other situations later.

```python
import numpy as np

a[500:1000, :500] = np.empty(500, 500)
```

Other situations include fancy indexing way.

```python
import numpy as np

a[[0, 1, 2, 3], [4, 5, 6, 7]] = np.asarray([8, 9, 10, 11])
```

# Read chunk data

Still we can only support the exact range of the chunk at first. User can get a numpy ndarray here.

```python
array = a[500:1000, :500]
```

Fancy indexing can also be considered.

```python
import numpy as np

array2 = a[[0, 1, 2, 3], [4, 5, 6, 7]]
```

# Transform mutable tensor into vanilla tensor

[...]

PR #464, opened 2019-06-17
Add mutable tensor support for local cluster

### LINKED ISSUE #439: Provide a MutableTensorActor to manage mutable tensor

Mutable tensor is mentioned in #415 , when a mutable tensor is created in a session, the `SessionActor` can create `MutableTensorActor` to manage the status of mutable tensor on the schedulers.

A `MutableTensorData` and `MutableTensor` can be added as well.

**Additional context**

Refer to https://github.com/mars-project/mars/issues/415#issuecomment-499382927 .

### LINKED ISSUE #450: Worker should provide a way to read or write a chunk data

Worker should provide a way to read or write a chunk data, it would be helpful for web or scheduler if a numpy ndarray needs to be written into a worker.

I thought we can leverage `ReceiverActor`, but it could be non-trivial. We can explore a better way if needed.

**Additional context**

This is important for implementation of mutable tensor.

sighingnow:
## What do these changes do?

Support mutable tensor.

+ A `MutableTensorActor` on scheduler to manager mutable tensors and perform the underlying operations.
+ A `MutableTensor` and `MutableTensorData` to manage chunks of mutable tensors.
+ `allocate/read/write` interface in `ResultSenderActor` (on workers).
+ Chunk allocation:

   - The chunks are stored in plasma store (currently pinned in memory for convenience) before seal.
   - No compression is performed on mutable tensor thunks, and `read/write` operation is implemented by reading/writing a `np.ndarray` constructed by the memory view directly (zero copy and no read/write overhead). 

TODO

- [x] Finish `write_mutable_tensor`.
- [x] Construct `Tensor` (and `TensorData`) using those thunks directly when seal.
- [x] Do the `read/write` dispatch logic on client (now it runs on scheduler)

## Related issue number

#415
Resolves #439 
Resolves #450.

### DIFF
--- a/mars/session.py
+++ b/mars/session.py
@@ -198,7 +198,8 @@ class Session(object):
                 self._sess.create_mutable_tensor(name, shape, dtype, *args, **kwargs)
         # Construct MutableTensor on the fly.
         tensor = create_fetch_tensor(chunk_size, shape, dtype, chunk_keys)
-        return MutableTensor(data=MutableTensorData(_name=name, _op=None, _shape=shape, _dtype=dtype,
+        return MutableTensor(chunk_size=chunk_size,
+                             data=MutableTensorData(_name=name, _op=None, _shape=shape, _dtype=dtype,
                                                     _nsplits=tensor.nsplits, _chunks=tensor.chunks))
 
     def get_mutable_tensor(self, name):
@@ -208,7 +209,8 @@ class Session(object):
         shape, dtype, chunk_size, chunk_keys = self._sess.get_mutable_tensor(name)
         # Construct MutableTensor on the fly.
         tensor = create_fetch_tensor(chunk_size, shape, dtype, chunk_keys)
-        return MutableTensor(data=MutableTensorData(_name=name, _op=None, _shape=shape, _dtype=dtype,
+        return MutableTensor(chunk_size=chunk_size,
+                             data=MutableTensorData(_name=name, _op=None, _shape=shape, _dtype=dtype,
                                                     _nsplits=tensor.nsplits, _chunks=tensor.chunks))
     def write_mutable_tensor(self, tensor, index, value):
         self._ensure_local_cluster()
--- a/mars/session.py
+++ b/mars/session.py
@@ -198,8 +198,7 @@ class Session(object):
                 self._sess.create_mutable_tensor(name, shape, dtype, *args, **kwargs)
         # Construct MutableTensor on the fly.
         tensor = create_fetch_tensor(chunk_size, shape, dtype, chunk_keys=chunk_keys)
-        return MutableTensor(chunk_size=chunk_size,
-                             data=MutableTensorData(_name=name, _op=None, _shape=shape, _dtype=dtype,
+        return MutableTensor(data=MutableTensorData(_name=name, _op=None, _shape=shape, _dtype=dtype,
                                                     _nsplits=tensor.nsplits, _chunks=tensor.chunks))
 
     def get_mutable_tensor(self, name):
@@ -209,8 +208,7 @@ class Session(object):
         shape, dtype, chunk_size, chunk_keys = self._sess.get_mutable_tensor(name)
         # Construct MutableTensor on the fly.
         tensor = create_fetch_tensor(chunk_size, shape, dtype, chunk_keys=chunk_keys)
-        return MutableTensor(chunk_size=chunk_size,
-                             data=MutableTensorData(_name=name, _op=None, _shape=shape, _dtype=dtype,
+        return MutableTensor(data=MutableTensorData(_name=name, _op=None, _shape=shape, _dtype=dtype,
                                                     _nsplits=tensor.nsplits, _chunks=tensor.chunks))
     def write_mutable_tensor(self, tensor, index, value):
         self._ensure_local_cluster()
@@ -229,7 +227,7 @@ class Session(object):
     def _ensure_local_cluster(self):
         from .deploy.local.session import LocalClusterSession
         if not isinstance(self._sess, LocalClusterSession):
-            raise RuntimeError("Only local cluster session can manipulate mutable tensors")
+            raise RuntimeError("Only local cluster session can be used to manipulate mutable tensors.")
 
 def new_session(scheduler=None, **kwargs):
     return Session(scheduler, **kwargs)
--- a/mars/session.py
+++ b/mars/session.py
@@ -197,7 +197,7 @@ class Session(object):
         shape, dtype, chunk_size, chunk_keys = \
                 self._sess.create_mutable_tensor(name, shape, dtype, *args, **kwargs)
         # Construct MutableTensor on the fly.
-        tensor = create_fetch_tensor(chunk_size, shape, dtype, chunk_keys)
+        tensor = create_fetch_tensor(chunk_size, shape, dtype, chunk_keys=chunk_keys)
         return MutableTensor(chunk_size=chunk_size,
                              data=MutableTensorData(_name=name, _op=None, _shape=shape, _dtype=dtype,
                                                     _nsplits=tensor.nsplits, _chunks=tensor.chunks))
@@ -208,7 +208,7 @@ class Session(object):
         self._ensure_local_cluster()
         shape, dtype, chunk_size, chunk_keys = self._sess.get_mutable_tensor(name)
         # Construct MutableTensor on the fly.
-        tensor = create_fetch_tensor(chunk_size, shape, dtype, chunk_keys)
+        tensor = create_fetch_tensor(chunk_size, shape, dtype, chunk_keys=chunk_keys)
         return MutableTensor(chunk_size=chunk_size,
                              data=MutableTensorData(_name=name, _op=None, _shape=shape, _dtype=dtype,
                                                     _nsplits=tensor.nsplits, _chunks=tensor.chunks))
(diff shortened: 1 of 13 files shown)


sighingnow:
I would like to share my current progress on `seal` with you.

My current implementation is allocating chunks directly on workers for mutable tensor. When `seal`, I create a `Tensor` from those thunks and return to the client. I use `Fetch` as the op of thunks of the generated tensor.

Now the trouble is, I cannot find a clean way to make `SessionActor` know that the tensor generated from mutable tensor has already been evaluated (has thunks).

+ If a tensor has been executed, there would be a reference in `SessionActor._tileable_to_graph`, thu s I need to build a `GraphActor` ref for the created tensor.
+ In `GraphActor`, for `Fetch` op, `tile_fetch_tileable` is called (in `prepare_graph`), that means there is already a GraphActor reference in the `SessionActor._tileable_to_graph`.
+ If I mark the GraphActor's state as `SUCCESSED` and don't run `prepare_graph`, the `_get_tileable_by_key` in GraphActor would fail when evaluating other tensors that depends on the generated tensor.
+ To make `_get_tileable_by_key` success, I have tried to set the tiled tensor (has chunks) to `_tileable_key_to_opid` and `_tileable_key_opid_to_tiled`, but it is uneasy to serialize/deserialize chunks between workers.

Another approach is creating `GraphActor` when creating the mutable tensor, and do `prepare_graph` then we could have those chunks, and the `GraphActor` also has those tensors. But that means we need to do mutable tensor's operations (read/write/seal) in `GraphActor`.

sighingnow:
Or, should I make `MutableTensorActor` Inherit from `GraphActor` to share those data structure?

sighingnow:
Now `MutableTensorActor` is a subclass of `GraphActor`, and `seal` could works!

We now have a `MutableTensor` as follows:

```
>>> import numpy as np
>>> from mars.deploy.local import new_cluster
>>> from mars.session import new_session
>>> cluster = new_cluster()
>>> session = new_session(cluster.endpoint)
>>> session.as_default()

>>> mtensor = session.create_mutable_tensor("test", (4, 4), chunk_size=3)
>>> mtensor[:]
array([[0, 0, 0, 0],
       [0, 0, 0, 0],
       [0, 0, 0, 0],
       [0, 0, 0, 0]], dtype=int8)
>>> mtensor[1:3] = np.arange(8).reshape(2, 4)
>>> mtensor[3:4]
array([[0, 0, 0, 0]], dtype=int8)
>>> mtensor[2:4]
array([[4, 5, 6, 7],
       [0, 0, 0, 0]], dtype=int8)
>>> x = mtensor.seal()
>>>
>>> session.run(x + 1)
array([[1, 1, 1, 1],
       [1, 2, 3, 4],
       [5, 6, 7, 8],
       [1, 1, 1, 1]], dtype=int8)
>>>
```

chineking:
Excellent work! I will take some time to review the implementation.

wjsi:
@sighingnow Please rebase your fork onto mars/master and update your code as we renamed all ``default_name`` into ``default_uid``.

sighingnow:
Thanks for the comprehensive review. I would like to explain some implementation decisions based on the comments. The expected design is just storing the affected index, timestamp, and data separately and merge into a whole tensor when `seal`. But the concern is:

+ First, how to record the affected index and value? The `__setitem__` of numpy could be really "fancy".
  - If we record them in a sparse manner (something like `(idx_x, idx_y, idx_z): value)` then the performance cannot be good when writing in a chunk (or continuous) fashion, and a write with broadcast will generate a large volume of such index-value record.
  - If we just record the initial index and value (something like `[slice(1, 2, None)], value`), then the merge operation is just *replay* all users operation _on the same worker_ to combine all these operations. The performance won't be good, too.
+ Second, how to do the merge? One of our assumptions is that, the data is so large that cannot be held in only one machine, then if the write is not sparse, it is not trivial to do a merge when sealing it.

Now back to the current design. The basic insight is that the mutable tensor won't be kept in memory for a long time since the mutable tensor is only used to create a tensor by a series of `write` operation, IMO. Then,

> It's not safe, if many clients try to modify a same chunk, no transaction guaranteed.

Based on the assumption that the mutable tensor is only used to create tensor, the write operations issued by the client won't overlap, or the overlapping doesn't affect the result. If we use the design that just recording the affected index and value, I think we couldn't give such transaction guarantee as well.

> The data would be preferred to be pinned in the plasma store always, it there are other computation, there would be no room for it.

If the mutable tensor won't exist in memory for a long time, keep it in memory could provide the best wirte performance. When it is sealed, the chunks would be unpinned.

> I doubt the performance would not be great especially for write when many small modifies happen.

I think the write performance should be even better compared to the design that just recording the write operation and then do a merge. Because,

+ The memory is pre-allocated.
+ When there are many small write, we also need to decide which worker to be used to write the record as well, just like what we do in this PR to route the write operation. (The route logic can be moved to client side, though).
+ In this PR, the write on plasma store is just writing on a view of raw memory and without any extra copy. The chunk is pinned in the memory and is not serialized in plasma store, guaranteeing the best write performance that can be achieved (IMHO).

sighingnow:
A previous bug has been fixed now and local test shows that now `write` is ok for such cases:

```python
mtensor = session.create_mutable_tensor("test", (4, 4), chunk_size=3)

# write slice cross chunks
mtensor[2:4] = np.arange(8).reshape(2, 4) 

# write with broadcast
mtensor[1] = np.arange(4).reshape(4)

# write scalar value (cross chunks)
mtensor[1:4, 2] = 8

x = mtensor.seal()
print(session.fetch(x))

```

sighingnow:
CI is green.

qinxuye:
> CI is green.

Great, I will review soon later.

mars/session.py
```diff
@@ -190,6 +190,46 @@ def default_or_local(cls):
...
+        self._ensure_local_cluster()
+        chunk_records_to_send = tensor._do_flush()
+        self._sess.send_chunk_records(tensor.name, chunk_records_to_send)
+        shape, dtype, chunk_size, chunk_keys = self._sess.seal(tensor.name)
+        # Construct Tensor on the fly.
+        return create_fetch_tensor(chunk_size, shape, dtype, chunk_keys)
+
+    def _ensure_local_cluster(self):
+        from .deploy.local.session import LocalClusterSession
+        if not isinstance(self._sess, LocalClusterSession):
```
qinxuye: So for now only local cluster is supported?

mars/tensor/core.py
```diff
@@ -15,16 +15,19 @@
...
-from ..serialize import ProviderType, ValueType, DataTypeField, ListField, TupleField
-from ..utils import on_serialize_shape, on_deserialize_shape
+from ..serialize import ProviderType, ValueType, DataTypeField, ListField, TupleField, BoolField, StringField
+from ..utils import log_unhandled, on_serialize_shape, on_deserialize_shape
 from .expressions.utils import get_chunk_slices
 
+import logging
+logger = logging.getLogger(__name__)
 
 class TensorChunkData(ChunkData):
```
qinxuye: Add a blank before due to pep8.

mars/tensor/core.py
```diff
@@ -391,6 +394,137 @@ def totiledb(self, uri, ctx=None, key=None, timestamp=None):
     def execute(self, session=None, **kw):
         return self._data.execute(session, **kw)
 
+class MutableTensorData(TensorData):
```
qinxuye: Ditto.

mars/tensor/core.py
```diff
@@ -391,6 +394,137 @@ def totiledb(self, uri, ctx=None, key=None, timestamp=None):
...
+
+    @property
+    def name(self):
+        return getattr(self, '_name', None)
+
+    @property
+    def compression(self):
+        return getattr(self, '_compression', None)
+
+class MutableTensor(Entity):
```
qinxuye: Ditto.

mars/tests/test_mutable.py
```diff
@@ -0,0 +1,172 @@
...
+@unittest.skipIf(sys.platform == 'win32', 'does not run in windows')
+class Test(unittest.TestCase):
+    def testMutableTensorCreateAndGet(self):
+        with new_cluster(scheduler_n_process=2, worker_n_process=2,
+                         shared_memory='20M') as cluster:
+            with new_session(cluster.endpoint) as session:
+                mut1 = session.create_mutable_tensor("test", (4, 5), dtype=np.double, chunk_size=3)
+                mut2 = session.get_mutable_tensor("test")
+
+                self.assertEqual(mut1.shape, (4, 5))
```
qinxuye: Add a test to make sure `mut1` is not `mut2`.

mars/session.py
sighingnow: The limitation is because we need the client can request the worker directly when writing to mutable tensor.

For `LocalSession`, I think there are no actors like `PlasmaStoreActor`? For `WebSession`, I haven't looked into the details, but I think many things would be different, compared with `LocalClusterSession`?

qinxuye: I think we can focus on local cluster for this PR, and we need to modify the name to mark the limitation.

For local session, there is no plasma, because it's threaded. I think it's not so reasonable for local session, because user can use numpy instead, but if we want to make mutable tensor available for all schedulers, we could implement a naive one for local session.

As for web session, we need to add some RESTful interface.

I think we can add support for local and distributed schedulers in the next PR.

mars/worker/seal.py
```diff
@@ -0,0 +1,85 @@
...
+                        ndarr[idx] = record.value
+            finally:
+                del buf
+
+            # clean up
+            self._chunk_holder_ref.unregister_chunk(session_id, key)
+            self.get_meta_client().delete_meta(session_id, key, False)
+            self._mem_quota_ref.release_quota(key)
+
+        self._chunk_store.put(session_id, chunk_key, ndarr)
```
wjsi: You need to hold the result of put() before register_chunk(), as the data put into plasma_store may be collected by GC.

sighingnow: Fixed. Thanks for letting me know this feature of plasma!

qinxuye:
Please rebase master branch to pass travis ci.

sighingnow:
Could someone help me relaunch the failed py35 job on appveyor? The [job under my personal account is ok](https://ci.appveyor.com/project/sighingnow/mars/builds/25766114) for the latest commit.

Issue #468, opened 2019-06-19
Fix the str representation for TensorChunkData and ChunkData


sighingnow:
Is there any particular reason to hide the `Data` suffix in `__repr__` and `__str__` of `ChunkData` and `TensorChunkData`? It is quite misleading when debugging and I spent almost one hour to find out what happended and then realize it is not `Chunk`, but `ChunkData`.....

 Related code:

https://github.com/mars-project/mars/blob/3f0bd7a36a532a007533edf0809ad61f29e60969/mars/core.py#L228-L230

https://github.com/mars-project/mars/blob/3f0bd7a36a532a007533edf0809ad61f29e60969/mars/tensor/core.py#L92-L107

chineking:
Yeah, I think we can repr Chunk data as `ChunkData <op=xxx, key=xxx>`, so does `TensorData`.

sighingnow:
The `__repr__` and `__str__` are usually called when *the users* print a `Tensor` in python/ipython shell. To fix the issue and without affecting users, we have

+ plan A: move the `__repr__` and `__str__` from `TensorData` to `Tensor`, and display it as `Tensor`
+ plan A: implement `__repr__` and `__str__` in `TensorData` as `TensorData`, and add a corresponding `__repr__` and `__str__` in `Tensor` as well, displaying as `Tensor`. That would lead to a lot of code duplication.
+ plan C: just leave it as current implementation the ordinary users won't be affected

Personally speaking, I now prefer plan C (and close this issue as invalid 😢)

NJUPole:
I also choose plan C, users don't need to know the difference between `Tensor` and `TensorData`.

So i will close this issue and anyone has other suggestions could reopen it.

Issue #508, opened 2019-07-03
[BUG] print(DataFrame) doesn't present its data in terminal in eager mode 


wdkwyf:
**Describe the bug**
In eager mode, print a DataFrame doesn't output it's data.

**To Reproduce**
1. Your Python version
3.7.3
2. The version of Mars you use
v0.2.0b1
3. Versions of crucial packages, such as numpy, scipy and protobuf
DataFrame
4. Full stack of the error.
Unwanted behavior:
print(DataFrame) in eager mode doesn't print the data of the DataFrame.
5. Minimized code to reproduce the error.
When I tried:
```
import mars.dataframe as md
from mars.config import options
options.eager_mode = True

a = md.DataFrame({'angles': [0, 3, 4]})
print(a)
```
The output is:
`<mars.dataframe.core.DataFrameData object at 0x7fafe8e54208>
`
which is not right.
**Expected behavior**
Output should be:
```
   angles
0       0
1       3
2       4
```

wjsi:
This may have something to do with #490 which will be fixed soon.

sighingnow:
I have opened #511 to catch the release timeline. Sorry for that I didn't notice the assignees field of this issue.

wdkwyf:
> I have opened #511 to catch the release timeline. Sorry for that I didn't notice the assignees field of this issue.

No problem. I haven't started doing this yet.
Thanks for your contributions.

qinxuye:
> This may have something to do with #490 which will be fixed soon.

This issue is due to the `repr` of DataFrame that the data is not shown when eager mode is on. It's not strongly related to corner or not. They can be fixed separately.

NJUPole:
> This issue is due to the `repr` of DataFrame that the data is not shown when eager mode is on. It's not strongly related to corner or not. They can be fixed separately.

Agree, we can merge #511 firstly to solve this issue.

sighingnow:
Fixed by #511.

PR #511, opened 2019-07-04
Define __str__ and __repr__ for DataFrame, Index, and Series.

sighingnow:
## What do these changes do?

Define `__str__` and `__repr__` for `DataFrameData`, `IndexData` and `SeriesData` properly, as what we have done for `TensorData`, and respect the eager mode settings.

## Related issue number

#508.


qinxuye:
Does this PR fixes #468 as well?

NJUPole:
It's better to add some tests for this.

sighingnow:
It doesn't fixes #468. I have commented under that issue.

PR #512, opened 2019-07-04
Refactor web using standalone tornado handlers

### LINKED ISSUE #510: Update progress dynamically
Bokeh supports updating graphs dynamically. This can be added to Mars to facilitate tracing the progress of a running task.

wjsi:
## What do these changes do?

Previously Mars web uses bokeh applications to render all the pages, which is inflexible as bokeh does not support urls with multiple slashes, and we have to use complex query strings. To avoid this and pave way for more information on the web ui, tornado handlers are now used and bokeh applications are limited to graphs and embedded into pages with scripts.

Dynamic refreshing issue of progress graphs are also fixed.

## Related issue number

Fixes #510

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,9 @@ matrix:
       language: generic
       env: PYTHON=3.7
 
+addons:
+  apt:
+    update: true
 before_install:
   - source ./bin/travis-install-conda.sh
   - if [[ "$PYTHON" =~ "2" ]]; then
--- a/mars/deploy/local/core.py
+++ b/mars/deploy/local/core.py
@@ -14,10 +14,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import print_function
+
 import atexit
 import multiprocessing
 import os
 import signal
+import sys
 import time
 
 from ...actors import create_actor_pool
@@ -284,7 +287,8 @@ class LocalDistributedClusterClient(object):
         self._ensure_process_finish(self._web_process)
 
 
-def new_cluster(address='0.0.0.0', web=False, n_process=None, shared_memory=None, **kw):
+def new_cluster(address='0.0.0.0', web=False, n_process=None, shared_memory=None,
+                open_browser=True, **kw):
     endpoint = gen_endpoint(address)
     web_endpoint = None
     if web is True:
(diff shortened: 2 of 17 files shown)


.travis.yml
```diff
@@ -59,10 +59,12 @@ install:
...
       done
     fi
+  # temporarily use older version of pyarrow
+  - pip install "pyarrow<0.14"
+  - conda install --quiet --yes -n test -c conda-forge python=$PYTHON "tiledb-py>=0.4" pyarrow || true
   - pip install -r requirements-dev.txt &&
     pip install -r requirements-extra.txt &&
     pip install virtualenv coveralls flake8 etcd-gevent
-  - conda install --quiet --yes -n test -c conda-forge python=$PYTHON "tiledb-py>=0.4" || true
   - virtualenv testenv && source testenv/bin/activate
```
sighingnow: We need to enforce the pyarrow version in requirements.txt, rather than ci configuration. Otherwise, our users who use py36 will be affected after they upgrade mars.

qinxuye: I think we can limit the ci version first, and then find a compatible way for new version of pyarrow. The limitation like ">=0.11,<0.14" is sometimes annoying especially packages conflict.

wjsi: The reason to add this is that we encounter an ImportError, not compatibility issue. This is only a workaround and I think pyarrow will fix that soon.

.travis.yml
```diff
@@ -59,10 +59,12 @@ install:
         sed -i.bak "s/plugins *= *Cython\.Coverage//g" $cf;
       done
     fi
+  # temporarily use older version of pyarrow
+  - pip install "pyarrow<0.14"
+  - conda install --quiet --yes -n test -c conda-forge python=$PYTHON "tiledb-py>=0.4" pyarrow || true
```
qinxuye: Is it duplicated that pyarrow is installed by pip and conda?

mars/deploy/local/core.py
```diff
@@ -300,6 +304,10 @@ def new_cluster(address='0.0.0.0', web=False, n_process=None, shared_memory=None
     web_process = None
     if web_endpoint:
         web_process = _start_web_process(endpoint, web_endpoint)
+        print('Web endpoint started at http://%s' % web_endpoint, file=sys.stderr)
```
qinxuye: Is it better to use logger instead of print?

wjsi: The default log level is warning, but it is strange to display this message as warning.

.travis.yml
wjsi: I just installed a pyarrow==0.14 under linux and it works well with Mars. Maybe the true cause of the error needs investigation.

sighingnow: I can reproduce [the error on Linux with python 36 and pyarrow 0.14](https://travis-ci.org/mars-project/mars/jobs/554169173), but haven't furtherly dug into the failure yet.

wjsi: The failure is caused by missing ``liblz4``. Already fixed.

PR #523, opened 2019-07-08
Fix bug that redundant graphs are submitted unexpectedly when eager mode is on

### LINKED ISSUE #522: [BUG] Redundant tasks are submitted when eager mode is on
**Describe the bug**
In eager execution, some empty graphs were generated and submitted to cluster which will lead to inefficiency.

**To Reproduce**

``` Python
In [5]: cluster = new_cluster(scheduler_n_process=2, worker_n_process=2, shared_memory='20M', web=True)

In [6]: import mars.tensor as mt                                                                                              

In [7]: from mars.config import options                                                                                       

In [8]: options.eager_mode = True                                                                                             

In [9]: a = mt.random.random((10,), chunk_size=3)                                                                             

In [10]: b = mt.random.rand(10, chunk_size=3)                                                                                 

In [11]: c = a + b
```
In theory, the code above should generate three graphs in this session, there are five however, the extra two graphs are empty.

image

hekaisheng:
## What do these changes do?
When eager mode is on, an entity will be executed once it is created, so we should use `kernel_mode` to decorate the function `build_graph` to avoid execution. 

## Related issue number
fixes #522

### DIFF
--- a/mars/deploy/local/tests/test_cluster.py
+++ b/mars/deploy/local/tests/test_cluster.py
@@ -20,7 +20,6 @@ import os
 import sys
 import time
 import unittest
-import json
 
 import numpy as np
 try:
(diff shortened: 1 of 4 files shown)


mars/deploy/local/tests/test_cluster.py
```diff
@@ -478,6 +479,11 @@ def testEagerMode(self, *_):
                     expected2 = expected1.dot(expected1)
                     np.testing.assert_array_almost_equal(r2, expected2)
 
+                    web_session = Session.default_or_local()._sess
+                    tasks_url = 'http://{0}/api/session/{1}/graph'.format(cluster.web_endpoint,
```
qinxuye: Can we add a web API like `n_tasks` to make it more clean?

hekaisheng: OK, I wasn't sure it's necessary to add an API for it.

PR #536, opened 2019-07-11
Mutable tensor for local session and web session

sighingnow:
## What do these changes do?

Support mutable tensor in local session and web session.

+ For the `LocalSession`, we
   1. maintain a map of `name -> mut tensor` in `LocalSession`, and do write to the ndarray directly.
  2. when `seal`, we use the ndarray to construct a `mt.tensor` and execute it with `self._executor` know the tensor is executed.
+ For the `WebSession`, we
  1. add several necessary endpoints to web api server.
  2. when `write`, we forward the `index` and `value` to `MutableTensorActor`, the `MutableTensorActor` maintains the buffer, and send record chunks to corresponding workers.
  3. When `seal`, do the same thing with `LocalClusterSession`.

Some design point that needs to clarify:

1. The `index` (such as `(slice(1, None, None))`) cannot be serialized to json directly, thus I add a helper class `MutableTensor.Index` and leverage the `Serializable.to_json` to do the serialization.
2. In `WebSession`, the index calculation and chunk transfer are done in `MutableTensorActor`.
3. All functionalities of mutable tensor in `LocalSession` work well with the `WebSession` (as shown by the unit test).
4. Along this PR I also do some refactor and improvement on the previous implementation to avoid code duplication.

About the (known) failed test cases:

It seems that there is no way to obtain the information and message of exception that initially raised in server at client side. We have `dump_exception` and `reraise`,

(Edit: this exception info issue has been fixed in [commit 31520ebc9](https://github.com/mars-project/mars/pull/536/commits/31520ebc9ab3ee31ce65b369c076998269783b34) of this PR.)

```python
    def _dump_exception(self, exc_info):
        pickled_exc = pickle.dumps(exc_info)
        self.write(json.dumps(dict(
            exc_info=base64.b64encode(pickled_exc),
        )))
        raise web.HTTPError(500, 'Internal server error')
```

```python
    if resp.status_code >= 400:
        resp_json = json.loads(resp.text)
        exc_info = base64.b64decode(resp_json['exc_info'])
        six.reraise(*exc_info)
```

But it doesn't work because:

1. The `base64.b64encode(pickled_exc)` is `bytes`, not `str`, which cannot be `json.dumps`.
2. Even after fixing the problem (by `.decode('ascii')`), we will found that `resp.text` is something like

```
<html><title>500: Internal Server Error</title><body>500: Internal Server Error</body></html>
```

rather than the serialized exception information. Not sure if this limitation is a bug or just by-design.

## Related issue number

#415

### DIFF
--- a/mars/scheduler/mutable.py
+++ b/mars/scheduler/mutable.py
@@ -13,10 +13,11 @@
 # limitations under the License.
 
 from collections import defaultdict
+import uuid
 
 import numpy as np
 
-from ..utils import log_unhandled
+from ..utils import log_unhandled, tokenize
 from .utils import SchedulerActor
 
 
@@ -46,14 +47,14 @@ class MutableTensorActor(SchedulerActor):
 
     @log_unhandled
     def post_create(self):
-        from ..tensor.expressions.utils import create_fetch_tensor
+        from ..tensor.expressions.utils import create_mutable_tensor
 
         super(MutableTensorActor, self).post_create()
         self.set_cluster_info_ref()
-        self._tensor = create_fetch_tensor(self._chunk_size, self._shape, self._dtype)
+        self._tensor = create_mutable_tensor(self._name, self._chunk_size, self._shape, self._dtype)
 
     def tensor_meta(self):
-        return self._shape, self._dtype, self._chunk_size, [c.key for c in self._tensor.chunks]
+        return self._shape, str(self._dtype), self._chunk_size, [c.key for c in self._tensor.chunks]
 
     def tensor_key(self):
         return self._tensor.key
@@ -66,6 +67,11 @@ class MutableTensorActor(SchedulerActor):
         raise NotImplementedError
 
     @log_unhandled
+    def write(self, index, value):
+        chunk_records_to_send = self._tensor._do_write(index, value)
+        self._send_chunk_records(chunk_records_to_send)
+
+    @log_unhandled
     def append_chunk_records(self, chunk_records):
         for chunk_key, record_chunk_key in chunk_records:
             self._chunk_map[chunk_key].append(record_chunk_key)
@@ -74,12 +80,41 @@ class MutableTensorActor(SchedulerActor):
     def seal(self):
         from ..worker.seal import SealActor
         self._sealed = True
+
+        # dump current buffers to worker
+        chunk_records_to_send = self._tensor._do_flush()
+        self._send_chunk_records(chunk_records_to_send)
+
+        # consolidate chunks
         for chunk in self._tensor.chunks:
             ep = self.get_scheduler(chunk.key)
             sealer_uid = SealActor.gen_uid(self._session_id, chunk.key)
             sealer_ref = self.ctx.create_actor(SealActor, uid=sealer_uid, address=ep)
-
             sealer_ref.seal_chunk(self._session_id, self._graph_key,
                                   chunk.key, self._chunk_map[chunk.key],
                                   chunk.shape, self._record_type, self._dtype)
-        return self._graph_key, self._tensor.key, self._tensor.id, self.tensor_meta()
+        # return the hex of self._graph_key since UUID is not json serializable.
+        return self._graph_key.hex, self._tensor.key, self._tensor.id, self.tensor_meta()
+
+    @log_unhandled
+    def _send_chunk_records(self, chunk_records_to_send):
+        from ..worker.dispatcher import DispatchActor
+        from ..worker.quota import MemQuotaActor
+        from ..worker.utils import put_chunk
+
+        chunk_records = []
+        for chunk_key, records in chunk_records_to_send.items():
+            record_chunk_key = tokenize(chunk_key, uuid.uuid4().hex)
+            ep = self.get_scheduler(chunk_key)
+            # register quota
+            quota_ref = self.ctx.actor_ref(MemQuotaActor.default_uid(), address=ep)
+            quota_ref.request_batch_quota({record_chunk_key: records.nbytes})
+            # send record chunk
+            dispatch_ref = self.ctx.actor_ref(DispatchActor.default_uid(), address=ep)
+            receiver_uid = dispatch_ref.get_hash_slot('receiver', chunk_key)
+            receiver_ref = self.ctx.actor_ref(receiver_uid, address=ep)
+            put_chunk(self._session_id, record_chunk_key, records, receiver_ref)
+            chunk_records.append((chunk_key, record_chunk_key))
+
+        # register the record chunks
+        self.append_chunk_records(chunk_records)
--- a/mars/scheduler/mutable.py
+++ b/mars/scheduler/mutable.py
@@ -29,7 +29,7 @@ class MutableTensorActor(SchedulerActor):
     def gen_uid(session_id, name):
         return 's:0:mutable-tensor$%s$%s' % (session_id, name)
 
-    def __init__(self, session_id, name, shape, dtype, graph_key, chunk_size=None, *args, **kwargs):
+    def __init__(self, session_id, name, shape, dtype, graph_key, fill_value=None, chunk_size=None, *args, **kwargs):
         super(MutableTensorActor, self).__init__(*args, **kwargs)
         self._session_id = session_id
         self._name = name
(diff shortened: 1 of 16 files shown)


mars/web/api.py
```diff
@@ -178,3 +221,6 @@ def get(self):
 register_web_handler('/api/session/(?P<session_id>[^/]+)/graph/(?P<graph_key>[^/]+)', GraphApiHandler)
 register_web_handler('/api/session/(?P<session_id>[^/]+)/graph/(?P<graph_key>[^/]+)/data/(?P<tileable_key>[^/]+)',
                      GraphDataHandler)
+register_web_handler('/api/session/(?P<session_id>[^/]+)/mutable-tensor', MutableTensorHandler)
+register_web_handler('/api/session/(?P<session_id>[^/]+)/mutable-tensor/write', MutableTensorWriteHandler)
```
wjsi: This API design does not conform to RESTful styles. May change to ``/api/session/(?P<session_id>[^/]+)/mutable-tensor/(?P<name>[^/]+) and use GET / PUT / POST methods to handle data read / write / seal.

sighingnow: Will revise.

sighingnow: hi @wjsi Mutable tensor requires four endpoints: create/get/write/seal. Is it ok to use the following mapping?

+ `POST` for `create`
+ `HEAD` for `get` (is `HEAD` ok here ?)
+ `PUT` for `write`
+ `GET` for `seal`

wjsi: ``GET`` shall not cause any side-effect on the storage. Therefore create and write can be merged into ``PUT`` and the implementation of API decide whether to create while ``POST`` for seal, or ``POST`` with an ``action`` indicating create or seal. If I were the writer of this PR, I would prefer the former solution.

sighingnow: > create and write can be merged into PUT and the implementation of API decide whether to create while POST for seal

Nice suggestion, thanks!

sighingnow: I have revised the endpoint by use `POST` for both `write` and `seal` (easier to distinguish since `write` has body payload and `seal` doesn't have, and `write`'s body payload is raw bytes, not json, making `create` and `write` are harder to distinguish without extra paramter).

Now all four API of mutable tensor share the same HTTP endpoint.

wjsi: Customized headers can also be used to pass string meta data if you do not like query strings.

mars/tensor/expressions/utils.py
```diff
@@ -554,7 +554,15 @@ def create_fetch_tensor(chunk_size, shape, dtype, tensor_key=None, tensor_id=Non
...
-def setitem_as_records(nsplits_acc, output_chunk, value, ts):
+def create_mutable_tensor(name, chunk_size, shape, dtype, chunk_keys=None):
+    from ..core import MutableTensor, MutableTensorData
+    # Construct MutableTensor on the fly.
+    tensor = create_fetch_tensor(chunk_size, shape, dtype, chunk_keys=chunk_keys)
+    return MutableTensor(data=MutableTensorData(_name=name, _op=None, _shape=shape, _dtype=tensor.dtype,
+                                                _nsplits=tensor.nsplits, _key=tensor.key, _chunks=tensor.chunks))
+
+
+def setitem_as_records(nsplits_acc, output_chunk, value, ts, is_scalar):
```
qinxuye: Doc should be updated.

qinxuye:
One more thing, really cool to see the error can be serialized and sent to client, but sadly no test is added for that, could you please try to add some ut?

hekaisheng:
We can move `MutableTensor.Index` out of `MutableTensor` as a unified way to serialize indexes. Will you do it in this PR or do it later in my pr #540 ?

sighingnow:
The exception info return by the http api of web session can be validated in the `self.assertRaises` part of this PR. I will add a standalone test for the exception info as well.

sighingnow:
@hekaisheng I will revise the patch and move the `Index` class out as soon as possible. Do you guys think is `SerializableIndex` a reasonable name for this class? (since we also have a `Index` class in dataframe part).

sighingnow:
+ Add tests for return exception info from web api.
+ Add `mars.tensor.core.mutable_tensor` constructor, I haven't import it in `mars.tensor.__init__` since there seems cycle import, will dig into it as soon as possible.
+ I will rebase to master (to adapt the serialable `Index` part) after #540 merged, will ping reviewers then.

qinxuye:
> * Add tests for return exception info from web api.
> * Add `mars.tensor.core.mutable_tensor` constructor, I haven't import it in `mars.tensor.__init__` since there seems cycle import, will dig into it as soon as possible.
> * I will rebase to master (to adapt the serialable `Index` part) after #540 merged, will ping reviewers then.

Great, I have an idea, can we add a `fill_value` parameter to `create_mutable_tensor`, so that we can initialize some value for the users?

sighingnow:
> Great, I have an idea, can we add a fill_value parameter to create_mutable_tensor, so that we can initialize some value for the users?

Agree. The `initial_value` can a property of `MutableTensorActor` and will be used to create the empty numpy tensor when `seal`. Will do that ASAP.

hekaisheng:
#540 has been merged, you can rebase master now.

sighingnow:
> Better to rebase after #546 merged.

Will do that.

sighingnow:
+ Rebase to master and delete `MutableTensor.index`
+ Implement `fill_value` for mutable tensor constructor.

mars/scheduler/mutable.py
```diff
@@ -74,12 +86,41 @@ def append_chunk_records(self, chunk_records):
...
+        from ..worker.quota import MemQuotaActor
+        from ..worker.utils import put_chunk
+
+        chunk_records = []
+        for chunk_key, records in chunk_records_to_send.items():
+            record_chunk_key = tokenize(chunk_key, uuid.uuid4().hex)
+            ep = self.get_scheduler(chunk_key)
+            # register quota
+            quota_ref = self.ctx.actor_ref(MemQuotaActor.default_uid(), address=ep)
+            quota_ref.request_batch_quota({record_chunk_key: records.nbytes})
```
wjsi: ReceiverActor takes little process memory, and quota request is not needed.

mars/worker/utils.py
```diff
@@ -165,3 +168,33 @@ def build_load_key(graph_key, chunk_key):
     if isinstance(chunk_key, tuple):
         chunk_key = '@'.join(chunk_key)
     return '%s_load_memory_%s' % (graph_key, chunk_key)
+
+
+def put_chunk(session_id, chunk_key, data, receiver_ref):
```
wjsi: This function shall be put in ``transfer.py`` and renamed as ``put_remote_chunk`` as it is not referenced by any code in worker module.

mars/web/api.py
```diff
@@ -175,6 +182,40 @@ def get(self):
...
+    def get(self, session_id, name):
+        try:
+            meta = self.web_api.get_mutable_tensor(session_id, name)
+            self.write(json.dumps(meta))
+        except:  # noqa: E722
+            self._dump_exception(sys.exc_info())
+
+    def post(self, session_id, name):
+        try:
+            # If the request contains no body payload, it is seal, otherwise it is create
```
wjsi: I still think it is better to use a customized header or a query string argument to define the action of ``POST``.

mars/scheduler/mutable.py
sighingnow: The quota is for chunk of records (the `(index, value)` record of write operations) and the record chunk may be spilled, thus the quota is required, IMO.

wjsi: ReceiverActor stores data in plasma_store or disk. This means the cost of **process memory** equals to zero when receiving data from other machines. What's more, we serialize with pyarrow with zero-copy and spill in small chunks (not chunks in Mars), hence the additional memory cost is no more than the size of these chunks. Therefore there is no need requesting for quotas before data transfer.

mars/web/api.py
sighingnow: Fixed. A parameter `action` has been added.

mars/worker/utils.py
sighingnow: Fixed.

mars/scheduler/mutable.py
sighingnow: Fixed. No need to requesting quota now.

qinxuye:
Could you please confirm that the new commits have resolved your comments? @wjsi

Issue #517, opened 2019-07-05
Support join_by of structured arrays (recarray)


sighingnow:
**Is your feature request related to a problem? Please describe.**

We have already supported structured arrays (`numpy.recarray`). It would be great if we can support some advanced method of structured arrays, for example `join_by`.

**Describe the solution you'd like**

Implement [`join_by`](https://www.numpy.org/devdocs/user/basics.rec.html#numpy.lib.recfunctions.join_by).

**Describe alternatives you've considered**

Otherwise, we need a `for-loop` to finish some requirements like `join`.

**Additional context**

qinxuye:
We don't have `mt.recarray` yet, and structured dtype is not fully tested. Thus how about adding support for them first?

PR #539, opened 2019-07-12
Fix assignment of structured data type arrays.

sighingnow:
## What do these changes do?

Current code from the master branch suffers a trouble about `setitem` on structured arrays, that is, if we have `a = mt.ones(4, dtype=np.dtype([('a', np.int32), ('b', np.double)]))`, execute `a` is ok and the result is consistent with numpy. However, if we `setitem` to a like `a[0] = 100`, numpy will set `a[0]` to `(100, 100)`, mars raises an exception instead:

```python
~/mars/mars/tensor/execution/indexing.py in _index_set_value(ctx, chunk)
    171     if hasattr(input_, 'flags') and not input_.flags.writeable:
    172         input_.setflags(write=True)
--> 173     input_[tuple(indexes)] = value
    174     ctx[chunk.key] = input_
    175

ValueError: setting an array element with a sequence.
```

The issue is caused by

https://github.com/mars-project/mars/blob/ea2db4c2b95ad8ad97d2490edbf378c0d041618d/mars/tensor/expressions/indexing/setitem.py#L110-L113

Here we do `a.dtype.type(value)`, but it `a` is a record type, it also returns `void(b'\x00')`. Thus the exception.

After further thinking, I found that we don't need to cast the type of scalar type and the underlying `setitem` of `numpy/cupy` will take care of it. When the dtype is record type and the value is a tuple, we should treat it as a _scalar_ value (of type `a.dtype`) as well, such behaviour is consistent with numpy.

This pr fixes the above exception and make `setitem` available to structured arrays.

## Related issue number

This pr is related to #517, but doesn't fix that.

### DIFF
--- a/mars/tensor/execution/tests/test_index_execute.py
+++ b/mars/tensor/execution/tests/test_index_execute.py
@@ -242,6 +242,29 @@ class Test(unittest.TestCase):
         raw[idx] = replace
         np.testing.assert_array_equal(res[0], raw)
 
+    def testSetItemStructuredExecution(self):
+        rec_type = np.dtype([('a', np.int32), ('b', np.double), ('c', np.dtype([('a', np.int16), ('b', np.int64)]))])
+
+        raw = np.zeros((4, 5), dtype=rec_type)
+        arr = tensor(raw.copy(), chunk_size=3)
+
+        arr[1:4, 1] = (3, 4., (5, 6))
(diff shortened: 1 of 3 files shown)


qinxuye:
Excellent work, I think this PR can be backported to v0.1, could you please send a backport PR for this? Thanks.

Issue #453, opened 2019-06-14
Merge expression and execution directories into one both for tensor and DataFrame


qinxuye:
Expression describes how an operand should be, and how it can tile tileable objects into chunks. Execution represents how an operand(currently it's chunk) should be executed for the runtime. They are clear.

But new guys come across Mars may be a little confused about their meaning. In my opinion, they can actually be merged into one. For instance, current directory architecture is 

```
|
|- tensor
   |- execution
   |- expressions
       |- arithmetic
       |- base
       |- ...
|- dataframe
   |- execution
   |- expressions
      |- arithmetic
      |- datasource
      |- ...
```

They will be:

```
|
|- tensor
   |- arithmetic
   |- base
   |- ...
|- dataframe
   |- arithmetic
   |- datasource
   |- ...
```

As for `Operand`,  except for `new_tileables`, `new_tileable`, `new_chunks`, `new_chunk`, `tile`, a few methods should be added:

* `execute_cpu(ctx, op)`: execute an operand on CPU.
* `execute_cuda(ctx, op)`: execute an operand on NV CUDA.
* `execute_sparse(ctx, op`: execute a sparse operand.
* `execute(ctx, op)`: the entry function for an operand execution.
* `estimate_size(ctx, op)`: estimate the memory usage for given operand.

`Executor` should be modified either. As I mentioned, currently the execution part is based on chunk instead of operand which look pretty weird. For an operand which has multiple outputs, the `Executor` will execute one output, and must prevent from executing other outputs. If the execution is operand-based, It would be more natural.

Besides, we can try to merge the work from #267 , we need more unittests to cover all expressions and execution parts.

PR #543, opened 2019-07-15
Make executor compatible with old and new execution registration mechanism

qinxuye:
## What do these changes do?

We plan to do a huge refactor described in #453 that `expressions` and `execution` directories in tensor and dataframe would be merged into one, and our execution function will apply on operand instead of chunk.

In the past, the execution registration mechanism is that:

1. If a function is registered to execute a chunk, call it.
2. If the function not found, scan all the functions, if some function is registered for a operand class, and the class is superclass of the class of destination chunk's operand, register the function for the operand's class, and then call it.

Now, the mechanism would be:

1. If a function is registered to execute an **operand**, call it.
2. If the operand has a method `execute`, call the method.
3. Same as point 2 in the past.

This PR is just a temporary PR which provides a compatible way for both old and new registration mechanism. After refactoring finishes, we will abandon the old-fashioned way completely.

## Related issue number

#453

### DIFF
--- a/mars/executor.py
+++ b/mars/executor.py
@@ -261,7 +261,7 @@ class GraphExecution(object):
             # note that currently execution is the chunk-level
             # so we pass the first operand's first output to Executor.handle
             first_op = ops[0]
-            Executor.handle(first_op.outputs[0], results, self._mock)
+            Executor.handle(first_op, results, self._mock)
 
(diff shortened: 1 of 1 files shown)


qinxuye:
cc @hekaisheng @wdkwyf

Issue #451, opened 2019-06-13
Remove all the constant operands


hekaisheng:
**Is your feature request related to a problem? Please describe.**
Now in our tensor operands,  constant operand is a type of special binary operand that one of the inputs is a scalar. However, after series refactors of operands, the binary operand itself could handle this situation and there is no need to keep constant operands.

**Describe the solution you'd like**
We should provide a unified binary operand class that could handle different types of inputs properly.

PR #553, opened 2019-07-17
Move arithmetic out of expressions and remove all constant operands

hekaisheng:
## What do these changes do?
Part of the refactor of merging expression and execution directories, and remove all constant operands.

## Related issue number
 #453  
 #451


mars/tensor/__init__.py
```diff
@@ -24,6 +24,7 @@
     expand_dims, rollaxis, swapaxes, moveaxis, ravel, atleast_1d, atleast_2d, atleast_3d, argwhere, \
     array_split, split, hsplit, vsplit, dsplit, roll, squeeze, ptp, diff, ediff1d, digitize, \
     average, cov, corrcoef, flip, flipud, fliplr, repeat, tile, isin
+from . import arithmetic
```
qinxuye: I don't think this import is required.

mars/tensor/arithmetic/abs.py
```diff
@@ -0,0 +1,72 @@
...
+from .utils import arithmetic_operand
+
+
+@arithmetic_operand(sparse_mode='unary')
+class TensorAbs(TensorUnaryOp):
+    _op_type_ = OperandDef.ABS
+    _handler_name = 'abs'
+
+    @property
+    def handler_name(self):
```
qinxuye: In my opinion, `fun_name` seems a better name.

qinxuye: And also, I think because `execute` etc is a classmethod of Operand, so the property of `fun_name` is unnecessary, just `cls._fun_name` seems pretty acceptable to me.

PR #564, opened 2019-07-19
[BACKPORT] Fix unify chunks when add a scalar with out (#542)

hekaisheng:
Backport of #542


qinxuye:
TileDB seems updated, maybe we should fix it first.

qinxuye:
I have tried to solve tiledb issue in #403 , thus this PR may wait for the backport of it.

Issue #514, opened 2019-07-05
Complete support for DataFrame Add


hekaisheng:
**Is your feature request related to a problem? Please describe.**
As a experimental operation for DataFrame, `DataFrameAdd` has support adding two DataFrames (#351) and implement `radd` in #506 .

A complete support also including:

- [x] add a scalar
- [ ] add a tensor
- [ ] add a series

NJUPole:
See #451 , the implementation of adding a scalar may be different from tensor, `AddConstant` op is not needed.

PR #582, opened 2019-07-25
Support add scalar with dataframe

wdkwyf:
## What do these changes do?
Support add a scalar with the dataframe

## Related issue number
#514

### DIFF
--- a/mars/dataframe/arithmetic/add.py
+++ b/mars/dataframe/arithmetic/add.py
@@ -80,7 +80,7 @@ def add(df, other, axis='columns', level=None, fill_value=None):
     if isinstance(other, DATAFRAME_TYPE) or np.isscalar(other):
         op = DataFrameAdd(axis=axis, level=level, fill_value=fill_value, lhs=df, rhs=other)
     else:
-        return NotImplementedError('Only support add with dataframe or scalar!')
+        raise NotImplementedError('Only support add with dataframe or scalar!')
     return op(df, other)
 
 
@@ -89,5 +89,5 @@ def radd(df, other, axis='columns', level=None, fill_value=None):
     if isinstance(other, DATAFRAME_TYPE) or np.isscalar(other):
         op = DataFrameAdd(axis=axis, level=level, fill_value=fill_value, lhs=df, rhs=other)
     else:
-        return NotImplementedError('Only support add with dataframe or scalar!')
+        raise NotImplementedError('Only support add with dataframe or scalar!')
     return op.rcall(df, other)
--- a/mars/dataframe/arithmetic/add.py
+++ b/mars/dataframe/arithmetic/add.py
@@ -19,6 +19,7 @@ from ...serialize import AnyField, Float64Field
 from ...utils import classproperty
 from ..operands import DataFrameOperand
 from .core import DataFrameBinOpMixin
+import numpy as np
 
 
 class DataFrameAdd(DataFrameOperand, DataFrameBinOpMixin):
--- a/mars/dataframe/arithmetic/add.py
+++ b/mars/dataframe/arithmetic/add.py
@@ -16,7 +16,7 @@ import operator
 import numpy as np
 
 from ... import opcodes as OperandDef
-from ...serialize import AnyField, Float64Field, KeyField
+from ...serialize import AnyField, Float64Field
 from ...utils import classproperty
 from ..operands import DataFrameOperand
 from ..utils import wrap_exception
@@ -29,8 +29,8 @@ class DataFrameAdd(DataFrameOperand, DataFrameBinOpMixin):
     _axis = AnyField('axis')
     _level = AnyField('level')
     _fill_value = Float64Field('fill_value')
-    _lhs = KeyField('lhs')
-    _rhs = KeyField('rhs')
+    _lhs = AnyField('lhs')
+    _rhs = AnyField('rhs')
     _func_name = 'add'
 
     def __init__(self, axis=None, level=None, fill_value=None, object_type=None, lhs=None, rhs=None, **kw):
--- a/mars/dataframe/arithmetic/add.py
+++ b/mars/dataframe/arithmetic/add.py
@@ -68,20 +68,20 @@ class DataFrameAdd(DataFrameOperand, DataFrameBinOpMixin):
     def _set_inputs(self, inputs):
         super(DataFrameAdd, self)._set_inputs(inputs)
         if len(self._inputs) == 2:
-            self._lhs = inputs[0]
-            self._rhs = inputs[1]
+            self._lhs = self._inputs[0]
+            self._rhs = self._inputs[1]
         else:
             if isinstance(self._lhs, DATAFRAME_TYPE):
-                self._lhs = inputs[0]
+                self._lhs = self._inputs[0]
             elif np.isscalar(self._lhs):
-                self._rhs = inputs[0]
+                self._rhs = self._inputs[0]
 
 
 def add(df, other, axis='columns', level=None, fill_value=None):
     if isinstance(other, DATAFRAME_TYPE) or np.isscalar(other):
         op = DataFrameAdd(axis=axis, level=level, fill_value=fill_value, lhs=df, rhs=other)
     else:
-        raise NotImplementedError('Only support add with dataframe or scalar!')
+        return NotImplementedError('Only support add with dataframe or scalar!')
     return op(df, other)
 
 
--- a/mars/dataframe/arithmetic/add.py
+++ b/mars/dataframe/arithmetic/add.py
@@ -16,10 +16,12 @@ import operator
 import numpy as np
 
 from ... import opcodes as OperandDef
-from ...serialize import AnyField, Float64Field
+from ...serialize import AnyField, Float64Field, KeyField
 from ...utils import classproperty
 from ..operands import DataFrameOperand
-from .core import DataFrameBinOpMixin
+from ..utils import wrap_exception
+from .core import DataFrameBinOpMixin, DATAFRAME_TYPE
+
 
 class DataFrameAdd(DataFrameOperand, DataFrameBinOpMixin):
     _op_type_ = OperandDef.ADD
@@ -31,7 +33,7 @@ class DataFrameAdd(DataFrameOperand, DataFrameBinOpMixin):
     _rhs = KeyField('rhs')
     _func_name = 'add'
 
-    def __init__(self, axis=None, level=None, fill_value=None, object_type=None, **kw):
+    def __init__(self, axis=None, level=None, fill_value=None, object_type=None, lhs=None, rhs=None, **kw):
         super(DataFrameAdd, self).__init__(_axis=axis, _level=level,
                                            _fill_value=fill_value,
                                            _object_type=object_type, _lhs=lhs, _rhs=rhs,
@@ -54,10 +56,6 @@ class DataFrameAdd(DataFrameOperand, DataFrameBinOpMixin):
         return self._fill_value
 
     @property
-    def scalar_value(self):
-        return self._scalar_value
-
-    @property
     def lhs(self):
         return self._lhs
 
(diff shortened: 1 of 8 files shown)


mars/dataframe/execution/tests/test_arithmetic_execution.py
```diff
@@ -256,6 +256,14 @@ def testAddWithMultiForms(self):
...
         pd.testing.assert_frame_equal(expected, result)
 
+    def testAddScalar(self):
+        # test add a dataframe with scalar
+        pdf = pd.DataFrame(np.random.rand(10, 10))
+        df = from_pandas(pdf, chunk_size=2)
+        expected = pdf + 1
+        result = self.executor.execute_dataframe(add(df, 1), concat=True)[0]
+        pd.testing.assert_frame_equal(expected, result)
+
```
qinxuye: Could add a test for `radd`.

mars/dataframe/expressions/arithmetic/add.py
```diff
@@ -19,6 +19,7 @@
 from ....utils import classproperty
 from ..core import DataFrameOperand
 from .core import DataFrameBinOpMixin
+import numpy as np
```
qinxuye: Basically, we have the habits.

```python
import builtin_module

import third_party_module

from ... import x
from .. import y
from . import z
```
First part are the builtin modules, second part are third-party libraries. For the third part, we'd like to sort imports from farther modules to closer ones.

mars/dataframe/expressions/arithmetic/add.py
```diff
@@ -27,11 +28,12 @@ class DataFrameAdd(DataFrameOperand, DataFrameBinOpMixin):
     _axis = AnyField('axis')
     _level = AnyField('level')
     _fill_value = Float64Field('fill_value')
+    _scalar_value = Float64Field('scalar_value')
```
qinxuye: I think this field is unnecessary. The `inputs` is enough. We can call `pd.add(*op.inputs)` in execution, what we need do is to check the legality.

mars/dataframe/expressions/arithmetic/add.py
```diff
@@ -49,12 +51,22 @@ def level(self):
...
 
+    @property
+    def scalar_value(self):
+        return self._scalar_value
+
 
 def add(df, other, axis='columns', level=None, fill_value=None):
-    op = DataFrameAdd(axis=axis, level=level, fill_value=fill_value)
+    if np.isscalar(other):
+        op = DataFrameAdd(axis=axis, level=level, fill_value=fill_value, scalar_value=other)
```
qinxuye: Like I said, we don't need this branch actually.

qinxuye: Instead, we need to check the type of `other`, make sure it's a `DataFrameType` or a scalar.

mars/dataframe/expressions/arithmetic/core.py
```diff
@@ -547,22 +547,43 @@ def _tile_both_dataframes(cls, op):
...
+
+        return new_op.new_dataframes(op.inputs, df.shape, nsplits=left.nsplits, dtypes=df.dtypes,
+                                     index_value=df.index_value, columns_value=df.columns, chunks=out_chunks)
+
     @classmethod
     def tile(cls, op):
         if all(isinstance(inp, DATAFRAME_TYPE) for inp in op.inputs):
             return cls._tile_both_dataframes(op)
-
+        if isinstance(op.inputs[0], DATAFRAME_TYPE) and np.isscalar(op.inputs[1]):
```
qinxuye: Use `elif`.

mars/dataframe/expressions/arithmetic/core.py
```diff
@@ -547,22 +547,43 @@ def _tile_both_dataframes(cls, op):
...
+            return cls._tile_scalar(op)
         raise NotImplementedError
 
     @classproperty
     def _operator(self):
         raise NotImplementedError
 
     @classmethod
-    def _calc_properties(cls, x1, x2):
```
qinxuye: When would `x2` be None?

mars/dataframe/expressions/arithmetic/core.py
```diff
@@ -547,22 +547,43 @@ def _tile_both_dataframes(cls, op):
...
+    def _calc_properties(cls, x1, x2=None):
+        if x2 is None or np.isscalar(x2):
```
qinxuye: For `1 + df`, x1 will be scalar, so either `x1` or `x2` could be scalar.

mars/dataframe/expressions/arithmetic/core.py
```diff
@@ -610,7 +631,11 @@ def _call(self, x1, x2):
             shape = kw.pop('shape', None)
             return self.new_dataframe([x1, x2], shape, **kw)
         else:
-            raise NotImplementedError('Only support add two dataframes for now')
+            setattr(self, '_object_type', ObjectType.dataframe)
```
qinxuye: Line 634-638 is identical to line 629-632.

mars/dataframe/expressions/tests/test_arithmetic.py
```diff
@@ -654,3 +655,11 @@ def testAbs(self):
             self.assertEqual(c2.index, c1.index)
             pd.testing.assert_index_equal(c2.columns.to_pandas(), c1.columns.to_pandas())
             pd.testing.assert_index_equal(c2.index_value.to_pandas(), c1.index_value.to_pandas())
+
+    def testAddScalar(self):
```
qinxuye: Add test for `radd`.

mars/dataframe/expressions/arithmetic/add.py
wdkwyf: hi, do you mean `chunk.op.inputs`(before refactor). I found that input only include the dataframe part.

qinxuye: OK, that's a problem, thus we can introduce the `_lhs` and `_rhs` fields in tensor, by calling `pd.add(op.lhs, op.rhs)` could look better.

mars/dataframe/expressions/arithmetic/core.py
wdkwyf: hi, Could I swap x1 and x2（in `rcall`） to ensure x2 is always the scalar?

qinxuye: For `add`, it's ok, but we have to consider that this is fundamental part of all arithmetic operations, e.g. `divide`.

mars/dataframe/expressions/arithmetic/core.py
wdkwyf: When it's called from `_new_chunks`. Because in `_tile_scalar`, I only create the chunk of dataframe (not include scalar). Is that right?

mars/dataframe/execution/arithmetic.py
```diff
@@ -108,14 +109,22 @@ def _index_align_reduce(ctx, chunk):
...
+        left = ctx[chunk.op.lhs.key]
+    except AttributeError:
+        left = chunk.op.lhs
+
+    try:
+        right = ctx[chunk.op.rhs.key]
+    except AttributeError:
+        right = chunk.op.rhs
+        # if right is scalar
+        ctx[chunk.key] = pd.DataFrame.add(left, right, axis=chunk.op.axis,
```
qinxuye: Duplicate line with 126.

mars/dataframe/execution/tests/test_arithmetic_execution.py
```diff
@@ -257,13 +263,17 @@ def testAddWithMultiForms(self):
         pd.testing.assert_frame_equal(expected, result)
 
     def testAddScalar(self):
-        # test add a dataframe with scalar
+        # test dataframe + scalar
         pdf = pd.DataFrame(np.random.rand(10, 10))
         df = from_pandas(pdf, chunk_size=2)
         expected = pdf + 1
```
qinxuye: Could add a test for 1 + pdf.

mars/dataframe/expressions/arithmetic/add.py
```diff
@@ -55,18 +58,37 @@ def fill_value(self):
...
+        return self._lhs
+
+    @property
+    def rhs(self):
+        return self._rhs
+
+    def _set_inputs(self, inputs):
+        super(DataFrameAdd, self)._set_inputs(inputs)
+        if len(self._inputs) == 2:
+            self._lhs = inputs[0]
```
qinxuye: Let `self._lhs = self._inputs[0]` seem better, because in super's `_set_inputs`, the entity will be transformed into data.

mars/dataframe/expressions/arithmetic/add.py
```diff
@@ -55,18 +58,37 @@ def fill_value(self):
...
+
 
 def add(df, other, axis='columns', level=None, fill_value=None):
-    if np.isscalar(other):
-        op = DataFrameAdd(axis=axis, level=level, fill_value=fill_value, scalar_value=other)
+    if isinstance(other, DATAFRAME_TYPE) or np.isscalar(other):
+        op = DataFrameAdd(axis=axis, level=level, fill_value=fill_value, lhs=df, rhs=other)
     else:
-        op = DataFrameAdd(axis=axis, level=level, fill_value=fill_value)
+        raise NotImplementedError('Only support add with dataframe or scalar!')
```
qinxuye: The `__add__` method could be a bit different, it must catch `NotImplementedError` and return `NotImplemented`.

mars/dataframe/expressions/arithmetic/core.py
```diff
@@ -630,9 +638,14 @@ def _call(self, x1, x2):
             kw = self._calc_properties(x1, x2)
             shape = kw.pop('shape', None)
             return self.new_dataframe([x1, x2], shape, **kw)
-        else:
+        elif np.isscalar(x1) or np.isscalar(x2):
             setattr(self, '_object_type', ObjectType.dataframe)
-            kw = self._calc_properties(x1, x2)
+            if np.isscalar(x1):
```
qinxuye: Can we simplify this logic a bit?

mars/dataframe/expressions/tests/test_arithmetic.py
```diff
@@ -660,6 +660,11 @@ def testAddScalar(self):
         data = pd.DataFrame(np.random.rand(10, 10), index=np.arange(10),
                             columns=np.arange(3, 13))
         df = from_pandas(data, chunk_size=5)
+        # test add with scalar
         result = add(df, 1)
+        # test radd with scalar
+        result2 = df.radd(1)
```
qinxuye: We can test `df.add(1)` as well as `df + 1`, so does `radd`.

qinxuye:
Please bump version to 0.3.0a1 as I think this would be the first PR for v0.3.

mars/dataframe/execution/arithmetic.py
wdkwyf: line 26 has swap the location of `left` and `right`. Because pd.DataFrame.add() must have the first parameter be the dataframe.

mars/dataframe/arithmetic/add.py
```diff
@@ -27,13 +29,15 @@ class DataFrameAdd(DataFrameOperand, DataFrameBinOpMixin):
     _axis = AnyField('axis')
     _level = AnyField('level')
     _fill_value = Float64Field('fill_value')
-
+    _lhs = KeyField('lhs')
```
qinxuye: I think it should be `AnyField`.

mars/dataframe/utils.py
```diff
@@ -431,3 +431,13 @@ def filter_index_value(index_value, min_max, store_data=False):
...
 
     return parse_index(pd_index[f], store_data=store_data)
+
+
+def wrap_exception(func):
+    def wrapper(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except NotImplementedError as e:
+            print(e)
```
qinxuye: Seems uncovered here, should return NotImplemented instead, try to write ut to cover this line.

PR #587, opened 2019-07-29
Fix numpy and pandas version for requirements

qinxuye:
## What do these changes do?

Fix pandas version to 0.24.2 in requirements-wheel.txt, and fix requirements for the reason that newest version of numpy and pandas drop support for Python 2.7.

## Related issue number

NA


qinxuye:
Replaced by #588 .

PR #598, opened 2019-08-01
Move optimizes out of tensor to make it universal.

### LINKED ISSUE #596: Move optimizes out of tensor module
We need a universal optimizer for both tensor and DataFrame(#593) , as a first step, move optimize directory out of tensor module.

wdkwyf:
## What do these changes do?
Move `optimizes` module out from `tensor`  module to make it universal.

## Related issue number
resolve #596


qinxuye:
I think this PR can be backported to v0.2 .

PR #600, opened 2019-08-01
Support create mars session with given session_id.

### LINKED ISSUE #605: Support create session with specified `session_id`
Support using given `session_id` to create `LocalClusterSession` or `WebSession`.

sighingnow:
## What do these changes do?

Support create mars session on the fly using given `session_id`, and do necessary checks. Available for `LocalClusterSession` and `WebSession`.

## Related issue number

resolve #605

### DIFF
--- a/mars/api.py
+++ b/mars/api.py
@@ -70,11 +70,11 @@ class MarsAPI(object):
     def delete_session(self, session_id):
         self.session_manager.delete_session(session_id)
 
-    def check_session(self, session_id):
+    def has_session(self, session_id):
         '''
         Check if the session with given session_id exists.
         '''
-        return self.session_manager.check_session(session_id)
+        return self.session_manager.has_session(session_id)
 
     def submit_graph(self, session_id, serialized_graph, graph_key, target,
                      compose=True, wait=True):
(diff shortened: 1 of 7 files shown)


mars/api.py
```diff
@@ -70,6 +70,12 @@ def create_session(self, session_id, **kw):
     def delete_session(self, session_id):
         self.session_manager.delete_session(session_id)
 
+    def check_session(self, session_id):
```
hekaisheng: how about rename to `has_session`?

mars/scheduler/session.py
```diff
@@ -226,6 +226,10 @@ def create_session(self, session_id, **kwargs):
         self._session_refs[session_id] = session_ref
         return session_ref
 
+    @log_unhandled
+    def check_session(self, session_id):
```
hekaisheng: Ditto.

mars/api.py
sighingnow: Renamed.

sighingnow:
Update:

+ Rename `check_session` to `has_session`
+ Add unit test cases

PR #601, opened 2019-08-01
Fix mutable tensor related bugs and some enhancement

### LINKED ISSUE #603: [BUG] Mutable tensor use wrong worker selection method
**Describe the bug**

The current mutable tensor implementation use `get_sheduler(chunk_key)` to select a _worker_ to place the chunk, which is totally WRONG! Which returns address of scheduler, not the worker.

The correct way to select a worker for chunk should be randomly choice a worker endpoint among all available workers.

**Expected behavior**

Chunks should be placed on workers, not on schedulers.

sighingnow:
## What do these changes do?

Fix the mutable tensor bugs and some enhancements:

+ Fix bug about worker selection for chunks
+ Support write mutable tensors using raw payload (for example, arrow tensors and arrow record batches), for better support for other languages.
+ Add an enhanced test for mutable tensor.

## Related issue number

Fixes #603.

### DIFF
--- a/mars/scheduler/mutable.py
+++ b/mars/scheduler/mutable.py
@@ -68,7 +68,7 @@ class MutableTensorActor(SchedulerActor):
         self._tensor = MutableTensor(data=MutableTensorData(
             _name=self._name, _op=None, _shape=self._shape, _dtype=tensor.dtype,
             _nsplits=tensor.nsplits, _key=tensor.key, _chunks=tensor.chunks,
-            _chunk_eps=list(self._chunk_ep_map.values())))
+            _chunk_eps=[self._chunk_ep_map[c.key] for c in tensor.chunks]))
 
     def tensor_meta(self):
         # avoid built-in scalar dtypes are made into one-field record type.
@@ -76,7 +76,7 @@ class MutableTensorActor(SchedulerActor):
             dtype_descr = self._dtype.descr
         else:
             dtype_descr = str(self._dtype)
-        chunk_keys = list(self._chunk_ep_map.keys())
+        chunk_keys = [c.key for c in self._tensor.chunks]
         chunk_eps = [self._chunk_ep_map[key] for key in chunk_keys]
         return self._shape, dtype_descr, self._chunk_size, chunk_keys, chunk_eps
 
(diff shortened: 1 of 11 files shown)


hekaisheng:
Could you please open an issue to describe the bug fixed in this PR ?

mars/web/server.py
```diff
@@ -141,7 +142,23 @@ def write_mutable_tensor(self, session_id, name, body):
...
+            value = dataserializer.loads(body[8+index_json_size:], raw=False)
+        elif payload_type == 'tensor':
+            tensor_chunk_offset = 8 + index_json_size
+            with pyarrow.BufferReader(body[tensor_chunk_offset:]) as reader:
+                value = pyarrow.read_tensor(reader).to_numpy()
+        elif payload_type == 'record_batch':
+            schema_size = np.asscalar(np.frombuffer(body[8+index_json_size:8+index_json_size+8], dtype=np.int64))
+            schema_offset = 8 + index_json_size + 8
+            with pyarrow.BufferReader(body[schema_offset:schema_offset+schema_size]) as reader:
+                schema = pyarrow.read_schema(reader)
```
qinxuye: Does schema required each time we write a record batch?

sighingnow: It is more convenient to include schema in every batch, otherwise, we may need to record it somewhere. Compared to the size of payload data, the size of schema is ignorable, IMHO.

mars/scheduler/mutable.py
```diff
@@ -44,23 +44,41 @@ def __init__(self, session_id, name, shape, dtype, graph_key, fill_value=None, c
...
 
         super(MutableTensorActor, self).post_create()
         self.set_cluster_info_ref()
-        self._tensor = create_mutable_tensor(self._name, self._chunk_size, self._shape, self._dtype)
+        self._resource_ref = self.ctx.actor_ref(ResourceActor.default_uid())
+
+        workers = [k for k in self._resource_ref.get_workers_meta()]
+
+        tensor = create_fetch_tensor(self._chunk_size, self._shape, self._dtype)
+        self._chunk_ep_map = dict((c.key, str(np.random.choice(workers))) for c in tensor.chunks)
```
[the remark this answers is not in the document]
sighingnow: Added to `ResourceActor`.

PR #611, opened 2019-08-02
Extract scheduler discoverer from cluster_info

### LINKED ISSUE #609: [ENH] Extract scheduler discoverer in cluster info
In ``cluster_info.py``, scheduler discovery is created with two options passed as args, namely ``schedulers`` and ``service_discover_addr``. Under more complicated deployment scenarios, these two arguments are far from sufficient. For instance, when we need to run Mars under Kubernetes and get information of the cluster from Kubernetes API. Therefore we need to extract these two args and create an abstraction called ``Discoverer`` which provides ``get`` and ``watch`` api for scheduler discovery.

wjsi:
## What do these changes do?

Add ``StaticSchedulerDiscoverer`` and ``KVStoreSchedulerDiscoverer`` to handle initializing services with static scheduler lists and kv-store respectively.

## Related issue number

Fixes #609

### DIFF
--- a/mars/scheduler/__main__.py
+++ b/mars/scheduler/__main__.py
@@ -34,15 +34,25 @@ class SchedulerApplication(BaseApplication):
         parser.add_argument('--nproc', help='number of processes')
 
     def create_pool(self, *args, **kwargs):
-        self._service = SchedulerService(
-            advertise_addr=self.args.advertise,
-        )
+        self._service = SchedulerService()
         self.n_process = int(self.args.nproc or resource.cpu_count())
         kwargs['distributor'] = MarsDistributor(self.n_process, 's:h1:')
         return super(SchedulerApplication, self).create_pool(*args, **kwargs)
 
+    def create_scheduler_discoverer(self):
+        advertise_endpoint = self.args.advertise or self.endpoint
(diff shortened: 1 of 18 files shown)


mars/scheduler/__main__.py
```diff
@@ -34,15 +34,25 @@ def config_args(self, parser):
...
-        )
+        self._service = SchedulerService()
         self.n_process = int(self.args.nproc or resource.cpu_count())
         kwargs['distributor'] = MarsDistributor(self.n_process, 's:h1:')
         return super(SchedulerApplication, self).create_pool(*args, **kwargs)
 
+    def create_scheduler_discoverer(self):
+        advertise_endpoint = self.args.advertise or self.endpoint
+        if ':' not in advertise_endpoint:
+            advertise_endpoint += ':' + self.endpoint.rsplit(':', 1)[-1]
```
qinxuye: Is it confirmed that ':' is in `self.endpoint`?

wjsi: ``create_scheduler_discoverer`` is called after the actor pool is created, and ``self.endpoint`` is generated by code then, not user input.

qinxuye:
Shall this PR be backported?

PR #624, opened 2019-08-07
Fix `infer_index_value` when one of input indexes is `RangeIndex`

### LINKED ISSUE #621: [BUG] infer_index_value is wrong
**Describe the bug**

We have implemented `infer_index_value` as follows, noticing the `out_index = operator(left_index, right_index)` part, it doesn't make any sence to perform `operator` on `index_value`, since operators like `+-*/` are _element-wise_ on `Index` of pandas.

A trivial example to trigger the problem:

```python
df1 = pd.DataFrame(np.arange(10).reshape(5, 2), index=range(5))
df2 = pd.DataFrame(np.arange(10).reshape(5, 2), index=np.arange(5))

x = add(from_pandas(df1), from_pandas(df2))
x.tiles()
```

https://github.com/mars-project/mars/blob/e6594f0b9232b71ac3e6860ed6ca56a72a16f308/mars/dataframe/arithmetic/utils.py#L33-L52

**To Reproduce**
1. Your Python version
4. Full stack of the error.

**Expected behavior**

Infer the correct value for `index_value`.

**Additional context**

hekaisheng:
## What do these changes do?
Use `pd.RangeIndex(0)` to infer the out index if one of the input indexes is `RangeIndex`

## Related issue number

Fixes #621

### DIFF
--- a/mars/dataframe/arithmetic/utils.py
+++ b/mars/dataframe/arithmetic/utils.py
@@ -12,12 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import operator
+import pandas as pd
 
-try:
-    import pandas as pd
-except ImportError:  # pragma: no cover
-    pass
+import operator
 
 from ..utils import build_empty_df, parse_index
 from ..core import IndexValue
@@ -46,7 +43,11 @@ def infer_index_value(left_index_value, right_index_value, operator):
         return left_index_value
 
     left_index = left_index_value.to_pandas()
+    if isinstance(left_index, pd.RangeIndex):
+        left_index = pd.RangeIndex(0)
     right_index = right_index_value.to_pandas()
+    if isinstance(right_index, pd.RangeIndex):
+        right_index = pd.RangeIndex(0)
     out_index = operator(left_index, right_index)
     key = tokenize(left_index_value.key, right_index_value.key, operator.__name__)
     return parse_index(out_index, key=key)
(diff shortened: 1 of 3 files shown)


mars/dataframe/arithmetic/utils.py
```diff
@@ -46,7 +43,11 @@ def infer_index_value(left_index_value, right_index_value, operator):
         return left_index_value
 
     left_index = left_index_value.to_pandas()
+    if isinstance(left_index, pd.RangeIndex):
+        left_index = pd.RangeIndex(0)
     right_index = right_index_value.to_pandas()
+    if isinstance(right_index, pd.RangeIndex):
+        right_index = pd.RangeIndex(0)
     out_index = operator(left_index, right_index)
```
sighingnow: I still think it makes no sense to do `out_index = operator(left_index, right_index)`, for example, when one of the index is `DatatimeIndex` and the other is `Float64Index`, the `operator` may cannot be applied.

I recommend to do something like `parse_index(left_index.join(right_index, how='outer'))` to get the correct type of result index, then we don't need to concern any exceptions even when the two indexes are both `RangeIndex` and even they have the same `slice` value.

hekaisheng: Thanks for the suggestions, it is indeed a better way to infer index value.

hekaisheng: Still have some problems:
 ``` Python
In: pd.Int64Index([]).join(pd.Float64Index([]), how='outer')                                                         
Out: Int64Index([], dtype='int64')
```
expected dtype is `float64`, maybe we can use `find_common_type` in `pandas` to infer dtype.

PR #630, opened 2019-08-07
Add `to_gpu` and `to_cpu` support for both tensor and DataFrame

### LINKED ISSUE #577: Operand's `gpu` property should record value if it's deterministic

Currently, `gpu` property will not have value unless its operand is a data source one, and `gpu` is specified. The subsequent operands' `gpu` property would be None.  It's not friendly enough for optimizer because it would not know whether optimize should be cpu or gpu specific. Thus, I suggest that  we should record true or false for `gpu` property when all inputs are on CPU or GPU.

### LINKED ISSUE #580: Add `to_gpu` and `to_cpu` support

For now, we can only specify `gpu` when creating a tensor, when we want to move a tensor to gpu if it's generated by other computation, there is no way.

**Additional context**

The `gpu` property would have value if #577 finished when all inputs are on GPU or CPU.

qinxuye:
## What do these changes do?

Add `to_gpu` and `to_cpu` to tensor and DataFrame.

- [x] `Tensor.to_gpu`
- [x] `Tensor.to_cpu`
- [x] `DataFrame.to_gpu`
- [x] `DataFrame.to_cpu`
- [x] `Series.to_gpu`
- [x] `Series.to_cpu`

## Related issue number

Resolves #580 .
Resolves #577 .


mars/core.py
```diff
@@ -667,6 +667,16 @@ class TileableOperandMixin(object):
     def check_inputs(self, inputs):
         pass
 
+    def _check_if_gpu(self, inputs):
```
wjsi: May change to ``_fill_if_gpu`` as the function changes the attributes of entities.

mars/dataframe/merge/concat.py
```diff
@@ -14,6 +14,10 @@
 
 import pandas as pd
 import numpy as np
+try:
+    import cudf
+except ImportError:  # pragma: no cover
+    cudf = None
```
wjsi: Use ``lazy_import``, as processes forked after ``cudaInit()`` may be problematic.

qinxuye: Done.

mars/core.py
qinxuye: I keep the function name, and return value in it.

PR #628, opened 2019-08-07
Add `searchsorted` support

qinxuye:
## What do these changes do?

This PR implements `numpy.searchsorted` operation for Mars tensor. 

~~One more thing to do is adding order support for `searchsorted`, that would happen once #595 got merged.~~

- [x] Support order.

## Related issue number

Resolves #627

PR #639, opened 2019-08-12
Add PCA decomposition algorithm support

### LINKED ISSUE #513: Implement PCA decomposition algorithm

We can implement PCA decomposition as a part of `mars.learn`. 

We can replicate the implementation in sklearn: https://github.com/scikit-learn/scikit-learn/blob/ec6d0bb2481dd756b7e992659ea0e7f864873916/sklearn/decomposition/pca.py#L104, but cautiously take care of tensor instead of ndarray.

As for the `svd_solver`, we have not parallelized `svds` yet, so we can provide solver including 'auto', 'full', 'randomized'. 

`randomized_svd` is introduced in #388 .

qinxuye:
## What do these changes do?

This PR adds the PCA decomposition algorithm.

TODO:

- [x] Wait for merging of #628 .
- [x] Add more unittests.

## Related issue number

Resolves #513 .

### DIFF
--- a/mars/fuse.pyx
+++ b/mars/fuse.pyx
@@ -19,10 +19,6 @@ from collections import deque
 from .operands import Fuse, VirtualOperand
 
 
-class InvalidComposedNodeError(Exception):
-    pass
-
-
 class Fusion(object):
     def __init__(self, graph):
         self._graph = graph
--- a/mars/fuse.pyx
+++ b/mars/fuse.pyx
@@ -96,13 +96,10 @@ class Fusion(object):
                 return n.composed[-1]
             return n
 
-        observed = set()
-
         composed_nodes = node.composed
         nodes_set = set(composed_nodes)
         tail_node = composed_nodes[-1]
         self._graph.add_node(tail_node)
-        observed.add(tail_node)
         q = deque()
         q.append(tail_node)
         while len(q) > 0:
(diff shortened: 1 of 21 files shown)


mars/fuse.pyx
```diff
@@ -96,30 +96,27 @@ class Fusion(object):
...
                 pre = get_node(pre)
                 if pre in nodes_set:
                     self._graph.add_node(pre)
                 if pre in self._graph:
                     self._graph.add_edge(pre, cur_node)
                 if pre in nodes_set:
                     q.appendleft(pre)
-                    observed.add(pre)
-        if len(observed) != len(nodes_set):
-            raise InvalidComposedNodeError("Invalid composed node data")
```
wjsi: The error definition can be removed when it is not needed.

mars/tensor/datasource/array.py
```diff
@@ -151,9 +152,18 @@ def tensor(data, dtype=None, order='K', chunk_size=None, gpu=None, sparse=False)
         return scalar(data, dtype=dtype)
     elif issparse(data):
         return _from_spmatrix(data, dtype=dtype, chunk_size=chunk_size, gpu=gpu)
+    elif hasattr(data, '__mars_tensor__'):
+        data = data.__mars_tensor__(dtype=dtype)
+        return data.astype(data.dtype, order=order, copy=False)
```
wjsi: Now that we already do ``data.astype`` in ``__mars_tensor__``, we may pass order as an argument as well.

mars/fuse.pyx
qinxuye: Removed already.

mars/tensor/datasource/array.py
qinxuye: Done.

PR #646, opened 2019-08-14
Use numpy only to calculate fft

### LINKED ISSUE #645: We should use numpy.fft only because it is faster than scipy.fftpack
Maybe at a point in the history, scipy's fftpack is faster than numpy's fft module. Thus we do some optimization in the execution of fft that  if scipy is installed, scipy's fftpack would be used to calculate fft. However, recently, I found that numpy's fft is obviously faster than scipy's fftpack. Sample code is shown below.

```python
In [1]: N = 1600000                                                             

In [2]: import numpy as np                                                      

In [3]: a = np.random.rand(N, 10)                                               

In [4]: %timeit np.fft.fft(a)                                                   
118 ms ± 1.33 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [6]: import scipy.fftpack as sfft                                            

In [7]: %timeit sfft.fft(a)                                                     
290 ms ± 3.65 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [9]: np.testing.assert_allclose(np.fft.fft(a), sfft.fft(a))
```

Hence, I suggest to use numpy only to calculate fft.

qinxuye:
## What do these changes do?

Nowadays, `numpy.fft` is way more faster than `scipy.fftpack`, thus in this PR, I remove the part that trying to use `scipy.fftpack` to calculate fft.

## Related issue number

Resolves #645 .

### DIFF
--- a/mars/tensor/fft/core.py
+++ b/mars/tensor/fft/core.py
@@ -21,13 +21,7 @@ from ...serialize import ValueType, KeyField, StringField, Int32Field, \
     Int64Field, ListField
 from ..utils import validate_axis, decide_chunk_sizes, recursive_tile
 from ..operands import TensorHasInput, TensorOperandMixin
-import numpy as np
 from ..array_utils import get_array_module
-
-try:
-    import scipy.fftpack as scifft
-except ImportError:  # pragma: no cover
-    scifft = None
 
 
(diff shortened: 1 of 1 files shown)


qinxuye:
This PR is all about deleting code thus wouldn't effect the result of coverage.

PR #660, opened 2019-08-16
Generate unique key for chunks of mutable tensors which have identical properties

### LINKED ISSUE #659: [BUG]Mutable tensor with identical properties should have different keys
**Describe the bug**

The key of mutable tensor, as well as its chunks, are generated based on the chunk's properties and the tensor's properties. When we create two mutable tensors with the same shape and dtype, we need to generate random keys for their chunks, as well as for the tensors.

sighingnow:
## What do these changes do?

As described in #659, we need to generate unique keys for mutable tensor chunks. This pr add a test that can reproduce the problem.

## Related issue number

Fixes #659

### DIFF
--- a/mars/tensor/utils.py
+++ b/mars/tensor/utils.py
@@ -537,7 +537,6 @@ def create_fetch_tensor(chunk_size, shape, dtype, tensor_key=None, tensor_id=Non
     as well as possible tensor_key, tensor_id and chunk keys.
     """
     from ..config import options
-    from ..utils import tokenize
     from .fetch import TensorFetch
 
     if not isinstance(dtype, np.dtype):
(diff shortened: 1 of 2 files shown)


mars/tensor/utils.py
```diff
@@ -554,9 +556,13 @@ def create_fetch_tensor(chunk_size, shape, dtype, tensor_key=None, tensor_id=Non
     for chunk_shape, chunk_idx, chunk_key in izip(itertools.product(*chunk_size),
                                                   itertools.product(*chunk_size_idxes),
                                                   chunk_keys):
+        if chunk_key is None:
+            chunk_key = tokenize(uuid.uuid4().hex)
```
qinxuye: I think you can just add some extra property to `new_chunk`, e.g. `new_chunk(None, shape=xx, index=xx, hex=uuid.uuid4().hex)`, which would make chunk key different.

mars/tensor/utils.py
```diff
@@ -554,9 +556,13 @@ def create_fetch_tensor(chunk_size, shape, dtype, tensor_key=None, tensor_id=Non
...
         chunk = fetch_op.copy().reset_key().new_chunk(None, shape=chunk_shape, index=chunk_idx,
                                                       _key=chunk_key)
         chunks.append(chunk)
+    if tensor_key is None:
+        tensor_key = tokenize(uuid.uuid4().hex)
```
qinxuye: Ditto.

mars/tensor/utils.py
sighingnow: Fixed.

PR #690, opened 2019-08-22
Ensure remaining data removed on session close

### LINKED ISSUE #689: [BUG] Chunks not freed when session exits
**Describe the bug**
When some expressions are executed in Mars and then the session exits explicitly, some data related to the session is not cleaned up, which makes subsequent sessions slow.

wjsi:
# What do these changes do?

Ensure remaining data removed on session delete by waiting all data removed before destroying GraphActor.

## Related issue number

Fixes #689

### DIFF
--- a/mars/web/session.py
+++ b/mars/web/session.py
@@ -36,7 +36,7 @@ logger = logging.getLogger(__name__)
 
 class Session(object):
     def __init__(self, endpoint, session_id=None, req_session=None, args=None):
-        self._endpoint = endpoint
+        self._endpoint = endpoint.rstrip('/')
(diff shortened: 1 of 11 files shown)


mars/web/session.py
```diff
@@ -309,8 +309,10 @@ def _update_tileable_shape(self, tileable):
         tileable._update_shape(tuple(sum(nsplit) for nsplit in new_nsplits))
         tileable.nsplits = new_nsplits
 
-    def decref(self, *keys):
+    def decref(self, *keys, **kwargs):
```
qinxuye: How about add a `delete` method.

PR #693, opened 2019-08-23
Fetch tileable directly by its key.

### LINKED ISSUE #608: Support fetch tensor/dataframe using only the tileable keys

After #605, we can support the web session and local cluster session to fetch tensor/dataframe directly using the tileable keys.

+ We first check if there are any GraphActor that contains in tileable, (the SessionActor already has such mapping)
+ When the key does not exist, raise an Exception.
+ For web session, return 404

**Additional context**

PR is coming.

sighingnow:
## What do these changes do?

Implement fetching tileable using its key and using `indexes` from `**kw`.

## Related issue number

Closes #608

### DIFF
--- a/mars/api.py
+++ b/mars/api.py
@@ -148,13 +148,13 @@ class MarsAPI(object):
         return state
 
     def fetch_data(self, session_id, graph_key, tileable_key, index_obj=None, compressions=None):
-        if graph_key is None:
+        if graph_key is None or graph_key == 'None':
             session_uid = SessionActor.gen_uid(session_id)
             session_ref = self.get_actor_ref(session_uid)
             graph_key = session_ref.get_graph_key(tileable_key)
         # We still cannot find the GraphActor for the given tileable key.
         if graph_key is None:
-            return None
+            raise GraphNotExists
 
         graph_uid = GraphActor.gen_uid(session_id, graph_key)
         graph_ref = self.get_actor_ref(graph_uid)
--- a/mars/api.py
+++ b/mars/api.py
@@ -148,6 +148,14 @@ class MarsAPI(object):
         return state
 
     def fetch_data(self, session_id, graph_key, tileable_key, index_obj=None, compressions=None):
+        if graph_key is None:
+            session_uid = SessionActor.gen_uid(session_id)
+            session_ref = self.get_actor_ref(session_uid)
+            graph_key = session_ref.get_graph_key(tileable_key)
+        # We still cannot find the GraphActor for the given tileable key.
+        if graph_key is None:
+            return None
+
         graph_uid = GraphActor.gen_uid(session_id, graph_key)
         graph_ref = self.get_actor_ref(graph_uid)
         nsplits, chunk_indexes = graph_ref.get_tileable_meta(tileable_key)
--- a/mars/deploy/local/session.py
+++ b/mars/deploy/local/session.py
@@ -19,6 +19,8 @@ import json
 import time
 from numbers import Integral
 
+import six
+
 from ...api import MarsAPI
 from ...compat import TimeoutError  # pylint: disable=W0622
 from ...scheduler.graph import GraphState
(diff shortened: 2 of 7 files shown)


mars/api.py
```diff
@@ -148,6 +148,14 @@ def get_graph_state(self, session_id, graph_key):
         return state
 
     def fetch_data(self, session_id, graph_key, tileable_key, index_obj=None, compressions=None):
+        if graph_key is None or graph_key == 'None':
```
qinxuye: The 'None' str looks pretty weird here, could the `graph_key` be None only?

mars/deploy/local/session.py
```diff
@@ -148,26 +150,34 @@ def run(self, *tileables, **kw):
         else:
             return self.fetch(*tileables)
 
-    def fetch(self, *tileables):
+    def fetch(self, *tileables, **kw):
```
qinxuye: We must keep consistency about `fetch` for all kinds of sessions. If `kw` is necessary, make sure all `fetch` have the same signature.

mars/deploy/local/session.py
```diff
@@ -148,26 +150,34 @@ def run(self, *tileables, **kw):
...
         from ...tensor.indexing import TensorIndex
         from ...dataframe.indexing import DataFrameIlocGetItem
 
+        fetch_indexes = kw.pop('indexes', ())
```
qinxuye: Check `kw` to make sure no other `kw` comes in accidentally.

mars/deploy/local/session.py
```diff
@@ -148,26 +150,34 @@ def run(self, *tileables, **kw):
...
         tileable_results = []
         for tileable in tileables:
             # TODO: support DataFrame getitem
-            if tileable.key not in self._executed_tileables and \
-                    isinstance(tileable.op, (TensorIndex, DataFrameIlocGetItem)):
-                key = tileable.inputs[0].key
-                indexes = tileable.op.indexes
-                if not all(isinstance(ind, (slice, Integral)) for ind in indexes):
-                    raise ValueError('Only support fetch data slices')
+            if isinstance(tileable, six.string_types):
```
qinxuye: Is it possible to save the executed tileable objects on the cluster side, so that we can check here if the tileable object is executer or not. `session.fetch(tileable.key)` looks unpleasant indeed. 

What's your idea about this? @hekaisheng

mars/api.py
sighingnow: The `graph_key` comes from the URL, where `None` is parsed as `'None'`. It is wired, and I could parse it as `None` in the web api handler....

mars/deploy/local/session.py
sighingnow: Will fix.

sighingnow:
> Assume two clients connect to a same session, once the user calls `session.fetch(tileable)`, is it possible to check if there is a corresponding graph which means the tileable is executed, so that we can fetch it.

Sometimes we might don't have any information about the tileable, except the `key`.

> Expose the key to public looks not so pleasant just in my opinion.

I would like to treat this "fetch by key" functionality as an internal (or debugging) API and the ordinary users shouldn't be aware that (when using numpy/pandas the id/address of ndarray/dataframe is never a concern).

> Is it possible to save the executed tileable objects on the cluster side

We already have such a mapping in `SessionActor`, the key is tileable key and the value is graph_ref.

hekaisheng:
Just a preliminary idea, for those tileables that used by different client, we might use a name to record them in server side, other clients can fetch them or construct a tileable by its name.

one client submit the job:
```
a = mt.random.rand(10, 10)
b = a + 1
session.run(b, name='b')
```
another client want to fetch it:
```
b = session.fetch(name='b')
```
or construct a fetch tileable:
```
b = mt.from_name('b')
# the logic like fetch `shape` is inside the method.
```

sighingnow:
I still think that adding `.name` is horrible... (for users). It will the mars users confused. We could just leave the pr here and mark it as "drafted" until we have a better idea to share session between multiple clients.

To be clear, this pr is a "workaround" for my case, where I need to create and seal a mutable tensor in one session and data is needed in another session. Maybe I will revise my side then we don't need this hack.

hekaisheng:
> I still think that adding `.name` is horrible... (for users). It will the mars users confused. We could just leave the pr here and mark it as "drafted" until we have a better idea to share session between multiple clients.
> 

As I mentioned above, `name` is not a tileable's attribute, it is used only when users want to share this tileable between multiple clients and only recorded the correspondence on the server side.

sighingnow:
> `name` is not a tileable's attribute

I see. The proposal looks pretty reasonable 👏 . Thanks for the clarification.

hekaisheng:
We support passing name when call `execute` and fetch data using the name in another sessions, you can check the code in #937, so close this PR.

PR #697, opened 2019-08-28
Support add dataframe with series

wdkwyf:
## What do these changes do?
Support dataframe + series feature. I add SeriesAlignMap and SeriesAlignReduce to achieve the similar function about dataframe + dataframe.

## Related issue number
#514


hekaisheng:
`SeriesIndexAlignMap` and `SeriesIndexAlignReduce` almost same as DataFrame's, should we use the same operands for both Series and DataFrame?

qinxuye:
Replaced by #737 , will be closed first.

PR #716, opened 2019-09-10
Add lock free option for workers

### LINKED ISSUE #715: [ENH] Lock on file IO can be freed when using network file systems
When using mounted network file systems, locks and concurrency controls can be turned off to improve throughput.

wjsi:
## What do these changes do?

Allow running workers without locks on file systems, useful when the path is a distributed FS.

## Related issue number

Fixes #715

### DIFF
--- a/mars/config.py
+++ b/mars/config.py
@@ -336,6 +336,7 @@ default_options.register_option('worker.transfer_block_size', 1 * 1024 ** 2, val
 default_options.register_option('worker.transfer_compression', 'lz4', validator=is_string, serialize=True)
 default_options.register_option('worker.prepare_data_timeout', 600, validator=is_integer)
 default_options.register_option('worker.peer_blacklist_time', 3600, validator=is_numeric, serialize=True)
+default_options.register_option('worker.lock_free_fileio', False, validator=is_bool, serialize=True)
 
 default_options.register_option('worker.plasma_socket', '/tmp/plasma', validator=is_string)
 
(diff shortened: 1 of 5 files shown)


mars/config.py
```diff
@@ -336,6 +336,7 @@ def validate(x):
 default_options.register_option('worker.transfer_compression', 'lz4', validator=is_string, serialize=True)
 default_options.register_option('worker.prepare_data_timeout', 600, validator=is_integer)
 default_options.register_option('worker.peer_blacklist_time', 3600, validator=is_numeric, serialize=True)
+default_options.register_option('worker.lock_free_io', False, validator=is_bool, serialize=True)
```
qinxuye: Only disk IO could be lock-free in my opinion. Could be a more specified name?

mars/worker/__main__.py
```diff
@@ -83,6 +85,7 @@ def create_pool(self, *args, **kwargs):
             n_cpu_process=self.args.cpu_procs,
             n_net_process=self.args.net_procs or self.args.io_procs,
             spill_dirs=self.args.spill_dir or os.environ.get('MARS_SPILL_DIRS'),
+            lock_free_io=self.args.lock_free_io or bool(int(os.environ.get('MARS_LOCK_FREE_IO', '0'))),
```
qinxuye: Shall we collect all environment variables together to make it more maintainable?

PR #717, opened 2019-09-10
[WIP]Fix mt.concatenate on sealed mutable tensor

### LINKED ISSUE #713: Mutable tensor doesn't work with mt.concatenate[BUG]
**Describe the bug**

Mutable tensor doesn't work with `mt.concatenate` (under local cluster deployment). If `a` is a mutable tensor, `mt.concatenate([a, a])` will get stuck when executing but `mt.concatenate([a.T, a.T])` works pretty well...

I'm wondering if there is any special code path for `TensorConcatenate`, but failed to find it.

**To Reproduce**
1. Your Python version: python 3.7
4. Full stack of the error.

**Expected behavior**

Sealed mutable tensor should could be used for `mt.concatentate` (directly).

sighingnow:
## What do these changes do?

When sealing mutable tensors `TensorFetch` chunks are generated. When execting `mt.concatenate` on sealed mutable tensor, `TensorFetch` chunks are direct dependencies of the target op, however `Fetch` chunks are not added to chunk graph and no operand actor is created for it. Then the `Fetch` chunks won't be added to finished chunk set for the target tileable. Thus the bug.

This pr adds `Fetch` chunks to finished chunk sets when preparing graph. This pr also includes some improvements (not related to the original bug).

## Related issue number

Fixes #713

### DIFF
--- a/mars/scheduler/graph.py
+++ b/mars/scheduler/graph.py
@@ -550,6 +550,9 @@ class GraphActor(SchedulerActor):
             for n in tileable_key_opid_to_tiled[(tk, topid)][-1].chunks:
                 self._terminal_chunk_op_tileable[n.op.key].add(tk)
                 self._target_tileable_chunk_ops[tk].add(n.op.key)
+                # treat fetch chunks (from sealed mutable tensor) as executed, see GH#713
+                if isinstance(n.op, Fetch):
+                    self._target_tileable_finished[tk].add(n.op.key)
 
         # sync chunk graph to kv store
         if self._kv_store_ref is not None:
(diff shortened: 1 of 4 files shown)


mars/scheduler/graph.py
```diff
@@ -550,6 +550,9 @@ def prepare_graph(self, compose=True):
             for n in tileable_key_opid_to_tiled[(tk, topid)][-1].chunks:
                 self._terminal_chunk_op_tileable[n.op.key].add(tk)
                 self._target_tileable_chunk_ops[tk].add(n.op.key)
+                # treat fetch chunks (from sealed mutable tensor) as executed, see GH#713
```
hekaisheng: It's better to not add fetch chunks to `self._target_tileable_chunk_ops` rather than add them into finished.

sighingnow:
Deprecated this pr, and will find a real fixes when we meet this problem again.

PR #720, opened 2019-09-11
[wip] Dataframe.__getitem__ by boolean series.

### LINKED ISSUE #721: [Feature request] Support more flexible DataFrame.__getitem__.

The argument of `DataFrame.[]` of pandas can be a boolean series. We should support this functionality in mars.

The challenge is when the index value of the mask series doesn't align with the index value of dataframe, we need perform a `shuffle` or `reindex` on the mask `series`. And when they have the same index value but are chunked in different `chunk_size`, a careful alignment will be needed.

We could start from the simplest case, that assuming they have the same index value, and are chunked in the same manner. This is also the most common case in daily use (since we normally `getitem` on a dataframe using a boolean column of the same dataframe, or a boolean series result of certain computation on some columns of the same dataframe).

sighingnow:
## What do these changes do?

Support `__getitem__` on dataframes using a series data (pandas series or mars series). Currently, only the mask series that has the same index value with the dataframe is supported.

## Related issue number

Fixes #721.

### DIFF
--- a/mars/dataframe/indexing/getitem.py
+++ b/mars/dataframe/indexing/getitem.py
@@ -18,8 +18,9 @@ import numpy as np
 from ... import opcodes as OperandDef
 from ...config import options
 from ...serialize import AnyField, Int32Field, BoolField
-from ..operands import DataFrameOperand, DataFrameOperandMixin, ObjectType
+from ..core import SERIES_TYPE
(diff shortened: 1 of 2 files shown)


sighingnow:
Close as in favor of #737.

PR #725, opened 2019-09-11
Spawn promise to utilize async network libs

### LINKED ISSUE #724: [ENH] Promise calls can be spawned to improve efficiency
When calling remote promises, the calls can be spawned with gevent to facilitate async network operations.

wjsi:
## What do these changes do?

Spawn remote promise calls with gevent to facilitate async operations.

## Related issue number

Fixes #724

### DIFF
--- a/mars/promise.py
+++ b/mars/promise.py
@@ -28,6 +28,7 @@ from .utils import wraps, build_exc_info
 
 logger = logging.getLogger(__name__)
 _promise_pool = dict()
+_pack_qword = struct.Struct('<Q').pack
 
 
 class Promise(object):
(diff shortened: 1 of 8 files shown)


mars/promise.py
```diff
@@ -28,6 +28,7 @@
 
 logger = logging.getLogger(__name__)
 _promise_pool = dict()
+_pack_qword = struct.Struct('<Q').pack
```
qinxuye: How about the name `_pack_longlong`? What does `qword` here mean?

wjsi: ``qword`` means quad bytes. This name consists with ``<Q``.

PR #731, opened 2019-09-12
Implements more flexible `.cix[]`.

sighingnow:
## What do these changes do?

An often-used access pattern for `.cix[]` may look like "access the first chunk of every rows" or "access the first chunks of every columns". In the previous solution we need to construct a series of indices like `[0, x]` or `[x, 0]` and `.cix[]` one by one, which is a bit boilerplate.

This pr ralex the constraint and enable the `.cix[]` for `slice()` indices, and return the chunks in a list. Note that we still require the indices has the same length with `chunk_shape` and don't do broadcasting. It is enough for nearly all use cases and can help avoid unintended bugs and errors.

No new test case is added since the correctness of `.cix[]` has been guaranteed by nearly all existing test cases (since nearly all ops use `.cix` in `tile()`). 

## Related issue number

N/A

### DIFF
--- a/mars/core.py
+++ b/mars/core.py
@@ -16,12 +16,12 @@
 
 import threading
 import itertools
-from operator import attrgetter, mul
+from operator import attrgetter
 from weakref import WeakKeyDictionary, WeakSet, ref
 
 import numpy as np
 
-from .compat import six, izip, builtins, reduce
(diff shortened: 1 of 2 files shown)


qinxuye:
Although the previous functionality of `cix` works fine, how about adding some tests about the new feature?

PR #762, opened 2019-09-22
Implements `numpy.random.shuffle` and `numpy.random.permutation` for tensor

### LINKED ISSUE #729: Implement `numpy.random.shuffle` and `numpy.random.permutation`

As `shuffle` and `permutation` are used to reshuffle data before training in deep learning or some stuff, we can introduce the API to Mars.

We can implement them with shuffle, in the shuffle map, we generate a random 1-d array in which each element indicates which downstream should sent to, and in the shuffle reduce, random shuffle the data.

qinxuye:
## What do these changes do?

This PR implements `mt.random.shuffle` and `mt.random.permutation` with shuffle.

## Related issue number

Resolves #729


mars/tensor/random/permutation.py
```diff
@@ -0,0 +1,231 @@
...
+        inputs = [ctx[(input_key, shuffle_key)] for input_key in input_keys]
+        inputs, device_id, xp = as_same_device(inputs, device=op.device, ret_extra=True)
+
+        with device(device_id):
+            rs = xp.random.RandomState(op.seed)
+            data = xp.concatenate(inputs)
+            ctx[op.outputs[0].key] = rs.permutation(data)
+
+
+def permutation(random_state, x):
```
hekaisheng: Should here pass `chunk_size`?

mars/tensor/random/shuffle.py
```diff
@@ -0,0 +1,55 @@
...
+           [3, 4, 5]])
+    """
+    from .permutation import permutation
+
+    if isinstance(x, (list, np.ndarray, TENSOR_TYPE)):
+        x = astensor(x)
+    else:
+        raise TypeError('x should be list, numpy ndarray or tensor')
+
+    ret = permutation(random_state, x)
```
hekaisheng: We should update key to differ from the tensor before shuffle.

mars/tensor/random/permutation.py
qinxuye: Done.

PR #767, opened 2019-09-24
Support DataFrame groupby.agg

hekaisheng:
Initial support for DataFrame groupby and aggregation. 

Now we can groupby and apply some aggregation functions (sum, prod, max, min) for each group.

``` Python
In [8]: df = pd.DataFrame({'c1': range(10), 
   ...:                    'c2': np.random.choice(['a', 'b', 'c'], (10,)), 
   ...:                    'c3': np.random.rand(10)})                                                                                             

In [9]: mdf = md.DataFrame(df, chunk_size=2)                                                                                                      

In [10]: mdf.groupby('c2').sum().execute()                                                                                                        
Out[10]: 
    c1        c3
c2              
a   20  3.112928
b    3  0.015811
c   22  1.555513
```


mars/dataframe/groupby/core.py
```diff
@@ -0,0 +1,154 @@
...
+            row_df = input_idx_to_df.get((row_idx, 0), None)
+            if row_df is not None:
+                res.append(row_df)
+        r = pd.concat(res, axis=0)
+        if chunk.index_value is not None:
+            r.index.name = chunk.index_value.value._name
+        ctx[chunk.key] = r
+
+
+class DataFrameGroupByOpernad(DataFrameOperand, DataFrameOperandMixin):
```
sighingnow: Typo: should be `DataFrameGroupByOperand`.

PR #805, opened 2019-11-05
Add ObjectData and ObjectChunk to represent data beyond ndarray, dataframe etc

qinxuye:
## What do these changes do?

This PR added ObjectData and ObjectChunk to represent data beyond ndarray, dataframe, series etc. Learn operand has been added as well, and xgboost operands have been refactored also.

## Related issue number

NA


mars/learn/contrib/xgboost/start_tracker.py
```diff
@@ -4,7 +4,7 @@
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
-#      http://www.apache.org/licenses/LICENSE-2.0
+#      http://www.apache.org/licenses/LICENSE-2.0git
```
hekaisheng: emm...remove it.

qinxuye:
> LGTM except one little typo and I think `output_type` is indeed a better name than `object_type` for many DataFrame operands like reductions as its output can be `DataFrame` or `Series`, I will rename it in next PR.

Agreed.

Issue #860, opened 2019-12-12
Optimize tileable graph using some planning rules


hekaisheng:
**Is your feature request related to a problem? Please describe.**
As described in #636, some optimization modules have been created now except 'tileable_graph', 
we need do some work on coarse-grained graph optimization. 

As a first step, we can apply some rules on the coarse-grained graph before tiles, predicate pushdown could be the first experimental rule.

qinxuye:
#531 is related to this issue.

PR #865, opened 2019-12-13
Add experimental column pruning rules for tileable graph optimization

hekaisheng:
## What do these changes do?

Adding an experimental implementation for tileable optimization rule, this rule works only when groupby aggregation operation follows the read CSV files, we can prune the columns that not used by the following operations when read the files.

## Related issue number

#860.


mars/optimizes/tileable_graph/tests/test_column_pruning.py
```diff
@@ -0,0 +1,72 @@
...
+import pandas as pd
+
+import mars.dataframe as md
+from mars.config import option_context
+from mars.tests.core import TestBase
+from mars.optimizes.tileable_graph.core import get_tileable_mapping
+
+
+class Test(TestBase):
+    def testGroupByPruneReadCSV(self):
```
qinxuye: We should also test 
1. two groupbyagg come after same `read_csv`.
2. `read_csv` which defines `usecols` to see if it works well.

mars/tiles.py
```diff
@@ -54,7 +54,7 @@ def _inplace_tile(self):
     def build_graph(self, graph=None, cls=DAG, tiled=False, compose=True,
                     **build_chunk_graph_kwargs):
         tileable_graph = graph if not tiled else None
-        tileable_graph_builder = TileableGraphBuilder(graph=tileable_graph, graph_cls=cls)
+        tileable_graph_builder = get_tileable_graph_builer()(graph=tileable_graph, graph_cls=cls)
```
qinxuye: Add an optimize option here and choose right graph builder according to the option.

mars/utils.py
```diff
@@ -702,3 +700,13 @@ def has_unknown_shape(tiled):
     if any(np.isnan(s) for s in itertools.chain(*tiled.nsplits)):
         return True
     return False
+
+
+def get_tileable_graph_builer():
```
qinxuye: This function is not need further.

mars/utils.py
```diff
@@ -428,8 +429,12 @@ def replace_with_fetch_or_copy(n):
                     copied.add(copy.data)
             return node_to_copy[n]
 
-        tileable_graph_builder = TileableGraphBuilder(
-            graph=graph, node_processor=replace_with_fetch_or_copy)
+        if options.optimize_tileable_graph:
```
qinxuye: `build_tileable_graph` is used on the client, shall we do the tileable graph optimization on the client? I don't think so.

mars/utils.py
```diff
@@ -729,3 +734,37 @@ def kill_process_tree(pid, include_parent=True):
             pass
     if plasma_sock_dir:
         shutil.rmtree(plasma_sock_dir, ignore_errors=True)
+
+
+def copy_tileables(*tileables, **kwargs):
+    inputs = kwargs.pop('inputs', None)
+    copy_key = kwargs.pop('copy_key', True)
+    copy_id = kwargs.pop('copy_id', True)
+    if len(tileables) == 1:
```
qinxuye: Unify logic for single and multiple outputs, no need to separate them.

mars/utils.py
```diff
@@ -740,31 +735,24 @@ def copy_tileables(*tileables, **kwargs):
...
+        params = t.params.copy()
+        if copy_key:
+            params['_key'] = t.key
+        if copy_id:
+            params['_id'] = t.id
+        params.update(t.extra_params)
+        kws.append(params)
+    inputs = inputs or op.inputs
+    copied = op.new_tileables(inputs, kws=kws, output_limit=len(kws))
+    if len(tileables) == 1:
```
qinxuye: How about len(tileables) > 1?

hekaisheng:
The implementation in this PR only works in local mode, distributed support will be a little different.

PR #879, opened 2019-12-25
Add public base class for entity data

wjsi:
## What do these changes do?

Add public base class, EntityData, for entity data

## Related issue number

NA


qinxuye:
Conflict needs to be resolved.

PR #880, opened 2019-12-25
Add `TestExecutor` that serde graph every time when executing to ensure all operands work well with serialize

### LINKED ISSUE #267: Check all operands about sparse, add unittests for both expressions and executions

qinxuye:
## What do these changes do?

This PR adds `TestExecutor` that serde graph every time when executing to ensure all operands work well with serialize.

`TestExecutor` only lives with unittest. The benefit of it is that we don't need to test each operand in a distributed environment.

## Related issue number

Fixes #267 .


mars/dataframe/core.py
```diff
@@ -322,13 +322,23 @@ class IndexChunk(Chunk):
     _allow_data_type_ = (IndexChunkData,)
 
 
+def _on_deserialize_index_value(index_value):
+    if index_value is None:
+        return
+    try:
+        getattr(index_value, 'value')
+        return index_value
+    except:
```
wjsi: Limit to AttributeError?

mars/serialize/jsonserializer.pyx
```diff
@@ -19,13 +19,11 @@ import pickle
 import weakref
 from collections import OrderedDict
 from io import BytesIO
+import sys
```
wjsi: Organize imports in alphabetical order and put imports before froms.

mars/serialize/pbserializer.pyx
```diff
@@ -18,6 +18,7 @@ import pickle
 import weakref
 from collections import OrderedDict
 from io import BytesIO
+import sys
```
wjsi: Ditto

PR #1002, opened 2020-02-17
Fix wrong result of count_nonzero

### LINKED ISSUE #999: [BUG] count_nonzero produces unexpected output for non-boolean tensor
**Describe the bug**
``count_nonzero`` only produces the expected output for boolean arrays and integer arrays of 1s and 0s.

**To Reproduce**
1. *Python:* 3.7
2. *The version of Mars you use:* github master
3. *Versions of crucial packages, such as numpy, scipy and protobuf:* 1.18.1, 1.4.1, 3.11.2

**Expected behavior**
`count_nonzero` should count the number of non-zero values of any type which are not, 0, 0.0 or False etc.

```python
import numpy as np
import mars.tensor as mt
```

The following two work as expected:
```python
values = [False, True, False, True]
print(np.count_nonzero(np.array(values)))
print(mt.count_nonzero(mt.tensor(values)).execute())
# 2
# 2

values = [0, 1, 0, 1]
print(np.count_nonzero(np.array(values)))
print(mt.count_nonzero(mt.tensor(values)).execute())
# 2
# 2
```
but then we run into problems:
```python
values = [0, 2, 0, 3]
print(np.count_nonzero(np.array(values)))
print(mt.count_nonzero(mt.tensor(values)).execute())
# 2
# 5

values = [0.0, 0.2, 0.0, 0.3]
print(np.count_nonzero(np.array(values)))
print(mt.count_nonzero(mt.tensor(values)).execute())
# 2
# 0.5
```
It seems like `mars` is just doing a sum.

hekaisheng:
## What do these changes do?

`count_nonzero` produces wrong result when tensor has only one chunk, this PR will fix this issue.

## Related issue number

Fixes #999

### DIFF
--- a/mars/tensor/reduction/tests/test_reduction_execute.py
+++ b/mars/tensor/reduction/tests/test_reduction_execute.py
@@ -173,6 +173,12 @@ class Test(unittest.TestCase):
         raw1 = np.random.random((20, 25))
         raw2 = np.random.randint(10, size=(20, 25))
 
+        arr0 = tensor(raw1)
+
+        res1 = self.executor.execute_tensor(arr0.var())
+        expected1 = raw1.var()
+        self.assertTrue(np.allclose(res1[0], expected1))
+
         arr1 = tensor(raw1, chunk_size=3)
 
         res1 = self.executor.execute_tensor(arr1.var())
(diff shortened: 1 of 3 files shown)


mars/tensor/reduction/tests/test_reduction_execute.py
```diff
@@ -173,6 +173,12 @@ def testVarExecution(self):
         raw1 = np.random.random((20, 25))
         raw2 = np.random.randint(10, size=(20, 25))
 
+        arr0 = tensor(raw1)
```
qinxuye: Better to specify `chunk_size` to a number, prevent from possible intelligent chunk size even though the tensor size is small.

mars/tensor/reduction/tests/test_reduction_execute.py
```diff
@@ -465,6 +471,13 @@ def testOutCumReductionExecution(self):
     def testCountNonzeroExecution(self):
         raw = [[0, 1, 7, 0, 0], [3, 0, 0, 2, 19]]
 
+        arr = tensor(raw)
```
qinxuye: Ditto.

Issue #1037, opened 2020-03-05
Implements md.{DataFrame,Series}.{sort_index,sort_values}


wjsi:
Sort functionality is available for tensors but not data frames, and sorting is more frequently used in data frames. Other functions also depend on sorting.

PR #1046, opened 2020-03-12
Add support for DataFrame.sort_values

hekaisheng:
## What do these changes do?

Support `DataFrame.sort_values` by a parallel sorting algorithm PSRS, the logic is similar to `mt.sort`.

## Related issue number
Solve part of #1037.

### DIFF
--- a/mars/dataframe/sort/sort_values.py
+++ b/mars/dataframe/sort/sort_values.py
@@ -1,0 +1,184 @@
+# Copyright 1999-2020 Alibaba Group Holding Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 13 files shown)


mars/dataframe/sort/sort_values.py
```diff
@@ -0,0 +1,254 @@
...
+                                     columns_value=out.columns_value, dtypes=out.dtypes)
+
+    @classmethod
+    def tile(cls, op):
+        df = op.inputs[0]
+
+        if op.axis == 0:
+            df = build_concatenated_rows_frame(df)
+        else:
+            df = df.rechunk({0: (df.shape[0],)})._inplace_tile()
```
qinxuye: I think for now we may drop support for axis == 1 because it generates unknown columns.

mars/tensor/base/partition.py
```diff
@@ -27,10 +27,10 @@
 from ..array_utils import as_same_device, device
 from ..datasource import tensor as astensor
 from ..utils import validate_axis, validate_order
-from .psrs import PSRSOperandMixin
+from .psrs import TensorPSRSOperandMixin
 
 
-class ParallelPartitionMixin(PSRSOperandMixin):
+class ParallelPartitionMixinTensor(TensorPSRSOperandMixin):
```
qinxuye: Rename it to Tensorxxxx

PR #1130, opened 2020-04-01
Add DataFrame docs

qinxuye:
## What do these changes do?

This PR added DataFrame docs, some docs are absent for now, we could add them later.

## Related issue number

NA

### DIFF
--- a/docs/doc-requirements.txt
+++ b/docs/doc-requirements.txt
@@ -4,3 +4,4 @@ pyarrow>=0.11.0
 sphinx>=1.7
 sphinx_rtd_theme>=0.2
 sphinx-intl>=0.9.9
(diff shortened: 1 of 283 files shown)


sighingnow:
Shall we consider to use [sphinx-autoapi](https://github.com/readthedocs/sphinx-autoapi) or something like to generate docs automatically for every modules ?

qinxuye:
> Shall we consider to use [sphinx-autoapi](https://github.com/readthedocs/sphinx-autoapi) or something like to generate docs automatically for every modules ?

Actually, the docs in dataframe are originated from pandas, I didn't spend much time on it, but we can pay attention to the tool later on.

Issue #1138, opened 2020-04-03
Implements aggregation with expanding and ewm


wjsi:
**Describe the solution you'd like**
Aggregation methods with ``md.expanding`` and ``md.ewm`` can be added. Computation methods are similar with cum operands.

PR #1160, opened 2020-04-16
Add support of `md.expanding`

wjsi:
## What do these changes do?

Add support of ``md.expanding``. The execution is done in two stages: the map stage and the combine stage. In the map stage, two chunks will be produced, i.e., the intermediate result chunk and the summary chunk. When producing the final result, all preceding summaries are collected together.

## Related issue number

Part of #1138 is resolved.

### DIFF
--- a/mars/dataframe/window/aggregation.py
+++ b/mars/dataframe/window/aggregation.py
@@ -1,0 +1,521 @@
+# Copyright 1999-2020 Alibaba Group Holding Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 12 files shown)


mars/dataframe/window/aggregation.py
```diff
@@ -0,0 +1,529 @@
...
+                chunks.append(chunk_op.new_chunk(
+                    [in_df.chunks[0]], dtypes=new_dtypes, index=(c.index[0], new_axis_idx),
+                    shape=(c.shape[0], len(new_dtypes)), index_value=c.index_value,
+                    columns_value=parse_index(new_dtypes.index, store_data=True)))
+            else:
+                params = c.params.copy()
+                params['dtype'] = out_df.dtype
+                chunks.append(chunk_op.new_chunk([in_df.chunks[0]], **params))
+
+        tensor_op = op.copy().reset_key()
```
qinxuye, line 302: Why called it `tensor_op`?

mars/dataframe/window/aggregation.py
```diff
@@ -0,0 +1,529 @@
...
+
+    @classmethod
+    def _remap_dtypes(cls, in_df, out_df):
+        if in_df.ndim == 1:
+            if out_df.ndim == 2:
+                return {0: (0, out_df.dtypes)}, (in_df.nsplits[0], (len(out_df.dtypes),))
+            return None, in_df.nsplits
+
+        axis = out_df.op.axis
+        out_dtypes_dict = dict()
```
qinxuye, line 256: Maybe we could call it `chunk_index_to_dtypes`.

Issue #1227, opened 2020-05-21
[PROPOSAL] Mars Remote API


qinxuye:
# Background

Currently, Mars has three distinguished APIs including

1. Mars tensor which mimics Numpy API for large tensor/ndarray operations.
2. Mars DataFrame which mimics pandas API for large DataFrame/Series/Index operations.
3. Mars learn which mimics scikit-learn for large-scale machine learning.

However, sometimes user may need to write some self-defined logic for those operations that cannot easily be interpreted by tensor, DataFrame or learn.

There are already some implementations in other systems, for instance, [Ray](http://github.com/ray-project/ray) which provides `ray.remote` decorator to send a function or Actor to remote. [Dask](http://github.com/dask/dask) which has a `delayed` API which provides the ability to composite a DAG with functions.

# API definition

## Create a remote function (resolved in #1238)

```python
import mars.remote as mr

result = mr.spawn(func, args=(), kwargs={})
```
At this time, it will not be executed unless `result.execute()` is triggered.

## Create a remote class

```python
import mars.remote as mr

class Counter:
    def __init__(self):
        self.val = 0

    def inc(self):
        self.val += 1

remote_cls = mr.spawn(Counter, args=())
result = remote_cls.inc()
```

Still, `result.fetch()` can only work when `result.execute()` is triggered first.

## Handle dependencies (resolved in #1238, #1296)

Remote functions should be able to depend on result of other remote ones. (#1238)

```python
import mars.remote as mr

def func1():
    return 1

def func2(x):
    return x + 1

result1 = mr.spawn(func1)
result2 = mr.spawn(func2, args=(result1,))
print(result2.execute())  # should print 2
```

What's more, remote function could reply on the result of DataFrame or tensor. (#1296)

```python
import mars.tensor as mt
import mars.remote as mr

a = mt.random.rand(10, 3)

def func(x, slc):
    arr = x[slc].fetch()
    return arr.sum()

result = mr.spawn(func, args=(a, slice(5)))
print(result.execute()) 
```

## Spawn a new function in a remote function (resolved in #1248)

We should provide the ability to spawn functions in a remote function. (#1248)

```python
import mars.remote as mr

def func1(x):
    return x + 1

def func2(xs):
    t = []
    for x in xs:
        res = mr.spawn(func1, args=(x,))
        t.append(res)
    return sum(mr.ExecutableTuple(t).execute().fetch())

print(mr.spawn(func2, args=([1, 2, 3],)).execute())  # print 9
```

[...]

PR #1238, opened 2020-05-22
Add preliminary remote function support

qinxuye:
## What do these changes do?

This PR added preliminary remote function support. Sample code:

```python
import mars.remote as mr

def f(x):
    return x + 1

def g(x, y):
    return x * y

a = mr.spawn(f, 3)
b = mr.spawn(f, 4)
c = mr.spawn(g, (a, b))

print(c.execute().fetch())  # print 20
```

## Related issue number

Finish part of #1227 .

### DIFF
--- a/mars/remote/core.py
+++ b/mars/remote/core.py
@@ -1,0 +1,145 @@
+# Copyright 1999-2020 Alibaba Group Holding Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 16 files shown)


mars/remote/core.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+class RemoteFunction(ObjectOperand, ObjectOperandMixin):
+    _op_type_ = opcodes.REMOTE_FUNCATION
+    _op_module_ = 'remote'
+
+    _function = FunctionField('function')
+    _function_args = ListField('function_args')
+    _function_kwargs = DictField('function_kwargs')
+
```
wjsi, line 33: LGTM

PR #1437, opened 2020-07-30
Fix pandas 1.1 compatibility

wjsi:
## What do these changes do?

Fix compatibility issues for pandas 1.1.


qinxuye:
About shift, already open an issue: https://github.com/pandas-dev/pandas/issues/35488

For now, we just manually call `_consolidate_inplace`.

qinxuye:
Compatibility issue about `DataFrame.agg` with multiple cumulative functions is opened in another issue: https://github.com/pandas-dev/pandas/issues/35490 

We need to keep `<1.1.0` until `1.1.1` released.

PR #1507, opened 2020-08-20
Update `mt.split` to support list and tuple

### LINKED ISSUE #1506: [BUG]some bugs about `mt.split`
## Describe the bug
I find `mt.split` does not work well when i try to use `mt.unique` and `mt.split` to group a sorted array.

I find two bugs:
- `mt.split` does not support tuple and array as the first argument. Could you consider supporting this for convenience of usage? See below:
```
>>> mt.split((1,1,1,2,2,3), (3,5))

~/anaconda3/lib/python3.7/site-packages/mars/tensor/base/split.py in split(ary, indices_or_sections, axis)
    199 
    200     """
--> 201     return _split(ary, indices_or_sections, axis=axis, is_split=True)

~/anaconda3/lib/python3.7/site-packages/mars/tensor/base/split.py in _split(a, indices_or_sections, axis, is_split)
    127 
    128 def _split(a, indices_or_sections, axis=0, is_split=False):
--> 129     op = TensorSplit(axis=axis, dtype=a.dtype)
    130     return op(a, indices_or_sections, is_split=is_split)
    131 

AttributeError: 'list' object has no attribute 'dtype'

>>> mt.split([1,1,1,2,2,3], (3,5))

~/anaconda3/lib/python3.7/site-packages/mars/tensor/base/split.py in split(ary, indices_or_sections, axis)
    199 
    200     """
--> 201     return _split(ary, indices_or_sections, axis=axis, is_split=True)

~/anaconda3/lib/python3.7/site-packages/mars/tensor/base/split.py in _split(a, indices_or_sections, axis, is_split)
    127 
    128 def _split(a, indices_or_sections, axis=0, is_split=False):
--> 129     op = TensorSplit(axis=axis, dtype=a.dtype)
    130     return op(a, indices_or_sections, is_split=is_split)
    131 

AttributeError: 'tuple' object has no attribute 'dtype'
```
- this one seems weird. I obtain the index from `mt.unique` with a sorted array and argument `return_index` True. When I passed the executed result of a slice of this index as the second arguments of `mt.split`, the returned sub-tensors all have nan shape and i cannot execute the sub-tensor to get the value. If i passed list, tuple or a `mt.array` with literal list, `mt.split` works well. See below for details:
```
In [67]: p = (1,1,1,2,2,3)                                                                                                                                                                           

In [68]: unique, index = mt.unique(p, return_index=True)                                                                                                                                             

In [69]: index.execute()                                                                                                                                                                             
Out[69]: array([0, 3, 5])

In [70]: index1 = index[1:].execute()                                                                                                                                                                

In [71]: index1                                                                                                                                                                                      
Out[71]: array([3, 5])

In [72]: mt.split(mt.array(p), index1)                                                                                                                                                               
Out[72]: 
(Tensor <op=TensorSplit, shape=(nan,), key=61fbfb5330abb901e5b56269d6fbad5f,
 Tensor <op=TensorSplit, shape=(nan,), key=f3a4dc180f9350bb7ac5ffcb18abd297,
 Tensor <op=TensorSplit, shape=(nan,), key=65fadafe692317ee6ba59630753cafb5)

In [73]: mt.split(mt.array(p), mt.array((3,5)))                                                                                                                                                      
Out[73]: 
(Tensor <op=TensorSplit, shape=(3,), key=6f04c920a336bb8ae733b50c91da57ba,
 Tensor <op=TensorSplit, shape=(2,), key=11ef35d05602119da39c0924a6787245,
 Tensor <op=TensorSplit, shape=(1,), key=dd1c7e9460185aef0991c5cfa8e77823)
```

YoshieraHuang:
## What do these changes do?

update `mt.split` to support list and tuple as first argument

## Related issue number

Fixes #1506

### DIFF
--- a/mars/tensor/base/split.py
+++ b/mars/tensor/base/split.py
@@ -20,8 +20,8 @@ from ... import opcodes as OperandDef
 from ...serialize import KeyField, AnyField, Int32Field
 from ...lib.sparse.core import get_array_module
 from ...core import ExecutableTuple
-from ..datasource import tensor as astensor
 from ..core import Tensor
+from ..datasource import tensor as astensor
 from ..utils import calc_sliced_size
 from ..operands import TensorHasInput, TensorOperandMixin
 
--- a/mars/tensor/base/split.py
+++ b/mars/tensor/base/split.py
@@ -20,6 +20,7 @@ from ... import opcodes as OperandDef
 from ...serialize import KeyField, AnyField, Int32Field
 from ...lib.sparse.core import get_array_module
 from ...core import ExecutableTuple
+from ..datasource import tensor as astensor
 from ..core import Tensor
 from ..utils import calc_sliced_size
 from ..operands import TensorHasInput, TensorOperandMixin
@@ -198,4 +199,4 @@ def split(ary, indices_or_sections, axis=0):
      array([], dtype=float64)]
 
     """
-    return _split(ary, indices_or_sections, axis=axis, is_split=True)
+    return _split(astensor(ary), indices_or_sections, axis=axis, is_split=True)
--- a/mars/tensor/base/tests/test_base.py
+++ b/mars/tensor/base/tests/test_base.py
@@ -373,12 +373,6 @@ class Test(unittest.TestCase):
         for a in ((1,1,1,2,2,3), [1,1,1,2,2,3]):
             splits = split(a, (3,5))
             self.assertEqual(len(splits), 3)
-            # np.testing.assert_array_equal(splits[0].execute(), (1,1,1))
-            # np.testing.assert_array_equal(splits[1].execute(), (2,2))
-            # np.testing.assert_array_equal(splits[2].execute(), (3,))
-            self.assertEqual(splits[0].shape, (3,))
-            self.assertEqual(splits[1].shape, (2,))
-            self.assertEqual(splits[2].shape, (1,))
 
     def testSqueeze(self):
         data = np.array([[[0], [1], [2]]])
--- a/mars/tensor/base/tests/test_base.py
+++ b/mars/tensor/base/tests/test_base.py
@@ -370,6 +370,13 @@ class Test(unittest.TestCase):
         self.assertTrue(splits[2].flags['F_CONTIGUOUS'])
         self.assertFalse(splits[0].flags['C_CONTIGUOUS'])
 
+        for a in ((1,1,1,2,2,3), [1,1,1,2,2,3]):
+            splits = split(a, (3,5))
+            self.assertEqual(len(splits), 3)
+            np.testing.assert_array_equal(splits[0].execute(), (1,1,1))
+            np.testing.assert_array_equal(splits[1].execute(), (2,2))
+            np.testing.assert_array_equal(splits[2].execute(), (3,))
+
     def testSqueeze(self):
         data = np.array([[[0], [1], [2]]])
         x = tensor(data)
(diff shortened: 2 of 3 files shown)


mars/tensor/base/tests/test_base.py
```diff
@@ -370,6 +370,13 @@ def testSplit(self):
         self.assertTrue(splits[2].flags['F_CONTIGUOUS'])
         self.assertFalse(splits[0].flags['C_CONTIGUOUS'])
 
+        for a in ((1,1,1,2,2,3), [1,1,1,2,2,3]):
+            splits = split(a, (3,5))
+            self.assertEqual(len(splits), 3)
+            np.testing.assert_array_equal(splits[0].execute(), (1,1,1))
```
qinxuye, line 376: In our unittests, like `test_base.py` only test the expression part, which means, `.execute()` will not show up in these tests. 

The unittests for execution can be added to `test_base_execute.py` to verify the result.

mars/tensor/base/split.py
```diff
@@ -20,6 +20,7 @@
 from ...serialize import KeyField, AnyField, Int32Field
 from ...lib.sparse.core import get_array_module
 from ...core import ExecutableTuple
+from ..datasource import tensor as astensor
```
qinxuye, line 23: Move this line under line 24, we prefer to keep the order as alphabet order.

qinxuye (commented):
Really nice job, I left some comments about code style as well as unittests. Could you please send another commit to resolve them, thank you in advance.

qinxuye:
IMO, This PR covers the first part of #1506 , right?

mars/tensor/base/tests/test_base.py
qinxuye: Here I think test the length is enough, no need to verify the result any further.

mars/tensor/base/tests/test_base_execute.py
```diff
@@ -444,6 +444,13 @@ def testArraySplitExecution(self):
         [np.testing.assert_equal(r, e) for r, e in zip(res, expected)]
 
     def testSplitExecution(self):
+        for a in ((1,1,1,2,2,3), [1,1,1,2,2,3]):
+            splits = split(a, (3,5))
+            self.assertEqual(len(splits), 3)
+            np.testing.assert_array_equal(splits[0].execute(), (1,1,1))
```
qinxuye, line 450: `.execute()` is sort of a high-level API, in our tests, we use the code like

```python
self.executor.execute_tensor(splits[0], concat=True)[0]
```

Could you please modify the code again for line 450-452. Thank you so so much.

YoshieraHuang:
> IMO, This PR covers the first part of #1506 , right?

yes, only first part of #1506

qinxuye (commented):
Further tiny comments are left, thank you again.

qinxuye:
Flake8 test failed, please remove the line

```python
from operator import concat
```

which is not used.

YoshieraHuang:
The autoimport extension in vscode adds the unused imports

qinxuye:
> The autoimport extension in vscode adds the unused imports

Sigh... We mainly use pycharm for development.

YoshieraHuang:
> > The autoimport extension in vscode adds the unused imports
> 
> Sigh... We mainly use pycharm for development.

vscode is a good editor for me because i have to handle many other languages

qinxuye:
Since the second part of #1506 is not actually a bug, but instead by design, we can close the issue after this PR, what do you think @YoshieraHuang ?

YoshieraHuang:
> Since the second part of #1506 is not actually a bug, but instead by design, we can close the issue after this PR, what do you think @YoshieraHuang ?

OK, we can close the issue after this PR. Thank you for your help.

qinxuye:
> > Since the second part of #1506 is not actually a bug, but instead by design, we can close the issue after this PR, what do you think @YoshieraHuang ?
> 
> OK, we can close the issue after this PR. Thank you for your help.

OK, I will edit the information so that it will close related issue after merged.

PR #1511, opened 2020-08-21
Optimize `mars.learn.cluster.KMeans` when `n_clusters` is relatively large

### LINKED ISSUE #1510: `mars.learn.cluster.KMeans` needs to be optimized when `n_clusters` is relatively large
When `n_clusters` is large, there may be some critical issues for init in `mars.learn.cluster.KMeans`.

1. For `k-means++`, if `n_clusters` is large, the iteration in it could be very slow, and the tiled graph could be very large(may be about 200k+ nodes). We may need to cut iterations into pieces to prevent from large chunk graph.
2. for `k-means||`, a.k.a scalable k-means++, the intermediate result could be large when calculating euclidean distances, assume `n_clusters=2000`, every time 4000 centers will be sampled, and assume the chunk shape of X is (1m, 64), then the chunk may occupy up to ~2TB which is quite insane. 

Thus, two things need to be achieved.

1. Do rechunk in `euclidean_distances` to prevent from potential large chunks.
2. Cut `k-means++` iterations into pieces and trigger execution to reduce size of chunk graph.

qinxuye:
## What do these changes do?

This PR did two things to optimize `mars.learn.cluster.KMeans` when `n_clusters` is relatively large.

1. Do rechunk in `euclidean_distances` to prevent from potential large chunks.
2. Cut k-means++ iterations into pieces and trigger execution to reduce size of chunk graph.

## Related issue number

Resolves #1510 .

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,7 +78,7 @@ jobs:
             if [[ $UNAME == "linux" ]] && [[ ! "$PYTHON" =~ "3.8" ]]; then
               pip install tensorflow\<2.3.0
               pip install torch torchvision
-              pip install tsfresh
+              pip install tsfresh statsmodels==0.11.1
             fi
             virtualenv testenv && source testenv/bin/activate
             pip install -r requirements.txt && pip install pytest pytest-timeout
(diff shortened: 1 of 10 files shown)


wjsi (approved):
LGTM

PR #1540, opened 2020-09-01
Use relative paths to avoid web rendering issues under backward proxies

### LINKED ISSUE #1539: [BUG] Cannot load Mars Web correctly when under backward proxy
When Mars Web is deployed under a backward proxy, for instance, under nginx configuration

```
location /redir/root/ {
  proxy_pass http://localhost:13452/;
  proxy_http_version 1.1;
  proxy_set_header Upgrade $http_upgrade;
  proxy_set_header Connection "Upgrade";
  proxy_set_header Host $host;
}
```

all styles are lost:

image

wjsi:
## What do these changes do?

Use relative paths in HTML resources to avoid rendering issues under backward proxies.

NOTE that as bokeh does not provide capability for applications under relative paths, we hack its autoload scripts. This shall be changed when the issue in bokeh is resolved.

## Related issue number

Fixes #1539


qinxuye (approved):
LGTM

PR #1572, opened 2020-09-14
Implements `df.rebalance()`

### LINKED ISSUE #1557: Add `df.rebalance()` to make data more balanced across cluster

For now, if a DataFrame, for instance, have 3 chunks which allocated on 3 workers, all the upcoming computations which rely on the data could not beyond 3 workers, `df.rebalance()` is used to rebalance the data to more workers.

For example, if there's 100 cores in the cluster, we can call `df.rebalance(factor=1.2)` which will rebalance the data to 120 chunks, and the rebalanced data will occupy all the workers, and be more efficient if the upcoming computation wants to leverage all the cores in the cluster.

Inside `rebalance`, we first calculate the chunk count, that is `ncores * factor`, assume it's 120, then we calculate the new chunk_size via `shape // 120`, trigger iterative tiling, finally, we just call `df.rechunk(chunk_size)`.

qinxuye:
## What do these changes do?

This PR added support for `df.rebalance()`.

## Related issue number

Resolves #1557 .

### DIFF
--- a/docs/source/reference/dataframe/api/mars.dataframe.DataFrame.map_chunk.rst
+++ b/docs/source/reference/dataframe/api/mars.dataframe.DataFrame.map_chunk.rst
@@ -1,0 +1,6 @@
+mars.dataframe.DataFrame.map\_chunk
+===================================
+
+.. currentmodule:: mars.dataframe
+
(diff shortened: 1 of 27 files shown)


wjsi (approved):
LGTM

PR #1574, opened 2020-09-15
Implements `fetch_log` API

### LINKED ISSUE #1564: Propose `fetch_log` API
For a distributed setting, if some functions from `mr.spawn` or `df.apply` etc are submitted, the users cannot see the log unless they login to the worker machine to see the outputs, it's quite inconvenient, especially when the workers number is huge. Thus, I suggest LOG API here to enhance the user experience.

# Code example

For a Mars tileable including Object(always created by remote API), DataFrame, add a `fetch_log` API. For instance.

```python
import mars.remote as mr

def f(i):
    print(f'Finish running part {i}')

result = [mr.spawn(f, args=i) for i in range(10)]
mr.ExecutableTuple(result).execute()
print(result[0].fetch_log())
```

Output will be

```
Finish running part 0
```

# API definition

`fetch_log` could be defined as:

```
def fetch_log(obj, offsets=None, sizes=None, indexes=None)
```

When a tileable has multiple chunks, by default, `fetch_log` will collect all logs, store them in a dict that stores chunk_index -> log, chunk_indexes can be specified in order to narrow down those chunks they focus.

Each time calling `fetch_log`, the current offsets for each chunk will be fetched, and the next time calling `fetch_log` will pass the last offsets so that new logs can be fetched.

`sizes` is used to limit the log size.

# Implementation

We need a couple of components.

1. LogMetaActor, which created for scheduler, owns a nested dict stores (session_id, tileable_op_key) -> {chunk_op_key -> worker_address+log_path}
2. A decorator can be added to redirect the stdout to a file when executing the custom functions.
3. LogFetchActor which created for workers, maybe holds by each process, that supports fetching a log which output by a function.
4. Add `fetch_log` method to context, api, web and session.

qinxuye:
## What do these changes do?

This PR introduced `fetch_log` API which described in #1564 .

Tasks

- [x] Support `fetch_log` for remote API.
- [x] Async execute.
- [x] Support `fetch_log` for DataFrame custom functions.
- [x] Test for web session.
- [x] Add doc.

## Related issue number

Resolves #1564 .


wjsi (approved):
LGTM

PR #1584, opened 2020-09-22
Added `DataFrame.isin` support

### LINKED ISSUE #756: Implement DataFrame.isin and Series.isin
Implements `DataFrame.isin(self, values)`:

> _**values**_ : iterable, Series, DataFrame or dict
> The result will only be true at a location if all the labels match. If values is a Series, that’s the index. If values is a dict, the keys must be the column names, which must match. If values is a DataFrame, then both the index and column labels must match.

qinxuye:
## What do these changes do?

This PR added `DataFrame.isin` support.

## Related issue number

Resolves #756 .

### DIFF
--- a/docs/source/getting_started/execution.rst
+++ b/docs/source/getting_started/execution.rst
@@ -46,6 +46,8 @@ intermediately.
 Asynchronous Execution
 ======================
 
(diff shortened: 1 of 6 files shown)


wjsi (approved):
LGTM

PR #1612, opened 2020-09-28
Add an extra sort key in PSRS to make distinct pivot

### LINKED ISSUE #1611: [BUG] sort_values with same value causes ExecutionFailed
When calling sort_values with same value on a huge DataFrame, ExecutionFailed is raised. ValueError(Cannot allocate size larger than the total capacity) is received on a worker.

wjsi:
## What do these changes do?

Add (chunk_key, row_id) as an extra sort key to resolve value bias in PSRS sort algorithm.

## Related issue number

Fixes #1611


hekaisheng (approved):
LGTM

PR #1640, opened 2020-10-19
Optimize memory usage for brute-force algorithm in `NearestNeighbors`

### LINKED ISSUE #1621: Brute-force KNN costs too much memory
For now, brute-force algorithm in KNN costs too much memory, the reason is that for now, the pairwise distances will be calculated first, then topk, however, when number of vectors is huge on both sides, the distances matrix may be insanely huge. Chunked way can be considered instead.

qinxuye:
## What do these changes do?

This PR introduced an internal function `pairwise_distances_topk` to reduce the memory cost for `NearestNeighbors`.

## Related issue number

Resolves #1621 .


wjsi (approved):
LGTM

PR #1673, opened 2020-11-02
Add a threshold for `DataFrame.head` optimization

### LINKED ISSUE #1658: [BUG] Run DataFrame getitem leads to memory insufficiency
**Describe the bug**
When execute df.iloc[:1000000], the job hangs and warning found in server side:
```
2020-10-28 14:08:40,463 mars.worker.quota 200 WARNING  w:0:MemQuotaActor met hard memory limitation: request 16613268648, available 9270380134, hard limit 61847529062
```
this may be caused by the optimization introduced in #1328, if `n` is too large, it will lead to a huge chunk when do reductions, we should skip the optimization if `n` greater than a threshold.

hekaisheng:
## What do these changes do?

Expressions like `df.head(n)` will be optimized to reduce the memory usage, however, the strategy will even requires more memory when `n` is large, in this PR, we set a threshold 1000, if `n` greater than it, the optimization will be disabled.

## Related issue number

Fixes #1658.

### DIFF
--- a/mars/config.py
+++ b/mars/config.py
@@ -377,6 +377,7 @@ default_options.register_option('worker.plasma_socket', '/tmp/plasma', validator
 default_options.register_option('optimize.min_stats_count', 10, validator=is_integer)
 default_options.register_option('optimize.stats_sufficient_ratio', 0.9, validator=is_float, serialize=True)
 default_options.register_option('optimize.default_disk_io_speed', 10 * 1024 ** 2, validator=is_integer)
+default_options.register_option('optimize.head_optimize_threshold', 1000, validator=is_integer)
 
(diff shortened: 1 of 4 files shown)


mars/config.py
```diff
@@ -379,6 +379,7 @@ def validate(x):
 default_options.register_option('optimize.default_disk_io_speed', 10 * 1024 ** 2, validator=is_integer)
 
 default_options.register_option('optimize_tileable_graph', True, validator=is_bool)
+default_options.register_option('optimize.head_limit', 1000, validator=is_integer)
```
wjsi, line 382: Rename to ``head_optimize_limit`` and move to the line below ``'optimize.default_disk_io_speed'``.

wjsi (approved):
LGTM

PR #1677, opened 2020-11-04
Mount /dev/shm on host to pods when starting Mars workers in Kubernetes

### LINKED ISSUE #1676: [BUG] Sometimes Mars workers launched in kubernetes terminate with SIGBUS
Mars workers launched in kubernetes sometimes terminate with SIGBUS when load is heavy.

wjsi:
## What do these changes do?

Mount `/dev/shm` on host to pods when starting Mars workers in kubernetes. This will effectively fix SIGBUS issue. Add lower bound for shared memory also.

## Related issue number

Fixes #1676


qinxuye (approved):
LGTM

PR #1700, opened 2020-11-10
Remount the shm directory in entrypoint.sh

hekaisheng:
## What do these changes do?

Remount the shm directory in entrypoint.sh.


wjsi (approved):
LGTM

PR #1702, opened 2020-11-10
Add shared limit option for Mars worker

wjsi:
## What do these changes do?

Add limit option on shared memory path (i.e., /dev/shm) for Mars worker.


hekaisheng (approved):
LGTM

Issue #1476, opened 2020-08-11
Support pandas.DataFrame.kurt and pandas.DataFrame.skew in mars DataFrame


brightcoder01:
**Is your feature request related to a problem? Please describe.**
[pandas.DataFrame.kurt](https://pandas.pydata.org/pandas-docs/version/1.0.0/reference/api/pandas.DataFrame.kurt.html) and [pandas.DataFrame.skew](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.skew.html) are very important for data analysis. 
It would be useful that mars DataFrame can support these two functions for a large scale of data.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

qinxuye:
Duplicate with #1342 , close first, just track there.

wjsi:
Reopen as split from #1342: it is better to implement kurt and skew independently.

Issue #1311, opened 2020-06-18
Implement customized aggregation


wjsi:
Allow specifying customized map-combine-reduce functions to aggregate.

Issue #1141, opened 2020-04-08
Support aggregation with args


wjsi:
**Is your feature request related to a problem? Please describe.**
Mars.agg() currently can only support aggregating without args. Args can be supported via

```python
df.agg(lambda x: x.var(ddof=0))
```

and when tiling and we detect that the return value is an aggregation object, the expression can still be executed in map-combine-agg manner.

PR #1705, opened 2020-11-11
Use ReductionCompiler to support function aggregation in `mars.dataframe.reduction`

wjsi:
## What do these changes do?

This PR adds a `ReductionCompiler` to `mars.dataframe.reduction` to handle function aggregations like

```python
df.agg(lambda x: (x ** 2).sum() + 1)
```

A graph is built for `(x ** 2).sum() + 1` with Mars tileable graph and separated with reduction operands into pre-process graphs and post-process graphs. Pre-handling functions and post-handling functions are then generated with a minimal codegen.

By doing this we successfully migrated implementations for `var` and `std` into function aggregations and supports for `skew` and `kurt` are also added.

To unify reduction functions in `mars.dataframe.reduction` such as `unique` and `nunique`, support for `CustonReduction` is also added.

Note that support for `mars.dataframe.groupby` is not added yet. This may rely on a refactor for `GroupbyWrapper` to fix potential bugs.

## Tests

Under Intel 10700K CPU and Python 3.8,

With the patch:
```python
In[1]: import mars.dataframe as md
In[2]: import numpy as np
In[3]: %%timeit
  ...: df = md.DataFrame(np.random.rand(100, 10), chunk_size=10)
  ...: df.sum().execute()
  ...: 
30.1 ms ± 218 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
In[4]: %%timeit
  ...: df = md.DataFrame(np.random.rand(100, 10), chunk_size=10)
  ...: df.sum().execute()
  ...: 
46 ms ± 272 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Without the patch:
```python
In[1]: import mars.dataframe as md
In[2]: import numpy as np
In[3]: %%timeit
  ...: df = md.DataFrame(np.random.rand(100, 10), chunk_size=10)
  ...: df.sum().execute()
  ...: 
36.1 ms ± 831 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
In[4]: %%timeit
  ...: df = md.DataFrame(np.random.rand(100, 10), chunk_size=10)
  ...: df.sum().execute()
  ...: 
54 ms ± 367 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

GPU cases are also tested with cudf. Now `var` and `agg` can be supported on Nvidia GPUs.

## Related issue number

(Note that none of these PRs fully fixed till `groupby` issue is resolved.)

#1476, #1311, #1141

### DIFF
--- a/docs/source/reference/dataframe/api/mars.dataframe.CustomReduction.rst
+++ b/docs/source/reference/dataframe/api/mars.dataframe.CustomReduction.rst
@@ -1,0 +1,25 @@
+﻿mars.dataframe.CustomReduction
+==============================
+
+.. currentmodule:: mars.dataframe
+
+.. autoclass:: CustomReduction
+
+   
+   .. automethod:: __init__
+
+   
+   .. rubric:: Methods
+
+   .. autosummary::
+   
+      ~CustomReduction.__init__
+      ~CustomReduction.agg
+      ~CustomReduction.post
+      ~CustomReduction.pre
+   
+   
+
+   
+   
+   
--- a/docs/source/reference/dataframe/api/mars.dataframe.DataFrame.kurt.rst
+++ b/docs/source/reference/dataframe/api/mars.dataframe.DataFrame.kurt.rst
@@ -1,0 +1,6 @@
+﻿mars.dataframe.DataFrame.kurt
+=============================
+
+.. currentmodule:: mars.dataframe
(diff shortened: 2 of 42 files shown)


qinxuye (approved):
LGTM

PR #1708, opened 2020-11-16
Write into and read from merged files when data sizes are small

wjsi:
## What do these changes do?

Write into  and read from merged files when data sizes are small. This reduces the number of inodes when using external storages.


qinxuye (approved):
LGTM

PR #1709, opened 2020-11-16
Remove `prepare_inputs` property on operands

wjsi:
## What do these changes do?

As all usages of `prepare_inputs` can be replaced with `pure_depends` without pains, the former property can be replaced.


hekaisheng (approved):
LGTM

PR #1769, opened 2020-12-09
Spawn serialization of executable graphs

wjsi:
## What do these changes do?

Spawn serialization of executable graphs so that web pages does hot get stuck.


qinxuye (approved):
LGTM

PR #1776, opened 2020-12-11
Integrate `str.cat` into reduction and groupby-aggregation

### LINKED ISSUE #1775: Allow performing `str.agg` after groupby
In Mars, `str.agg` is not only available for series. In pandas, ``df.groupby('col').agg(lambda x: x.str.cat(sep='-')`` can be used. Now with ReductionCompiler we can support this kind of aggregation with ease.

wjsi:
## What do these changes do?

Add ``str.cat`` as base aggregation method. Now users can call ``df.groupby('col').agg(lambda x: x.str.cat(sep='+')``.

## Related issue number

Fixes #1775

### DIFF
--- a/mars/dataframe/base/string_.py
+++ b/mars/dataframe/base/string_.py
@@ -16,14 +16,12 @@ import numpy as np
 import pandas as pd
 
 from ... import opcodes as OperandDef
-from ...config import options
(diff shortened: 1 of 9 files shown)


qinxuye (approved):
LGTM

Issue #1036, opened 2020-03-05
Support full featured groupby function


wjsi:
Currently df.groupby() in Mars only implements ``as_index``. ``sort`` is added but not implemented yet, while ``level`` is not implemented which is also useful. What's more, when ``GroupBy`` object is generated, keys and grouped data frames are serialized instead of ``GroupBy`` object itself, which may not be a universal solution for different scenarios of groupby.

PR #1779, opened 2020-12-12
Implements reduction with `level` argument

wjsi:
## What do these changes do?

Implements reductions like `df.sum(level=1)` using groupby and reduction compiler.

## Related issue number

Deal with full groupby issue #1036.


qinxuye (approved):
LGTM

PR #1810, opened 2020-12-22
Support using compound agg function in lambda

wjsi:
## What do these changes do?

Allow calling compound aggregation functions, such as ``df.agg(lambda x: x.mean() + x.skew())`` in df.agg(). Also simplify definitions of builtin functions.


qinxuye (approved):
LGTM

PR #1827, opened 2020-12-25
Refine kubedl cluster-api.

SimonCqk:
## What do these changes do?

1. remove `tolerations` with single `Exists` operator so as to avoid being scheduled to a unexpected node.
2. provide a ratio to set replicas resources limits, scheduler/webservice `limit_resources_ratio=1` defaulted, worker `limit_resources_ratio=2` defaulted.


wjsi (approved):
LGTM

PR #1823, opened 2020-12-25
Add `incremental_index` argument to `reset_index` which by default is False

qinxuye:
## What do these changes do?

This PR added `incremental_index` argument which by default is False to `reset_index`.

**Remember that this is a break change**.

## Related issue number

Resolves #1822 .

PR #1842, opened 2020-12-25
[BACKPORT] Add `incremental_index` argument to `reset_index` which by default is False (#1823)

wjsi:
Backport of #1823 and #1827 .


qinxuye (approved):
LGTM

PR #1856, opened 2020-12-30
Support specifying memory limit scale in kubernetes

### LINKED ISSUE #1855: Allow specifying memory limit scale in kubernetes
Kubernetes supports ``resource_request`` and ``resource_limit``, and the latter can be specified to allow Mars cluster to use more memory if current node has memory to spare. We can make use of this design to make Mars execution more robust.

wjsi:
## What do these changes do?

Allow specifying memory limit scale. This will reduce Mars workers being killed by cluster management.

## Related issue number

Fixes #1855


hekaisheng (approved):
LGTM

PR #1857, opened 2020-12-31
Set marsjob worker cache by memoryTuningPolicy.

SimonCqk:
## What do these changes do?

1. set worker cache by memoryTuningPolicy instead of directly environments.


hekaisheng (approved):
LGTM

PR #1866, opened 2021-01-05
Use mmap files to reduce memory usage in proxima builder

hekaisheng:
## What do these changes do?

Use numpy.memmap to create a mmap file to reduce memory usage when training data is too large.


qinxuye (approved):
LGTM

Issue #1840, opened 2020-12-25
Implements iteration methods of DataFrame objects


wjsi:
Implements iteration methods of DataFrame objects, including

* [x] `items`
* [x] `iteritems`
* [ ] `keys`
* [ ] `values`

BobinMathew:
hi @wjsi, where do yo want to add this checklist?

qinxuye:
> hi @wjsi, where do yo want to add this checklist?

I think `items` as well `iteritems` may be more friendly.

You can refer to https://github.com/mars-project/mars/blob/1b3801ac669d810e5513351f8b66f436f082fb05/mars/dataframe/core.py#L1131-L1136

PR #1867, opened 2021-01-05
Implements `{DataFrame,Series}.rename_axis`

### LINKED ISSUE #1665: `dataframe.index` needs to add  attribute  like `name` of the index

I find it hard to change the name of  index of dataframes when I use mars. Compared to mars, you can change the name by the following ways:

```python
df.index.name = "_id" #this way means inplace=True; single-level index
df.index.names = ["_id"] # multi-level index
```
or
```python
df.index.set_names(['foo'],inplace=True)
```
With it, I can change the name of single-level index and multi-level index more convenient.

wjsi:
## What do these changes do?

Implements `{DataFrame,Series}.rename_axis`. Also allows setting names of columns or indexes.

## Related issue number

Fixes #1665 
Part of #1840 is done (iteritems() and items() on series)


hekaisheng (approved):
LGTM

qinxuye (approved):
LGTM

PR #1869, opened 2021-01-06
Support unknown shape for mt.reshape, mt.histogram and md.DataFrame

qinxuye:
## What do these changes do?

This PR added support for mt.reshape, mt.histogram and md.DataFrame when input tileable has unknown shape.

## Related issue number

Fixes #1868 
Fixes #1843

PR #1874, opened 2021-01-07
Fix reshape when input tensor has unknown shape and 1 chunk

### LINKED ISSUE #1873: [BUG] histogram fails on DataFrames with only one chunk without deterministic shape
**Describe the bug**

Histogram fails on DataFrames with only one chunk without deterministic shape.

Code reproducing this bug:

```python
import mars.dataframe as md, mars.tensor as mt
import numpy as np
rnd = np.random.rand(100, 10)
mdf = md.DataFrame(rnd)
mdf1 = mdf.reset_index().rename(columns={0: 'a'})
ser = mdf1.a.fillna(0)
tp = mt.histogram(ser[np.isfinite(ser)])
tp.execute()
```

qinxuye:
## What do these changes do?

This PR fixes reshape when input tensor has unknown shape and 1 chunk.

This is a followed fix of #1869 .

## Related issue number

Fixes #1873 .

### DIFF
--- a/mars/dataframe/core.py
+++ b/mars/dataframe/core.py
@@ -894,6 +894,10 @@ class SeriesData(_BatchedFetcher, BaseSeriesData):
 
     items = iteritems
 
+    def to_dict(self, into=dict, batch_size=10000, session=None):
+        fetch_kwargs = dict(batch_size=batch_size)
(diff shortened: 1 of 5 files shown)


wjsi (approved):
LGTM

PR #1880, opened 2021-01-11
Allow internal serialization to use JSON

wjsi:
## What do these changes do?

As it is quite slow to construct protobuf objects in Python due to type checks, we change serialization method into JSON and make serialization methods configurable.


qinxuye (approved):
LGTM

PR #1884, opened 2021-01-12
Optimize performance of `df.sort_values().head()`

qinxuye:
## What do these changes do?

This PR optimized performance of `df.sort_values().head()`

## Related issue number

NA

PR #1886, opened 2021-01-12
Support column pruning for groupby().agg() on data sources

qinxuye:
## What do these changes do?

This pR supported column pruning for groupby().agg() on data sources.

#1884 should be merge first.

## Related issue number

NA


hekaisheng (approved):
LGTM

PR #1892, opened 2021-01-14
Add `stop_cluster` for kubedl

hekaisheng:
## What do these changes do?

Add `stop_cluster` for kubedl and print url when creating kubedl.


hekaisheng:
Will submit another PR to solve this problem, close now.

PR #1899, opened 2021-01-19
To vineyard: avoid copy when chunks are already in vineyard (vineyard is the backend).

sighingnow:
## What do these changes do?

+ Avoid copy chunks again when they are already in vineyard.
+ Use latest released vineyard in CI, rather than a hacked one.

## Related issue number

N/A

### DIFF
--- a/.github/workflows/platform-ci.yml
+++ b/.github/workflows/platform-ci.yml
@@ -91,13 +91,8 @@ jobs:
         run: |
           source ./.github/workflows/reload-env.sh
 
-          # pull vineyard image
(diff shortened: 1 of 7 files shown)


sighingnow:
The coverage missing for diff looks OK, since we doesn't test such case for non-distributed/clustered testing. This PR added two extra cases that has already covered some uncovered branches.

qinxuye (approved):
LGTM

wjsi (approved):
LGTM

sighingnow:
Thank for merging!

PR #1911, opened 2021-01-21
Allow set verify_ssl to False for kubernetes configuration

hekaisheng:
## What do these changes do?

Add `KUBE_VERIFY_SSL` environment in kubernetes to ignore the certificate when connect to api server.

## Related issue number

NA


wjsi (approved):
LGTM

Issue #1903, opened 2021-01-20
[oscar][ray backend] Implements actor creation


fyrestone:
Implement `create_actor`, `destroy_actor` and `has_actor` in ray backend.

PR #1916, opened 2021-01-22
[oscar][ray backend] Actor creation

fyrestone:
## What do these changes do?
Actor creation on ray backend.

## Related issue number

https://github.com/mars-project/mars/issues/1903

### DIFF
--- a/mars/oscar/backends/ray/context.pyx
+++ b/mars/oscar/backends/ray/context.pyx
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import asyncio
 from ...core cimport _Actor, ActorRef
 from ...context cimport BaseActorContext
 from ...errors import ActorNotExist, ActorAlreadyExist
@@ -21,60 +22,46 @@ from ....utils import lazy_import
 ray = lazy_import("ray")
 
 
-cdef class _RayActorHandle:
-    cdef object actor_handle
-    cdef object uid
-    cdef object address
-
-
 cdef class RayActorContext(BaseActorContext):
-    cdef public dict actors
-
     def __init__(self, str address=None, index=0):
-        self.actors = dict()
+        pass
 
     async def create_actor(self, object actor_cls, *args, object uid=None, object address=None, **kwargs):
-        cdef _RayActorHandle actor = _RayActorHandle()
-
-        if uid in self.actors:
+        uid = str(uid)
+        try:
+            # TODO(fyrestone): We should make the actor dead when current job is dropped.
+            actor_handle = ray.remote(actor_cls).options(
+                name=uid, lifetime="detached").remote(*args, **kwargs)
+        except ValueError:
             raise ActorAlreadyExist(f'Actor {uid} already exist, cannot create')
-
-        actor_handle = ray.remote(actor_cls).options(name=str(uid)).remote(*args, **kwargs)
-        actor.actor_handle = actor_handle
-        actor.uid = uid
-        actor.address = address
-        self.actors[uid] = actor
-        actor_handle._set_uid.remote(uid)
-        actor_handle._set_address.remote(address)
-        actor_handle.__post_create__.remote()
+        await asyncio.gather(
+            actor_handle._set_uid.remote(uid),
+            actor_handle._set_address.remote(address))
+        await actor_handle.__post_create__.remote()
         return ActorRef(address, uid)
 
     async def has_actor(self, ActorRef actor_ref):
-        return actor_ref.uid in self.actors
+        try:
+            # Maybe poor performance.
+            ray.get_actor(actor_ref.uid)
+            return True
+        except Exception:
+            return False
 
     async def destroy_actor(self, ActorRef actor_ref):
-        cdef _RayActorHandle actor
-
         try:
-            actor = self.actors[actor_ref.uid]
+            actor_handle = ray.get_actor(actor_ref.uid)
         except KeyError:
             raise ActorNotExist(f'Actor {actor_ref.uid} does not exist')
 
-        actor.actor_handle.__pre_destroy__.remote()
-        del self.actors[actor_ref.uid]
+        await actor_handle.__pre_destroy__.remote()
+        ray.kill(actor_handle)
         return actor_ref.uid
 
     async def send(self, ActorRef actor_ref, object message, bint wait_response=True):
-        cdef _RayActorHandle actor
-
         try:
-            actor = self.actors[actor_ref.uid]
-        except KeyError:
+            actor_handle = ray.get_actor(actor_ref.uid)
+        except ValueError:
             raise ActorNotExist(f'Actor {actor_ref.uid} does not exist')
-        ret = await actor.actor_handle.__on_receive__.remote(message)
+        ret = await actor_handle.__on_receive__.remote(message)
         return ret if wait_response else None
-
-
-def register_backend():
-    from ...context import register_backend_context
-    register_backend_context("ray", RayActorContext)
--- a/mars/oscar/backends/ray/context.pyx
+++ b/mars/oscar/backends/ray/context.pyx
@@ -1,0 +1,80 @@
+# Copyright 1999-2020 Alibaba Group Holding Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


mars/oscar/__init__.py
```diff
@@ -12,11 +12,66 @@
...
+
+
+def unregister_actor_implementation(actor_cls):
+    try:
+        del _actor_implementation[id(actor_cls)]
+    except KeyError:
+        pass
+
+
+# Cython has a bug about inspecting async function: https://github.com/cython/cython/issues/2273
```
wjsi, line 40: We do not need to use ``asyncio.iscoroutinefunction``. Checking return value is enough in most scenarios. Thus please reconsider moving this code into ``__init__``.

mars/oscar/backends/ray/context.pyx
```diff
@@ -0,0 +1,80 @@
...
+            raise ActorAlreadyExist(f'Actor {uid} already exist, cannot create')
+
+        actor_handle = ray.remote(actor_cls).options(name=str(uid)).remote(*args, **kwargs)
+        actor.actor_handle = actor_handle
+        actor.uid = uid
+        actor.address = address
+        self.actors[uid] = actor
+        actor_handle._set_uid.remote(uid)
+        actor_handle._set_address.remote(address)
+        actor_handle.__post_create__.remote()
```
wjsi, line 49: Is this method synchronous? Can this call utilize asyncio?

mars/oscar/backends/ray/context.pyx
```diff
@@ -0,0 +1,80 @@
...
+
+
+cdef class _RayActorHandle:
+    cdef object actor_handle
+    cdef object uid
+    cdef object address
+
+
+cdef class RayActorContext(BaseActorContext):
+    cdef public dict actors
```
wjsi, line 31: I wonder how can this actor registry be visible to all actor processes? If not, ``has_actor`` may not work.

mars/oscar/backends/ray/context.pyx
```diff
@@ -0,0 +1,80 @@
...
+        if uid in self.actors:
+            raise ActorAlreadyExist(f'Actor {uid} already exist, cannot create')
+
+        actor_handle = ray.remote(actor_cls).options(name=str(uid)).remote(*args, **kwargs)
+        actor.actor_handle = actor_handle
+        actor.uid = uid
+        actor.address = address
+        self.actors[uid] = actor
+        actor_handle._set_uid.remote(uid)
+        actor_handle._set_address.remote(address)
```
wjsi, line 48: Can address here determine the placement group of the actor?

mars/oscar/backends/ray/tests/test_ray_actor_context.py
```diff
@@ -0,0 +1,219 @@
...
+        ref1 = await mo.create_actor(DummyActor, 1)
+        with self.assertRaises(TypeError):
+            await ref1.add(1.0)
+        ref2 = await mo.create_actor(DummyActor, 2)
+        with self.assertRaises(TypeError):
+            await ref1.send(ref2, 'add', 1.0)
+        with self.assertRaises(mo.ActorNotExist):
+            await mo.actor_ref('fake_uid').add(1)
+
+    # Fail on ray backend, the DummyActor.create will create an actor in actor process.
```
wjsi, line 169: Configuring RayActorContext as the global context when starting an actor in Ray will enable creating actor in actors.

mars/oscar/core.pyx
```diff
@@ -13,8 +13,10 @@
 # limitations under the License.
 
 import asyncio
+import cython
```
wjsi, line 16: ``cimport cython`` ?

mars/oscar/__init__.py
fyrestone: But ray uses `asyncio.iscoroutinefunction` to determine whether to run async or not.

mars/oscar/backends/ray/context.pyx
fyrestone: The `remote()` call just send the method name and args to remote. This method does not block waiting for the result.

mars/oscar/backends/ray/context.pyx
fyrestone: From the discussion, mars holds a context in each node, the `actors` is not shared among them. Am I right?

mars/oscar/__init__.py
wjsi: I'd rather add a util to enable ``asyncio.iscoroutinefunction`` by adding a util function like:

```python
# add in a pure .py file
def aio_coro_function(func):
    async def dummy():
        pass
    func._is_coroutine = getattr(dummy, '_is_coroutine')
```

mars/oscar/backends/ray/context.pyx
fyrestone: Yes, we can use placement group here. But I need to figure out the strategy that how to place the group by address.

mars/oscar/core.pyx
fyrestone: I can use `cimport` instead of `import`.

Thanks. I will fix it.

mars/oscar/__init__.py
fyrestone: Then we need to use this `iscoroutinefunction` in ray or monkey patch ray.

mars/oscar/backends/ray/context.pyx
wjsi: ``create_actor`` need to return when ``__post_create__`` is finished. Thus we need to wait asynchronously.

mars/oscar/backends/ray/tests/test_ray_actor_context.py
fyrestone: The problem is that `DummyActor, 1` and `DummyActor, 2` are in different processes.
- The RayActorContext in current process has actor `DummyActor, 1`
- The RayActorContext in `DummyActor, 1` has actor `DummyActor, 2`.

Then we get `DummyActor, 2` in current process as `ref2`, any call on `ref2` will be failed. Maybe we can use ray named actor to fix this issue.

mars/oscar/__init__.py
wjsi: A tested diff can be https://github.com/mars-project/mars/compare/refactor/oscar...wjsi:refactor/oscar

mars/oscar/backends/ray/context.pyx
wjsi: In my understanding, has_actor is needed to work in the scope of a node with processes hosting actors. I cannot see how Actor A created in Actor B can be seen in Actor C when A, B and C are all inside one node.

mars/oscar/__init__.py
wjsi: > Then we need to use this `iscoroutinefunction` in ray or monkey patch ray.

Sorry deleted a response by mistake. In my implementation above, no changes needed for ``iscoroutinefunction``.

fyrestone: A bit hacky. I will try it, thanks.

mars/oscar/backends/ray/context.pyx
fyrestone: Got it. I will await the `__post_create__`.

mars/oscar/backends/ray/context.pyx
fyrestone: It sounds like the named actor is the solution.

qinxuye:
Hi, @fyrestone , we added an actor driver class which has a method called `setup_cluster`, it can be used to setup a cluster for Ray. Besides, new backend now just needs to inherit from `BaseActorBackend`.

mars/oscar/__init__.py
fyrestone: Ray is using the `inspect. iscoroutinefunction` not the `asyncio. iscoroutinefunction`: https://github.com/ray-project/ray/blob/master/python/ray/_raylet.pyx#L421

qinxuye: > Ray is using the `inspect.iscoroutinefunction` not the `asyncio.iscoroutinefunction`: https://github.com/ray-project/ray/blob/master/python/ray/_raylet.pyx#L421
> 
> I can create a PR for ray to use `asyncio.iscoroutinefunction` instead.

Now in #1925 , the Actor class has been moved to `api.py` which is not a cython file any more.

fyrestone: Thanks.

wjsi:
LGTM in general. Please rename tests in ray backend from "\*Local*" into "\*Ray*". What's more, copyright info for `mars/oscar/backends/ray/tests/__init__.py` is missing.

fyrestone:
> backend

Fixed, Thanks.

qinxuye (approved):
LGTM, thank you for your contribution!

PR #1934, opened 2021-01-26
Fix memory estimation of StartTracker for XGBoost

qinxuye:
## What do these changes do?

This PR fixes memory estimation of StartTracker for XGBoost

## Related issue number

Fixes #1932

PR #1931, opened 2021-01-26
Fix shuffling data in mars.learn module

wjsi:
## What do these changes do?

``mars.learn.shuffle`` fails to set output type for ``LearnShuffleProxy``. When executed distributedly, Mars cannot generate proper FetchShuffle object, and execution fails.

## Related issue number

Fixes #1930

PR #1936, opened 2021-01-26
Fix error when shuffling data in `mars.learn` module (#1931) and estimation of StartTracker for XGBoost (#1934)

wjsi:
Backport of #1931 and #1934


qinxuye (approved):
LGTM

PR #1937, opened 2021-01-27
Add new serializer implementation

wjsi:
## What do these changes do?

Add new implementation of serialization. Serializer now provides header and buffers to facilitate serialization in detail.


qinxuye (approved):
LGTM

PR #1956, opened 2021-02-02
Allow wrapping existing models with Mars class constructors

wjsi:
## What do these changes do?

Allow wrapping existing models with Mars class constructors.


hekaisheng (approved):
LGTM

PR #1958, opened 2021-02-02
[oscar] Fix actor promise & add tests

wjsi:
## What do these changes do?

Fix actor promise calls. Coverage is also refined.


.github/workflows/platform-ci.yml
```diff
@@ -133,7 +133,7 @@ jobs:
             coverage report
           fi
           if [ -n "$WITH_RAY" ]; then
-            pytest $PYTEST_CONFIG --cov-config .coveragerc-threaded mars/ray
+            pytest $PYTEST_CONFIG --cov-config .coveragerc-threaded mars/ray mars/oscar/backends/ray
```
qinxuye, line 136: Shall us use -k option? Refer to https://docs.pytest.org/en/stable/example/markers.html#using-k-expr-to-select-tests-based-on-their-name

qinxuye (dismissed):
LGTM, -k option can be done later.

qinxuye (approved):
LGTM

PR #1961, opened 2021-02-03
Optimize performance of `DataFrame.describe()`

### LINKED ISSUE #1942: `DataFrame.describe()` is too slow when df has many columns
`DataFrame.describe()` is too slow when df has many columns.

qinxuye:
## What do these changes do?

For now, users cannot skip calculating percentiles, because it's very time-consuming.  Now we can specify `percentiles=False` which is illegal for pandas, but used to skip percentiles computation for Mars.

Besides, this PR used `DataFrame.agg` to perform aggregation computation instead of doing reduction for each column.

## Related issue number

Fixes #1942 .

### DIFF
--- a/mars/dataframe/base/describe.py
+++ b/mars/dataframe/base/describe.py
@@ -17,12 +17,16 @@ import pandas as pd
 
 from ... import opcodes as OperandDef
 from ... import tensor as mt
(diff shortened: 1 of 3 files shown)


wjsi (approved):
LGTM

PR #1994, opened 2021-02-20
Implements `DataFrameGroupBy.sample`

### LINKED ISSUE #1984: Implements DataFrameGroupBy.sample
Implements GroupBy.sample. Optimization before shuffle groupby need to be considered.

wjsi:
## What do these changes do?

Implements `DataFrameGroupBy.sample`. Note that two issues of pandas may affect this PR.
1. pandas-dev/pandas#39927, making it impossible to sample distributedly with weights, and
2. pandas-dev/pandas#39928, making it inconsistent between Mars and pandas when selection of multiple columns is specified.

What's more, SeriesGroupBy does not have sample method, thus we don't test it.

## Related issue number

Fixes #1984


qinxuye (approved):
LGTM

PR #1996, opened 2021-02-23
Implements Mars backend for oscar

qinxuye:
## What do these changes do?

This PR implements Mars backend for oscar.

TODO:

- [x] Rewrite `test_local_context.py` based on new implemnetation.
- [x] Ray test needs to be refined with pytest and new fashion.

## Related issue number

NA


wjsi (approved):
LGTM

PR #2002, opened 2021-02-24
Add `errors` argument for groupby.sample to ignore errors when group size less than `n`

### LINKED ISSUE #2000: Add options to ignore errors when sample size is greater than group size
When executing `groupby.sample`, it is possible to have size of groups larger than size to sample, making it hard to get samples from huge DataFrames esp. when number of groups is large. We may need an option to ignore errors when encountering small groups.

wjsi:
## What do these changes do?

Add `errors` argument for groupby.sample to ignore errors when group size less than `n`. This also adapts logic from pandas groupby.sample and then bugs of weighted sampling in pandas will not affect Mars.

## Related issue number

Fixes #2000


qinxuye (approved):
LGTM

PR #2010, opened 2021-02-25
Add initial service implementations

wjsi:
## What do these changes do?

First blood for structured Mars service implementations.


qinxuye (approved):
LGTM

PR #2013, opened 2021-03-01
Add support for batch interfaces for actors

wjsi:
## What do these changes do?

Add support for batch interfaces for actors. Actor writers can define actors as

```python
class TestActor(mo.Actor):
    @extensible
    def method(self, arg1, arg2):
        # implementation for single data

    @method.batch
    def method(self, args_list, kwargs_list):
        # implementation for batch data
```

and invoke as

```python
ref = await mp.actor_ref('TestActor', address='address')
await ref.batch(
    ref.method(val1_1, val2_1),
    ref.method(val1_2, val2_2),
)
```


qinxuye (approved):
LGTM

PR #2020, opened 2021-03-05
Support setting column with different index for DataFrame

### LINKED ISSUE #2018: [BUG] DataFrame set columns should support Series with different index
**Describe the bug**

Now, when calling `df['field'] = series` when series has a different index, NotImplementedError will be raised, since we have `reindex` implementation now, it can be supported.

**To Reproduce**
1. Your Python version
2. The version of Mars you use
4. Full stack of the error.

```
In [52]: df = md.DataFrame(index=[1, 2, 3])                                     

In [53]: df['a'] = md.Series(['a', 'b', 'c'], index=[2, 3, 4])                  
NotImplementedError                       Traceback (most recent call last)
<ipython-input-53-f550a59ef82c> in <module>
----> 1 df['a'] = md.Series(['a', 'b', 'c'], index=[2, 3, 4])

~/Workspace/mars/mars/dataframe/indexing/setitem.py in dataframe_setitem(df, col, value)
    182 def dataframe_setitem(df, col, value):
    183     op = DataFrameSetitem(target=df, indexes=col, value=value)
--> 184     return op(df, value)

~/Workspace/mars/mars/utils.py in _inner(*args, **kwargs)
    457         @functools.wraps(func)
    458         def _inner(*args, **kwargs):
--> 459             with self:
    460                 return func(*args, **kwargs)
    461 

~/Workspace/mars/mars/dataframe/indexing/setitem.py in __call__(self, target, value)
     83 
     84             if value.index_value.key != target.index_value.key:  # pragma: no cover
---> 85                 raise NotImplementedError('Does not support setting value '
     86                                           'with different index for now')
     87 

NotImplementedError: Does not support setting value with different index for now
```

qinxuye:
## What do these changes do?

This PR addresses a few issues:

1. Support setting column with different index for DataFrame
2. `md.DataFrame` can create with empty data.
3. `md.DataFrame` now supports creating from  a list of tileables.

## Related issue number

Fixes #2018 .

### DIFF
--- a/mars/dataframe/datasource/from_tensor.py
+++ b/mars/dataframe/datasource/from_tensor.py
@@ -120,7 +120,10 @@ class DataFrameFromTensor(DataFrameOperand, DataFrameOperandMixin):
 
         if index is not None:
             tileable_size = tileables[0].shape[0]
-            index_size = index.shape[0]
+            if hasattr(index, 'shape'):
+                index_size = index.shape[0]
+            else:
+                index_size = len(index)
             if not pd.isna(tileable_size) and not pd.isna(index_size) and \
                     tileable_size != index_size:
                 raise ValueError(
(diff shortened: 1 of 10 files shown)


wjsi (approved):
LGTM

Issue #2023, opened 2021-03-05
Optimize continuous binary operands using tree reduction


wjsi:
Continuous arithmetic calculations like `a + b + c + d` can be optimized with tree reductions (i.e., TreeReduction(a, b, c, d) which makes operand tree more shallow and can be parallelized easier.

This may be implemented with

* [ ] operands for arithmetic tree reduction, and
* [ ] optimization strategy for continuous operands

PR #2024, opened 2021-03-05
Implements tree arithmetic for tensor add and multiplication

wjsi:
## What do these changes do?

Implements tree arithmetic for tensor add and multiplication to make operand tree of consecutive arithmetic operations shallower. Most of the codes moved from `mars.tensor.special.core`.

## Related issue number

Deal with #2023


qinxuye (approved):
LGTM

PR #2027, opened 2021-03-06
[oscar] Add cancel support, optimize error handling, add `kill_actor` API

qinxuye:
## What do these changes do?

This PR did a few things:

1. Add cancel support, now an actor operation can be cancelled. 
2. Error handling is optimized.
3. Actor pool now can auto recovering when process is done, if `auto_recover` is specified as True or 'actor', the actors created on the process will be created automatically.
4. `kill_actor` API is added to make sure an Actor can be destroyed even if it's blocked, but remember this is a dangerous operation that it actually did it by killing a process, make sure the actors that created in the same process are stateless.
5. Now more than 1 pools can be created.

## Related issue number

NA

### DIFF
--- a/mars/lib/gipc.pyx
+++ b/mars/lib/gipc.pyx
@@ -442,6 +442,30 @@ except ImportError:
     ForkProcess = multiprocessing.Process
 
 
+if not WINDOWS:
+    try:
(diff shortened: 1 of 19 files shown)


wjsi (approved):
LGTM

PR #2031, opened 2021-03-08
Support cuda buffer serializations in actor communication

hekaisheng:
## What do these changes do?

Add serialize and deserialize utils for cuda buffers as they need to be copied to host in actor calls.


wjsi (approved):
LGTM

PR #2045, opened 2021-03-17
Fix __pre_destroy__ not called when actors are in main pool

### LINKED ISSUE #2044: [BUG] __pre_destroy__ not called when running oscar actors
`__pre_destroy__` is not invoked when actors are destroyed.

```python
import mars.oscar as mo
        
class TestActor(mo.Actor):
    async def __pre_destroy__(self):
        print('__pre_destroy__')
        
pool = await mo.create_actor_pool('127.0.0.1', n_process=0)
await pool.start()
ref = await mo.create_actor(TestActor, address=pool.external_address)
await ref.destroy()
# nothing printed
```

wjsi:
## What do these changes do?

Fix call `__pre_destroy__` when actors are in main pool.

## Related issue number

Fixes #2044

### DIFF
--- a/mars/oscar/api.py
+++ b/mars/oscar/api.py
@@ -77,6 +77,10 @@ def setup_cluster(address_to_resources: Dict[str, Dict[str, Number]]):
 
 
 class Actor(_Actor):
+    @classmethod
+    def default_uid(cls):
+        return cls.__name__
+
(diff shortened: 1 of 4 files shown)


qinxuye (approved):
LGTM

PR #2050, opened 2021-03-18
Allow sync actor methods to be extensible

wjsi:
## What do these changes do?

Allow sync actor methods to be extensible. Scheduling APIs also added.


qinxuye (approved):
LGTM

PR #2052, opened 2021-03-19
Fix for `mt.insert` when insert values is a mars tensor

hekaisheng:
## What do these changes do?

Fix for `mt.insert` and small changes in deploy for kubedl.


wjsi (approved):
LGTM

PR #2054, opened 2021-03-21
[Ray] Fix ray cluster_utils import

chaokunyang:
## What do these changes do?
Fix ray cluster_utils import on ray 2.0.


qinxuye (approved):
LGTM

PR #2058, opened 2021-03-22
Allow serializer to serialize recursive objects

wjsi:
## What do these changes do?

Allow serializer ro serialize recursive objects. Duplicated objects will be serialized only once.


qinxuye (approved):
LGTM

PR #2060, opened 2021-03-23
Implements initial version of lifecycle service

wjsi:
## What do these changes do?

Implements initial version of cluster service


qinxuye:
Replaced by #2117 .

Issue #2059, opened 2021-03-23
`mt.unique()` fails when passing an empty list as input


shizhiqiang007:
import mars.tensor as mt
mt.unique([]).execute(）

qinxuye:
This error could only emerge for numpy 1.18.x, not for 1.19.x.

```
In [1]: import mars.tensor as mt

In [2]: mt.unique([]).execute()
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-04232e429360> in <module>
----> 1 mt.unique([]).execute()

~/Workspace/mars/mars/core.py in execute(self, session, **kw)
    656 
    657         if wait:
--> 658             return run()
    659         else:
    660             thread_executor = ThreadPoolExecutor(1)

~/Workspace/mars/mars/core.py in run()
    652 
    653         def run():
--> 654             self.data.execute(session, **kw)
    655             return self
    656 

~/Workspace/mars/mars/core.py in execute(self, session, **kw)
    389 
    390         if wait:
--> 391             return run()
    392         else:
    393             # leverage ThreadPoolExecutor to submit task,

~/Workspace/mars/mars/core.py in run()
    384         def run():
    385             # no more fetch, thus just fire run
--> 386             session.run(self, **kw)
    387             # return Tileable or ExecutableTuple itself
    388             return self

~/Workspace/mars/mars/session.py in run(self, *tileables, **kw)
    504         tileables = tuple(mt.tensor(t) if not isinstance(t, (Entity, Base)) else t
    505                           for t in tileables)
--> 506         result = self._sess.run(*tileables, **kw)
    507 
    508         for t in tileables:

~/Workspace/mars/mars/session.py in run(self, *tileables, **kw)
    109             # set number of running cores
    110             self.context.set_ncores(kw['n_parallel'])
--> 111             res = self._executor.execute_tileables(tileables, **kw)
    112             return res
    113 

~/Workspace/mars/mars/utils.py in _inner(*args, **kwargs)
    458         def _inner(*args, **kwargs):
    459             with self:
--> 460                 return func(*args, **kwargs)
    461 
    462         return _inner

~/Workspace/mars/mars/executor.py in execute_tileables(self, tileables, fetch, n_parallel, n_thread, print_progress, mock, compose, name)
    881                     chunk_graph.compose(list(temp_result_keys))
    882                 # execute chunk graph
--> 883                 self.execute_graph(chunk_graph, list(temp_result_keys),
[...]

PR #2061, opened 2021-03-23
Fix `mt.unique` on empty arrays

wjsi:
## What do these changes do?

Fix `mt.unique` on empty arrays for numpy 1.18.x.

## Related issue number

Deal with #2059 on v0.6.


qinxuye (approved):
LGTM

PR #2063, opened 2021-03-25
Oscar actor pool on ray

### LINKED ISSUE #2066: oscar actor pool on ray
Currently `mars actor on ray` implementation just map mars actor to ray actor directly. This model has a huge gap with mars actor architecture. Mars supports multiple actors running on same process and supports actor allocate strategy to create actor on the specified process. When multiple actors are on the same process, there won't be data serialization which is very important for controller actors. However, when running on ray, every ray actor is a separate process, so data serialization is inevitable and may be inconsistent, we can't support allocate strategy, thus we can't make multiple controller actors running on the same process too.

The implementation is totally different with mars oscar actor implementation which also has extra maintainance cost. It would be great if we can support running mars actor pool on ray directly. This way we can share all mars oscar actor architecture benefits and envolve with  mars oscar actor too.

chaokunyang:
## What do these changes do?
This PR make mars oscar actor pool running on ray. The main changes include three parts:
* Ray client/server channel based on ray call
* Ray Server/Client
* Mars actor pool code restructure 
* Mars actor pool management abstraction
* Ray actor pool implementation
* Ray cluster setup/stop

In this implementation, ray actors/placementgroup doesn't have to be detached. When job is finished, all resoruce used by the job will be recycled.
## Related issue number
closes #2066


mars/oscar/backends/mars/ray.py
```diff
@@ -0,0 +1,275 @@
...
+
+class NodeResourceSpec:
+    __slots__ = 'n_process', 'num_cpus', 'memory', 'num_gpus'
+
+    def __init__(self, n_process, num_cpus=None, num_gpus=None, memory=None):
+        self.n_process = n_process
+        self.num_cpus = num_cpus or n_process
+        assert n_process <= self.num_cpus
+        self.num_gpus = num_gpus
+        self.memory = memory
```
fyrestone, line 217: This `NodeResourceSpec` type does not support custom resource well, and it's not consistent with ray's placement group API https://docs.ray.io/en/master/placement-group.html#starting-a-placement-group

mars/oscar/backends/mars/ray.py
```diff
@@ -0,0 +1,275 @@
...
+
+    def addresses(self):
+        return [process for node in self.get_all_nodes_info() for process in node]
+
+    def address_to_resources(self) -> Dict[str, Dict[str, Number]]:
+        # TODO(chaokunyang) support address resources mapping
+        return {process: {"CPU": 1} for node in self.get_all_nodes_info() for process in node}
+
+
+def create_cluster(cluster_name, resource_specs: List[NodeResourceSpec]):
```
fyrestone, line 259: Why defining a `create_cluster` here? There is a `setup_cluster` in `backends/ray/driver.py`.

mars/oscar/backends/mars/ray.py
```diff
@@ -0,0 +1,275 @@
...
+        pool = await RaySubActorPool.create({
+            'actor_pool_config': actor_config,
+            'process_index': process_index
+        })
+        await pool.start()
+        self.actor_pool = pool
+        asyncio.create_task(pool.join())
+
+
+def address_to_placement_info(address):
```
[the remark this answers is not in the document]
chaokunyang: If we use mars actor pool on ray, the `mars actor on ray` implementation under `backends/ray` seems not necessary. We can remove it later or merge related code into `mars oscar actor pool`

mars/oscar/backends/mars/ray.py
chaokunyang: The API will be refined when `deploy on ray` is finished.

mars/oscar/backends/mars/ray.py
chaokunyang: It's an convenience API for `deploy on ray`. `Mars actor pool on ray` use `MarsActorDriver. The whole implementation is based on `mars actor pool`. It's not very realted to implementation under `backends/ray`

chaokunyang:
If this PR is accepted, I'll submitted another PR to remove `mars actor to ray actor` implementation unser `mars/oscar/backends/ray`. The addressing of actors for both implementation is conflicting

mars/oscar/backends/mars/ray.py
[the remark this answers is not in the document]
fyrestone: If this ray actor pool feature is ready, we can use it instead of `mars actor on ray actor`.

mars/oscar/backends/mars/ray.py
fyrestone: I think it's OK to use a plain dict to represent a bundle resource. Then we just put the dict to ray's placement group API.

mars/oscar/backends/mars/ray.py
chaokunyang: removed

mars/oscar/backends/mars/pool.py
```diff
@@ -878,147 +841,267 @@ async def create(cls, config: Dict) -> "MainActorPool":
...
         await super().stop()
         if self._monitor_task and not self._monitor_task.done():
             await self._monitor_task
             self._monitor_task = None
 
+    @classmethod
+    def get_sub_pool_manager_cls(cls):
+        return MainActorPool.ProcessSubActorPoolManager
+
+    class SubActorPoolManager(ABC):
```
hekaisheng, line 892: If there is a parent class for `ActorPool`, this class may not be necessary.

mars/oscar/backends/ray/communication.py
```diff
@@ -0,0 +1,340 @@
...
+            # Wait on ray object ref
+            object_ref = await self._in_queue.get()
+            return await object_ref
+        except RuntimeError as e:
+            if not self._closed.is_set():
+                raise e
+
+
+class RayOneWayActorChannel(RayChannelBase):
+    """A channel from ray actor to ray driver. Since ray actor can't call ray actor,
```
hekaisheng, line 149: Typo 'ray actor' -> 'ray driver'

mars/oscar/backends/ray/communication.py
```diff
@@ -0,0 +1,340 @@
...
+
+    # Multiple instances for ray local mode
+    _address_to_instances: Dict[str, "RayServer"] = dict()
+    scheme = 'ray'
+
+    def __init__(self, address, channel_handler: Callable[[Channel], Coroutine] = None):
+        super().__init__(DEFAULT_DUMMY_RAY_ADDRESS, channel_handler)
+        self._address = address
+        self._closed = asyncio.Event()
+        self._channels: Dict[ChannelID, RayTwoWayChannel] = dict()
```
hekaisheng, line 209: The server holds two different channels looks a bit weird, split into `DriverServer` and `ActorServer` seems better, how do you think? @qinxuye

hekaisheng (commented):
I left some comments about the actor pool's implementation, some details need more discussions.

mars/oscar/backends/mars/pool.py
chaokunyang: `SubActorPoolManager` is only used for `MainActorPool`, `SuhActorPool` doesn't need this

mars/oscar/backends/mars/pool.py
```diff
@@ -16,26 +16,29 @@
...
     CancelMessage, ControlMessage, ControlMessageType
 from .router import Router
+from ...api import Actor
+from ...core import ActorRef
+from ...errors import ActorAlreadyExist, ActorNotExist, ServerClosed
+from ...utils import create_actor_ref
+from ....utils import implements, get_next_port, to_binary
+from ....utils import lazy_import
+
+ray = lazy_import("ray")
```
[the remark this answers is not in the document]
chaokunyang: Used by `SubProcessHandle = Union[multiprocessing.Process, 'ray.actor.ActorHandle']`

mars/oscar/backends/ray/communication.py
chaokunyang: Because ray driver cab't be called by ray actors, ray driver doesn't have a`Server`

mars/oscar/backends/mars/pool.py
```diff
@@ -597,56 +603,13 @@ def _parse_config(config: Dict, kw: Dict) -> Dict:
...
-    return process
+SubProcessHandle = Union[multiprocessing.Process, 'ray.actor.ActorHandle']
 
 
 @_register_message_handler
 class MainActorPool(ActorPoolBase):
-    __slots__ = '_allocated_actors', '_sub_processes', \
-                '_subprocess_start_method', '_auto_recover', \
-                '_monitor_task', '_on_process_down', '_on_process_recover'
+    __slots__ = '_allocated_actors', '_subprocess_start_method', 'sub_actor_pool_manager', \
```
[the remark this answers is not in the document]
chaokunyang: I tried to move these basic classes into `mars/oscar/pool.py`, but it will almostly move everythings to `mars/oscar/pool.py`. Should we do it in this PR?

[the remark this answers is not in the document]
hekaisheng: I think so, one backend import something from other backend seems not very natural

mars/oscar/backends/mars/pool.py
hekaisheng: So the parent `MainActorPool` class has some methods to manage subprocesses.

mars/oscar/backends/ray/communication.py
hekaisheng: When connection happens, the channel type could be decided by the client, for example, actor or driver, and then create the corresponding channel.

mars/oscar/backends/mars/pool.py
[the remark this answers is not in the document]
hekaisheng: So I prefer to move common code to `mars/oscar/pool.py` to avoid importing ray in mars backend

mars/oscar/backends/ray/communication.py
chaokunyang: The point is that the ray actor will receive the requests from both the ray driver and ray actors. It needs different channels to send send messages back . When the request is from the ray driver, it use `RayOneWayActorChannel`;  When the request is from the ray actor, it use `RayTwoWayChannel`. The channel is decided by peer process address. If peer process is ray driver ,the address will be None, or else it will be the actor's name.

mars/oscar/backends/mars/pool.py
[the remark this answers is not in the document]
chaokunyang: Moved common code to `mars/oscar/backends`

chaokunyang:
@hekaisheng I've addressed all your comments, could you please review it again?

mars/oscar/backends/ray/pool.py
```diff
@@ -0,0 +1,165 @@
...
+    async def start(self, *args, **kwargs):
+        raise NotImplementedError
+
+    async def __on_ray_recv__(self, channel_id: ChannelID, message):
+        return await self.actor_pool.__on_ray_recv__(channel_id, message)
+
+    def health_check(self):  # noqa: R0201  # pylint: disable=no-self-use
+        return PoolStatus.HEALTHY
+
+    async def __proxy_call__(self, attribute, *args, **kwargs):
```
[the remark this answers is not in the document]
chaokunyang: used by `ProxyPool` to query attributes of `RayActorPool` running in Ray actors

mars/oscar/backends/ray/utils.py
```diff
@@ -28,26 +96,23 @@ def address_to_placement_group_bundle(address):
...
+    parts = list(reversed(parts))
+    return parsed_url.netloc, parts
+
+
+def process_placement_to_address(pg_name: str, bundle_index: int, process_index: int = 0):
+    return f"ray://{pg_name}/{bundle_index}/{process_index}"
 
 
-def placement_group_bundle_to_address(pg_name, bundle_index):
+def pg_bundle_to_node_address(pg_name, bundle_index):
```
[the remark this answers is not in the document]
chaokunyang: `node_placement_to_address` is a better name

mars/oscar/backends/ray/pool.py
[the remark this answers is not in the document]
fyrestone: It seems that the `ProxyPool` is only used in test cases. Is it possible to move the `__proxy_call__` to the test case?

mars/oscar/backends/ray/communication.py
```diff
@@ -0,0 +1,299 @@
...
+    def __init__(self,
+                 local_address: str = None,
+                 dest_address: str = None,
+                 channel_index: int = None,
+                 channel_id: ChannelID = None,
+                 compression=None):
+        super().__init__(local_address=local_address,
+                         dest_address=dest_address,
+                         compression=compression)
+        self._channel_index = channel_index or RayChannelBase.next_channel_index()
```
fyrestone, line 61:
```suggestion
        self._channel_index = channel_index or next(self._channel_index_gen)
```

mars/oscar/backends/ray/communication.py
```diff
@@ -0,0 +1,299 @@
...
+        await self._in_queue.put(message)
+        return await self._out_queue.get()
+
+
+@register_server
+class RayServer(Server):
+    __slots__ = '_closed', '_address', '_channels', '_tasks'
+
+    # Multiple instances for ray local mode
+    _address_to_instances: Dict[str, "RayServer"] = dict()
```
fyrestone, line 176: Why do you need this `_address_to_instances`?

mars/oscar/backends/ray/pool.py
```diff
@@ -0,0 +1,141 @@
...
+ray = lazy_import('ray')
+logger = logging.getLogger(__name__)
+
+
+class RayActorPoolMixin(AbstractActorPool, ABC):
+
+    async def __on_ray_recv__(self, channel_id: ChannelID, message):
+        """Method for communication based on ray actors"""
+        if not hasattr(self, '_external_servers'):
+            ray_servers = [server for server in self._servers if isinstance(server, RayServer)]
```
fyrestone, line 37: Why not set the `_external_servers` in `__init__`?

mars/oscar/backends/ray/utils.py
```diff
@@ -28,26 +90,23 @@ def address_to_placement_group_bundle(address):
...
+    return f"ray://{pg_name}/{bundle_index}/{process_index}"
 
 
-def placement_group_bundle_to_address(pg_name, bundle_index):
+def node_placement_to_address(pg_name, bundle_index):
     return f"ray://{pg_name}/{bundle_index}"
 
 
-def addresses_to_placement_group_info(address_to_resources):
+def node_addresses_to_pg_info(address_to_resources):
```
fyrestone, line 105: The `node_addresses_to_pg_info` is not clear, and the naming rules are inconsistent.

We can abstract some concepts: `node address`, `process address`, `placement group`, `placement group info`, `placement info`. Then naming the utils by these concepts.
- node_address_to_placement_info
- placement_info_to_node_address
- process_address_to_placement_info
- placement_info_to_process_address
- node_addresses_to_placement_group_info
- placement_group_info_to_node_addresses

It's just an example. The name is longer but clearer.

mars/oscar/backends/ray/communication.py
```diff
@@ -0,0 +1,299 @@
...
+    scheme = 'ray'
+
+    def __init__(self, address, channel_handler: Callable[[Channel], Coroutine] = None):
+        super().__init__(DEFAULT_DUMMY_RAY_ADDRESS, channel_handler)
+        self._address = address
```
fyrestone, line 181: What is `self._address` for?

mars/oscar/backends/ray/pool.py
chaokunyang: ActorPool constructor has too many arguments, it will make RayMainActorPool/RaySUbActorPool constructor compilicated and doesn't get too much gains

mars/oscar/backends/ray/communication.py
chaokunyang: Current server address, it's the same as current actor's name

mars/oscar/backends/ray/communication.py
chaokunyang: It's originally for TwoWayActorChannel, which can be removed now.

mars/oscar/backends/ray/communication.py
```diff
@@ -0,0 +1,299 @@
...
+        # Not that we can only send once for every read message in channel, otherwise
+        # it will be taken as other message's reply.
+        await self._out_queue.put(message)
+        self._msg_sent_counter += 1
+        assert self._msg_sent_counter <= self._msg_recv_counter,\
+            "One way channel doesn't support multiple replies for one message."
+
+    @implements(Channel.recv)
+    async def recv(self):
+        if self._closed.is_set():  # pragma: no cover
```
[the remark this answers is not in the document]
chaokunyang: good catch

mars/oscar/backends/ray/pool.py
chaokunyang: fixed

mars/oscar/backends/ray/communication.py
fyrestone: But, I can't see where uses the `self._address`.

mars/oscar/backends/ray/communication.py
```diff
@@ -0,0 +1,278 @@
...
+    __slots__ = '_peer_actor',
+
+    def __init__(self,
+                 dest_address: str = None,
+                 channel_index: int = None,
+                 channel_id: ChannelID = None,
+                 compression=None):
+        super().__init__(None, dest_address, channel_index, channel_id, compression)
+        # ray actor should be created with the address as the name.
+        self._peer_actor: 'ray.actor.ActorHandle' = ray.get_actor(dest_address) if dest_address else None
```
fyrestone, line 97: The `RayClientChannel` should asserts that the peer actor is not None. So, just `self._peer_actor = ray.get_actor(dest_address)`?

mars/oscar/backends/ray/communication.py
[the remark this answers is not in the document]
fyrestone: Then the `self._closed.is_set()` should not throw an exception in `recv` and `send`?

mars/oscar/backends/ray/communication.py
```diff
@@ -0,0 +1,278 @@
...
+    @property
+    @implements(Server.stopped)
+    def stopped(self) -> bool:
+        return self._closed.is_set()
+
+    async def __on_ray_recv__(self, channel_id: ChannelID, message):
+        channel = self._channels.get(channel_id)
+        if not channel:
+            self._tasks[channel_id] = asyncio.create_task(self.on_connected(channel_id))
+        while not channel:
```
fyrestone, line 244: It's dangerous to do the `while loop` to get the channel. If the `self.on_connected` raises an exception before set the channel to `self._channels`, then the loop will not break, because the `self.on_connected` is running as an asyncio task.

Two ways to solve this problem:
1. Break the loop if the `self._tasks[channel_id]` is not running.
2. Returns the channel from `self.on_connected`, then runs the `self.channel_handler` as an asyncio task in `self.on_connected`.

mars/oscar/backends/ray/communication.py
```diff
@@ -0,0 +1,278 @@
...
+        super().__init__(local_address, None, channel_index, channel_id, compression)
+        self._out_queue = asyncio.Queue()
+        self._msg_recv_counter = 0
+        self._msg_sent_counter = 0
+
+    @implements(Channel.send)
+    async def send(self, message: Any):
+        if self._closed.is_set():  # pragma: no cover
+            raise ChannelClosed('Channel already closed, cannot send message')
+        # Current process is ray actor, we use ray call reply to send message to ray driver/driver.
```
fyrestone, line 141:
```suggestion
        # Current process is ray actor, we use ray call reply to send message to ray driver/actor.
```

mars/oscar/backends/ray/communication.py
chaokunyang: The `on_connected` will run forever, you can't return channel from it

chaokunyang: Create channel outside seems better, just like `DummyServer`

mars/oscar/backends/ray/communication.py
```diff
@@ -0,0 +1,282 @@
...
+        self._msg_recv_counter += 1
+        await self._in_queue.put(message)
+        task = asyncio.create_task(self._out_queue.get())
+        while True:  # pragma: no cover
+            # Avoid hang when channel is closed after `self._out_queue.get()` is awaited.
+            done, pending = await asyncio.wait([task], timeout=0.1)
+            if self._closed.is_set():
+                raise ChannelClosed('Channel already closed')
+            if done:
+                return await task
```
fyrestone, line 168: Is it simpler to ?
```python
done, pending = asyncio.wait([self._out_queue.get(), self._closed.wait()], return_when=asyncio.FIRST_COMPLETED)
if self._closed.is_set():
    raise ChannelClosed('Channel already closed')
else:
    return await done[0]  # May be it's safer to check that the done is the `self._out_queue.get()`.
```

chaokunyang: `self._out_queue.get()` will never finish when When using `asyncio.FIRST_COMPLETED` and there is an exception throwed in `channel send` or `channel_handler` , `self._out_queue.get()` will never return. That's why we use `wait` and `timeout`.
`asyncio.FIRST_COMPLETED` here has same effect with `await self._out_queue.get()` which is we try to avoid here.

chaokunyang: How do you check `May be it's safer to check that the done is the `self._out_queue.get()`.`?

fyrestone:
``` python
# pseudo-code
task = create_task(self._out_queue.get())
done, pending = asyncio.wait([task, self._closed.wait()], ...)
# check self._close.is_set
if task in done:
    return await task.
else:
    raise Excepion
```

But, just `await done[0]` is OK. if both of `self._closed.wait()` and `self._out_queue.get()` are done, then we raises ChannelClosed('Channel already closed').

mars/oscar/backends/ray/communication.py
```diff
@@ -0,0 +1,281 @@
...
+            task.cancel()
+        self._tasks = dict()
+
+    @property
+    @implements(Server.stopped)
+    def stopped(self) -> bool:
+        return self._closed.is_set()
+
+    async def __on_ray_recv__(self, channel_id: ChannelID, message):
+        channel = self._channels.get(channel_id)
```
fyrestone, line 244: Is the `__on_ray_recv__` should be blocked if the `self._closed.is_set()`?

fyrestone (approved):
LGTM.

mars/oscar/backends/ray/communication.py
chaokunyang: if the `self._closed.is_set()`, `__on_ray_recv__ ` will throw `ChannelClosed` exception

fyrestone: If `__on_ray_recv__` is called after the `RayServer` is stopped, the `self._channels` will be empty, it will create a new session.

fyrestone: I think it's better to raise an exception or just pass the `__on_ray_recv__ ` logic here if the `self._closed.is_set()`.

mars/oscar/backends/communication/core.py
```diff
@@ -43,25 +42,28 @@ def _check_scheme(scheme: str, types: Dict):
...
-        scheme = None
-    else:
-        scheme = urlparse(address).scheme
-    scheme = _check_scheme(scheme, _scheme_to_server_types)
+    scheme = _check_scheme(get_scheme(address), _scheme_to_server_types)
     return _scheme_to_server_types[scheme]
 
 
-def gen_internal_address(process_index: int) -> str:
+def gen_internal_address(process_index: int,
```
qinxuye, line 63: `get_internal_address` looks a bit weird to be here, it should be backend related.

mars/oscar/backends/communication/dummy.py
```diff
@@ -17,10 +17,10 @@
 from typing import Any, Callable, Coroutine, Dict, Type
 from urllib.parse import urlparse
 
-from .....utils import implements, classproperty
 from .base import Channel, ChannelType, Server, Client
 from .core import register_client, register_server
 from .errors import ChannelClosed
+from ....utils import implements, classproperty
```
qinxuye, line 23: Why move such line?

mars/oscar/backends/pool.py
```diff
@@ -0,0 +1,1040 @@
...
+    ResultMessage, ErrorMessage, CreateActorMessage, HasActorMessage, \
+    DestroyActorMessage, ActorRefMessage, SendMessage, TellMessage, \
+    CancelMessage, ControlMessage, ControlMessageType
+from .router import Router
+from ..api import Actor
+from ..core import ActorRef
+from ..errors import ActorAlreadyExist, ActorNotExist, ServerClosed
+from ..utils import create_actor_ref
+from ...utils import implements, to_binary
+from ...utils import lazy_import
```
qinxuye, line 38: Normally, we import as the order from outer to inner.

mars/oscar/backends/ray/backend.py
```diff
@@ -12,10 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from ...backend import BaseActorBackend, register_backend
 from .driver import RayActorDriver
-from .context import RayActorContext
-
+from ..mars.context import MarsActorContext
```
qinxuye, line 16: Maybe implement a RayActorContext, even though they are identical.

mars/oscar/backends/ray/utils.py
```diff
@@ -12,11 +12,68 @@
...
+        return ray.util.get_placement_group(pg_name)
+    else:
+        logger.warning("Current installed ray version doesn't support named placement group. "
+                       "Actor will be created on arbitrary node randomly.")
+        return None
+
 
+def process_address_to_placement(address):
+    """
+    Args:
```
qinxuye, line 38: We use numpy-style document.

mars/oscar/backends/router.py
```diff
@@ -37,7 +37,7 @@ def get_instance() -> "Router":
 
     @staticmethod
     def get_instance_or_empty() -> "Router":
-        return Router._instance or _empty_router
+        return Router._instance or Router(list(), None)
```
qinxuye, line 40: Why remove`_empty_router`?

mars/oscar/backends/mars/backend.py
```diff
@@ -40,6 +41,6 @@ async def create_actor_pool(
             address: str,
             n_process: int = None,
             **kwargs):
-        from .pool import create_actor_pool
+        from ..pool import create_actor_pool
```
qinxuye, line 44: Seems I didn't see the counterpart `create_actor_pool` for Ray backend, am I miss it or not?

qinxuye (commented):
Looks fantastic overall, I left some comments about some small issues.

mars/oscar/backends/communication/core.py
chaokunyang: Make it as a abstract method of `MainActorPool`

mars/oscar/backends/pool.py
chaokunyang: I use pycharm to format imports. Is there a way to format imports automatically?

mars/oscar/backends/ray/backend.py
chaokunyang: MarsActorContext is pretty much code, about 150 lines. Do we need to duplicate this. I mean there won't be any difference for `RayActorContext`. All difference introduced by this PR are `actor pool` and `communication` level

mars/oscar/backends/mars/backend.py
chaokunyang: `create_actor_pool` is located at `mars/oscar/backends/mars/backend.py` and shared by mars and ray backend

mars/oscar/backends/router.py
chaokunyang: `_empty_router` is global and hold references of previous tests such as `_curr_external_addresses`/`_local_mapping`/`_mapping`/`_cache` if we use same driver.
BTW, I think global state should be avoided as much as possible.

mars/oscar/backends/ray/backend.py
chaokunyang: Moved `MarsActorContext` to `mars/oscar/backends/context.py`

mars/oscar/backends/router.py
chaokunyang: On ray, this will cause later tests crash

qinxuye (approved):
LGTM

PR #2064, opened 2021-03-25
Fix main pool that creating servers only when sub pools finished creation

qinxuye:
## What do these changes do?

This PR fixes main pool that creating servers only when sub pools finished creation.

Before, because creating servers is quicker than forking sub processes, so the servers on main pool will be forked.


wjsi (approved):
LGTM

PR #2072, opened 2021-03-29
Fix batch get / delete object id and stop launching plasma when working on vineyard

sighingnow:
## What do these changes do?

+ Fixes the bug in batch get/delete vineyard object ID
+ Don't launch plasma if vineyard is enabled.

## Related issue number

This pull request may could fixes recent random failures in CI.


wjsi (approved):
LGTM

PR #2077, opened 2021-03-31
Use a ref count to make delete works when multiple workers connect to the same vineyardd

sighingnow:
Add a log to `delete` as well to find out what happens when failed.

Fixes the failure in CI: https://github.com/mars-project/mars/runs/2233774757?check_suite_focus=true


wjsi (approved):
LGTM

PR #2078, opened 2021-03-31
[BACKPORT] Use a ref count to make delete works when multiple workers connect to the same vineyard (#2077)

wjsi:
Backport of #2077


qinxuye (approved):
LGTM

PR #2086, opened 2021-04-09
Use chunk index as default shuffle key

wjsi:
## What do these changes do?

Use chunk index as default shuffle key, also rename variables for shuffle.


qinxuye (approved):
LGTM

PR #2095, opened 2021-04-14
Fix coverage result on SubActorPool

wjsi:
## What do these changes do?

Shutdown subprocesses gracefully to let coverage data being recorded.


qinxuye (approved):
LGTM

PR #2096, opened 2021-04-15
Fix result error when yield after exceptions

wjsi:
## What do these changes do?

Fix result error when yield after exceptions.


qinxuye (approved):
LGTM

PR #2099, opened 2021-04-20
Implement last_idle_time API

fyrestone:
## What do these changes do?

Implement a `last_idle_time` API which returns the last idle time. The last idle time can only be updated when the `last_idle_time` API is called.

- Add `SessionAPI.last_idle_time(session_id: Union[str, None] = None) -> Union[float, None]`.
- Add `TaskAPI.last_idle_time() -> Union[float, None]`.

## Related issue number
N/A


qinxuye (approved):
LGTM

PR #2102, opened 2021-04-21
Services web handler and api

chaokunyang:
## What do these changes do?
Implements services web api and remote session. The code changes includes:
* service web api framework
* task/session/meta/storage web handler/api
* web-api based remote session

## Related issues
#2112


mars/services/web/core.py
```diff
@@ -0,0 +1,98 @@
...
+    _web_address_local.web_address = web_address
+
+
+def get_web_address():
+    assert hasattr(_web_address_local, 'web_address'), 'Please set web_address first before get web address'
+    return _web_address_local.web_address
+
+
+class ServiceWebHandlerBase(MarsRequestHandler):
+    _api_instances = dict()
```
fyrestone, line 51: Where to destroy the api instance?

mars/services/web/core.py
```diff
@@ -0,0 +1,98 @@
...
+    def __init__(self, http_client: AsyncHTTPClient, service_name: str, api_cls, api_id: int):
+        self._http_client = http_client
+        self._service_name = service_name
+        self._api_cls = api_cls
+        self._api_id = api_id
+
+    def __getattr__(self, method_name):
+        async def _func(*args, **kwargs):
+            method = getattr(self._api_cls, method_name)
+            arg_spec = inspect.getfullargspec(method)
```
fyrestone, line 87: We can use the `inspect.signature` to validate the args and kwargs in  client side.

mars/services/web/core.py
```diff
@@ -0,0 +1,98 @@
...
+    _api_instances = dict()
+
+    async def post(self, path, **kwargs):
+        api_method_name = path
+        params = deserialize(self.request.body)
+        try:
+            if hasattr(self, api_method_name):
+                method = getattr(self, api_method_name)
+            else:
+                api_id, params = params[0], params[1:]
```
fyrestone, line 60: If the params is the `(api_id, args, kwargs)`, we can assign the params to them.
```python
api_id, args, kwargs = params
```

mars/services/web/core.py
```diff
@@ -0,0 +1,98 @@
...
+                method = getattr(self._api_instances[api_id], api_method_name)
+            arg_spec = inspect.getfullargspec(method)
+            if arg_spec.varkw:
```
fyrestone, line 63: Why treat the method with kwargs separately?

chaokunyang: All kwargs related code has been removed

mars/services/web/core.py
chaokunyang: Will use reference count to send destroy request to server to destroy the instance

chaokunyang: See `mars.services.web.core.ServiceWebHandlerBase.__destroy_api__`

mars/services/web/core.py
```diff
@@ -0,0 +1,145 @@
...
+                # Some methods are decorated with @alru_cache, which doesn't support dict as part of cache key.
+                result = method(*args, **kwargs)
+            else:
+                result = method(*args)
+            if inspect.iscoroutine(result):
+                result = await result
+            self.write(serialize(result))
+        except Exception as e:
+            logger.info(f'Execute method {api_method_name} with {api_id, args, kwargs} failed, got exception {e}')
+            traceback.format_exc()
```
fyrestone, line 86: What is `traceback.format_exc()` for?

mars/services/web/core.py
```diff
@@ -0,0 +1,145 @@
...
+            exc_type, exc_value, exc_traceback = sys.exc_info()
+            self.write(serialize(_HandlerException(exc_type, exc_value, exc_traceback)))
```
fyrestone, line 88: I think it's better to serialize the `_HandlerException ` with a formatted traceback string.

mars/services/web/core.py
```diff
@@ -0,0 +1,145 @@
...
+    def __del__(self):
+        try:
+            self._sync_post('__destroy_api__', api_id=self._api_id, req_config=dict(timeout=(0.1, 0.5)))
+        except:  # noqa: E722  # nosec  # pylint: disable=bare-except  # pragma: no cover
+            # server is closed
+            pass
+
+    @classmethod
+    async def _post(cls, http_client: AsyncHTTPClient, api_method_name: str,
+                    *args, api_id=None, req_config=None, **kwarg):
```
fyrestone, line 117: It's wired to put the parameter`api_id` and `req_config` between `*args` and `**kwargs`.

mars/services/web/core.py
```diff
@@ -0,0 +1,145 @@
...
+            if not api_id:
+                if hasattr(self, api_method_name):
+                    method = getattr(self, api_method_name)
+                else:
+                    method = getattr(self._api_cls, api_method_name)
+            else:
+                # call method of api instance
+                method = getattr(self._api_instances[api_id], api_method_name)
+            if kwargs:
+                # Some methods are decorated with @alru_cache, which doesn't support dict as part of cache key.
```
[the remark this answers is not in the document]
chaokunyang: It's not a typo, see `mars.lib.aio.lru.alru_cache`

mars/services/web/core.py
chaokunyang: I prefer returning raw exception info

mars/services/web/core.py
chaokunyang: The exception hanppens in the server. Print the traceback can make troubleshooting much more easy.

mars/services/web/core.py
```diff
@@ -0,0 +1,145 @@
...
+        self._http_client = http_client
+        self._api_id = api_id
+
+    def __getattr__(self, method_name):
+        async def _func(*args, **kwargs):
+            return await self._post(self._http_client, method_name, *args, api_id=self._api_id, **kwargs)
+
+        return _func
+
+    def __del__(self):
```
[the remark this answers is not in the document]
chaokunyang: If the api instance is evicted from the cache, the current request will fail and the client will need to repost the constructors params to recreate the api instance. After the instance is recreated, the client need to repost the previous request. I think thw whole process is complicated and the gains seems not very big. Considering we'll have only one seesion for a mars cluster, which is to say we will have only one web session and one group of web api, thus one group of api instance. Leaking won't be a big problem. I think adding some comments to `__del__ ` would be enough. @qinxuye @wjsi What do you think?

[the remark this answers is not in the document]
chaokunyang: If we do need destroy api instance, we can add a heartbeat mechanism between service web api and web handers if we do need this.

mars/services/web/core.py
fyrestone: `traceback.format_exc()` does not print the exception. Please refer to the Python doc: https://docs.python.org/3/library/traceback.html#traceback.format_exc

mars/services/web/core.py
fyrestone: I am not sure if the `serialize` can handle all types of call frames (traceback + globals + locals of each frame).

mars/services/web/core.py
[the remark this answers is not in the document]
fyrestone: Currently, the Web API is stateful, so using the heartbeat to check the client session alive is a good choice. If we can reconstruct the api instance in server side, then the Web API is stateless which is easier to use.

mars/services/web/core.py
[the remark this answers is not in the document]
fyrestone: The `mars.lib.aio.lru.alru_cache` supports kwargs. Why do you treat kwargs differently?

```python
@alru_cache
async def foo(a, b):
    print(a, b)

await foo(**{"a": 1, "b": 2}) # OK, **kwargs is good.
await foo(*(1, 2)) # OK, *args is good.
await foo(1, *(2,), **{}) #OK, mixed *args and **kwargs is good.
await foo(1, b=2) # OK.
await foo(a=1, b=2) # OK.
```

mars/deploy/oscar/ray.py
```diff
@@ -124,7 +129,7 @@ class RayClient:
     def __init__(self,
                  cluster: RayCluster,
                  session: AbstractSession):
-        self._cluster = cluster
+        self.cluster = cluster
```
fyrestone, line 132: Why change `self._cluster` to `self.cluster`?

mars/services/web/core.py
chaokunyang: My mistake, changed to `print_exc`

mars/services/web/core.py
chaokunyang: This is consisitent with `mars.oscar.backends.pool._ErrorProcessor`. If the `mars.oscar.backends.pool._ErrorProcessor` can handle all types, this should do tooo.

mars/services/meta/web.py
```diff
@@ -0,0 +1,57 @@
...
+
+class MetaWebHandler(ServiceWebHandlerBase):
+    _api_cls = MetaAPI
+
+    async def create(self, session_id: str, address: str):
+        api_instance = await MetaAPI.create(session_id, address)
+        self._api_instances[id(api_instance)] = api_instance
+        return id(api_instance)
+
+    async def create_session(self, session_id: str, address: str):
```
fyrestone, line 28: What's the difference between `create` and `create_session`? These two methods are the same.

mars/services/web/core.py
```diff
@@ -0,0 +1,143 @@
...
+        req_config = req_config or dict()
+        if 'request_timeout' not in req_config:
+            req_config['request_timeout'] = 2 * 60 * 60  # timeout for two hours
+        resp = await http_client.fetch(f'{get_web_address()}/api/service/{cls._service_name}/{api_method_name}',
+                                       method="POST", body=cls._serialize_args(api_id, args, kwarg),
+                                       **(req_config or dict()))
+        return cls._deserialize_result(resp.body)
+
+    @classmethod
+    def _sync_post(cls, api_method_name: str, api_id, req_config, *args, **kwarg):
```
fyrestone, line 126: Where is the `_sync_post` method used?

mars/services/web/core.py
```diff
@@ -0,0 +1,143 @@
...
+    _api_cls = None
+
+    async def post(self, path, **_):
+        api_method_name = path
+        # params format: api_id[None], args, kwargs
+        api_id, args, kwargs = deserialize(self.request.body)
+        try:
+            if not api_id:
+                if hasattr(self, api_method_name):
+                    method = getattr(self, api_method_name)
```
fyrestone, line 70: The method in `self` shadows the method in `self._api_cls`. It's too hack.

mars/services/web/core.py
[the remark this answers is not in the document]
fyrestone: You did nothing about this comment. Why did you mark it as resolved?

mars/services/web/core.py
```diff
@@ -0,0 +1,143 @@
...
+                # Some methods are decorated with @alru_cache, which doesn't support dict as part of cache key.
+                result = method(*args, **kwargs)
+            else:
+                result = method(*args)
+            if inspect.iscoroutine(result):
+                result = await result
+            self.write(serialize(result))
+        except Exception as e:
+            logger.info(f'Execute method {api_method_name} with {api_id, args, kwargs} failed, got exception {e}')
+            traceback.print_exc()
```
fyrestone, line 86: We can use `logger.exception` replaces `logger.info` + `traceback.print_exc()`.

mars/services/meta/web.py
chaokunyang: This is a typo, `create_session` should invoke `create_session` of `MetaAPI`

mars/services/web/core.py
chaokunyang: It's designed by purpuse, so that we can add an isntance method to overrite default implementation in `_api_cls`

mars/services/web/core.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+_expired_sec = 60
+_keep_alive_interval = 3
+
+
+class ApiRegistry:
+
+    def __init__(self, expired_sec):
+        self._apis = dict()
```
fyrestone, line 67: Every API instance belongs to a session, Is it possible to use session id as the key of api instance?
```python
{"session id": {TaskAPI: instance}}
```
Then, all web client shares the same instance for the same API class. We can recycle the session instances through last idle time.

mars/services/web/core.py
fyrestone: It's better to make another API for calling the method in `self`.

mars/deploy/oscar/session.py
```diff
@@ -78,19 +81,21 @@ async def init(cls,
             from .local import new_cluster
             return (await new_cluster(address, **kwargs)).session
 
-        session_api = await SessionAPI.create(address)
+        web_address = kwargs.pop('web_address', None)
```
wjsi, line 84: I don't think it a nice design to have a separate ``web_address`` argument. Simply parse ``address`` and check the protocol part instead.

mars/deploy/oscar/session.py
```diff
@@ -78,19 +81,21 @@ async def init(cls,
...
+        if web_address:
+            set_web_address(web_address)
```
wjsi, line 86: Thread-local is not a good way to pass arguments.

mars/deploy/oscar/session.py
```diff
@@ -78,19 +81,21 @@ async def init(cls,
...
+        session_api = await (SessionWebAPI if web_address else SessionAPI).create(address)
         # create new session
         session_address = await session_api.create_session(session_id)
-        meta_api = await MetaAPI.create(session_id, session_address)
-        task_api = await TaskAPI.create(session_id, session_address)
+        meta_api = await (MetaWebAPI if web_address else MetaAPI).create(session_id, session_address)
```
wjsi, line 90: We would like to implement separate ``MetaOscarAPI`` and ``MetaWebAPI`` basing on an abstract class ``MetaAPI``. ``create`` method will construct both subclasses given different types of addresses.

mars/services/meta/web.py
```diff
@@ -0,0 +1,53 @@
...
+    _service_name = _service_name
+
+    @classmethod
+    async def create(cls, session_id: str, address: str):
+        http_client = AsyncHTTPClient()
+        api_id = await cls._post(http_client, 'create', None, {}, session_id, address)
+        return MetaWebAPI(http_client, api_id)
+
+    @classmethod
+    async def create_session(cls, session_id: str, address: str):
```
wjsi, line 46: ``create_session`` shall not be exposed as a separate web API. It is called by session service internally.

mars/services/meta/web.py
```diff
@@ -0,0 +1,53 @@
...
+        http_client = AsyncHTTPClient()
+        api_id = await cls._post(http_client, 'create_session',  None, {}, session_id, address)
+        return MetaWebAPI(http_client, api_id)
+
+    @classmethod
+    async def destroy_session(cls, session_id: str, address: str):
+        return await cls._post(AsyncHTTPClient(), 'destroy_session',  None, {}, session_id, address)
```
wjsi, line 53: Ditto.

mars/services/web/core.py
```diff
@@ -0,0 +1,175 @@
...
+        self.exc_traceback = exc_traceback
+
+
+_expired_sec = 60
+_keep_alive_interval = 3
+
+
+class ApiRegistry:
+
+    def __init__(self, expired_sec):
```
wjsi, line 66: Do we **REALLY** need to cache API instances by time?

mars/services/web/supervisor.py
```diff
@@ -86,13 +87,15 @@ def __init__(self, config):
...
         self._web_server = None
+        self._web_address = None
 
     async def start(self):
         static_path = os.path.join(os.path.dirname(__file__), 'static')
         supervisor_addr = self.address
 
-        host = self._config.get('host') or '0.0.0.0'
-        port = self._config.get('port')
+        host = self._config.get('host', socket.gethostbyname(socket.gethostname()))
```
wjsi, line 96: This may not work when DNS is not accessible.

mars/services/web/core.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+class ServiceWebAPIBase:
+    _service_name = None
+
+    def __init__(self, http_client: AsyncHTTPClient, api_id: int):
+        self._http_client = http_client
+        self._api_id = api_id
+        # TODO(chaokunyang) how to cancel this task
+        self._keep_api_alive_task = asyncio.create_task(self._keep_api_alive())
```
wjsi, line 143: By putting keep-alive request in every Web API, unnecessary heart beats will be sent given num of API instances, which is not needed.

mars/services/web/core.py
```diff
@@ -0,0 +1,175 @@
...
+        async def _func(*args, **kwargs):
+            return await self._post(self._http_client, method_name, self._api_id, {}, *args, **kwargs)
+
+        return _func
+
+    @classmethod
+    async def _post(cls, http_client: AsyncHTTPClient, api_method_name: str, api_id, req_config, *args, **kwarg):
+        req_config = req_config or dict()
+        if 'request_timeout' not in req_config:
+            req_config['request_timeout'] = 2 * 60 * 60  # timeout for two hours
```
wjsi, line 160: Try avoiding magic numbers in code. Add a constant here.

mars/services/meta/web.py
```diff
@@ -0,0 +1,53 @@
...
+    async def create(self, session_id: str, address: str):
+        return self._api_registry.add_instance(await MetaAPI.create(session_id, address))
+
+    async def create_session(self, session_id: str, address: str):
+        return self._api_registry.add_instance(await MetaAPI.create_session(session_id, address))
+
+
+_service_name = 'meta'
+web_handlers = {
+    f'/api/service/{_service_name}/(.*)': MetaWebHandler,
```
wjsi, line 32: This API design is not RESTful.

mars/services/web/core.py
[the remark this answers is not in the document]
chaokunyang: Changed to creating api everytime

mars/services/web/core.py
chaokunyang: Removed api instance

mars/services/web/supervisor.py
chaokunyang: Added `mars.utils.get_node_ip_address` implementation from ray

.github/workflows/core-ci.yml
```diff
@@ -73,7 +73,7 @@ jobs:
             pip install statsmodels tsfresh
           fi
           virtualenv testenv && source testenv/bin/activate
-          pip install -r requirements.txt && pip install pytest pytest-timeout
+          pip install -r requirements.txt && pip install -r requirements-extra.txt && pip install pytest pytest-timeout
```
wjsi, line 76: This group of tests if used to test if Mars can work under minimal requirements. Thus no extra requirements should be installed.

chaokunyang: `bokeh` is needed when import `service` api. Or can we add `bokeh` to `requirement.txt`

wjsi: Then try to make it optional.

mars/services/session/api.py
```diff
@@ -159,5 +159,5 @@ async def create_session(self, session_id: str) -> str:
     async def delete_session(self, session_id: str):
         return await self._call_method({}, 'delete_session', session_id)
 
-    async def last_idle_time(self, session_id: Union[str, None] = None) -> Union[float, None]:
+    async def get_last_idle_time(self, session_id: Union[str, None] = None) -> Union[float, None]:
         return await self._call_method({}, 'last_idle_time', session_id)
```
fyrestone, line 163: Should be `return await self._call_method({}, 'get_last_idle_time', session_id)`? It's not necessary to change the API name, the name `last_idle_time` is clear enough.

wjsi: @fyrestone We recommend all method names to be verb phrases.

wjsi (approved):
LGTM

PR #2109, opened 2021-04-26
Wrap sync method in session so that they will be running in threads

### LINKED ISSUE #2108: [BUG] Can't create session when there's an event loop running
**Describe the bug**
When there is en event loop running, create session by `mars.core.session.new_session` will fail.

**To Reproduce**
1. Python version: py3.8
2. Mars version: master
3. numpy 1.20.1  pandas 1.2.3 scipy 1.6.1
4. Full stack of the error:
```
RuntimeError                              Traceback (most recent call last)
<ipython-input-4-801ae027078e> in <module>
     29     assert get_default_session() is None
     30 
---> 31 test_sync_execute()

<ipython-input-4-801ae027078e> in test_sync_execute()
      5     start_method = os.environ.get('POOL_START_METHOD', 'forkserver') \
      6         if sys.platform != 'win32' else None
----> 7     session = new_session(n_cpu=2, subprocess_start_method=start_method,
      8                           default=True)
      9 

~/ant/Development/DevProjects/python/mars/mars/core/session.py in new_session(address, session_id, backend, default, **kwargs)
    394         if 'init_local' not in kwargs:
    395             kwargs['init_local'] = True
--> 396     session = _loop.run_until_complete(
    397         _new_session(address, session_id=session_id,
    398                      backend=backend, default=default, **kwargs))

~/.pyenv/versions/3.8.1/lib/python3.8/asyncio/base_events.py in run_until_complete(self, future)
    597         future.add_done_callback(_run_until_complete_cb)
    598         try:
--> 599             self.run_forever()
    600         except:
    601             if new_task and future.done() and not future.cancelled():

~/.pyenv/versions/3.8.1/lib/python3.8/asyncio/base_events.py in run_forever(self)
    554             raise RuntimeError('This event loop is already running')
    555         if events._get_running_loop() is not None:
--> 556             raise RuntimeError(
    557                 'Cannot run the event loop while another loop is running')
    558         self._set_coroutine_origin_tracking(self._debug)

RuntimeError: Cannot run the event loop while another loop is running
```
```
import sys, os
from mars.core.session import get_default_session, \
    new_session, execute, fetch, stop_server
def test_sync_execute():
    start_method = os.environ.get('POOL_START_METHOD', 'forkserver') \
        if sys.platform != 'win32' else None
    session = new_session(n_cpu=2, subprocess_start_method=start_method,
                          default=True)

    with session:
        raw = np.random.RandomState(0).rand(10, 5)
        a = mt.tensor(raw, chunk_size=5).sum(axis=1)
        b = a.execute(show_progress=False)
        assert b is a
        result = a.fetch()
        np.testing.assert_array_equal(result, raw.sum(axis=1))

        c = b + 1
        c.execute(show_progress=False)
        result = c.fetch()
        np.testing.assert_array_equal(result, raw.sum(axis=1) + 1)

        c = mt.tensor(raw, chunk_size=5).sum()
        d = session.execute(c)
        assert d is c
        assert abs(session.fetch(d) - raw.sum()) < 0.001

    session.stop_server()
    assert get_default_session() is None

test_sync_execute()
```

**Expected behavior**
Whether there is an event loop running or not, creating session should always secceed.

**Additional context**
Running pandas in notebook is very important for engineers, mars should support too.

qinxuye:
## What do these changes do?

This PR wraps sync method in session so that they will be running in threads.

## Related issue number

Fixes #2108 .

### DIFF
--- a/mars/core/session.py
+++ b/mars/core/session.py
@@ -17,7 +17,7 @@ import threading
 import uuid
 import warnings
 from abc import ABC, abstractmethod
-from typing import Dict, List, Type, Tuple, Union
+from typing import Callable, Dict, List, Type, Tuple, Union
 
 from ..config import options
 from ..core import TileableGraph, enter_mode
@@ -27,6 +27,40 @@ from .typing import TileableType
 
 _loop = asyncio.new_event_loop()
 _get_session_lock = asyncio.Lock(loop=_loop)
+
+
+def _wrap_in_thread(func: Callable):
+    """
+    Function is those wrapping async function,
+    when there is a running event loop,
+    error will raise (see GH#2108),
+    so we wrap them to run in a thread.
+    """
+
+    def inner(*args, **kwargs):
+        result = None
+        default_session_in_thread = None
+        default_session = get_default_session()
+
+        def run_in_thread():
+            nonlocal result, default_session_in_thread, default_session
+
+            if default_session:
+                # set default session in this thread
+                default_session.as_default()
+
+            result = func(*args, **kwargs)
+            default_session_in_thread = get_default_session()
+
+        t = threading.Thread(target=run_in_thread)
+        t.start()
+        t.join()
+
+        if default_session_in_thread:
+            default_session_in_thread.as_default()
+
+        return result
+    return inner
 
 
 class ExecutionInfo(ABC):
@@ -181,6 +215,17 @@ class AbstractSession(ABC):
         data
         """
 
+    def decref(self, *tileables):
+        """
+        Decref tileables.
+
+        Parameters
+        ----------
+        tileables
+            Tileables.
+        """
+        # TODO(qinxuye): implement this function when lifecycle service ready.
+
     async def stop_server(self):
         """
         Stop server.
@@ -228,6 +273,7 @@ def get_default_session() -> AbstractSession:
     return AbstractSession.default
 
 
+@_wrap_in_thread
 def stop_server():
     if AbstractSession.default:
         SyncSession(AbstractSession.default).stop_server()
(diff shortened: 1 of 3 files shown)


mars/core/session.py
```diff
@@ -29,6 +29,40 @@
...
+
+        t = threading.Thread(target=run_in_thread)
+        t.start()
+        t.join()
+
+        if default_session_in_thread:
+            default_session_in_thread.as_default()
+
+        return result
+    return inner
```
fyrestone, line 63: One thread per call? I think we can create only one daemon thread for the event loop. Then send the async call to the event loop and get the result.

```python
import asyncio
import threading

loop = asyncio.new_event_loop()
thrd = threading.Thread(target=loop.run_forever)
thrd.daemon = True
thrd.start()


async def async_call():
    return 1


def sync_call():
    return asyncio.run_coroutine_threadsafe(async_call(), loop).result()


print(sync_call())
```

qinxuye: Now I used ThreadPoolExecutor instead, check the latest commit to see if it's ok?

fyrestone: There is only one thread in the `ThreadPoolExecutor` to ensure that the loop interface call will not have thread safety issues. It's OK.

Is it OK to copy the default session between main thread and worker pool thread? It seems that the thread local var `AbstractSession._default_session_local` in multiple threads hold one same session instance?

qinxuye: Yeah, it should be in the same thread, due to the reason of event loop, they have to be in a different one, thus the default session should be synced in and out the thread.

fyrestone (dismissed):
LGTM

wjsi (approved):
LGTM

PR #2110, opened 2021-04-26
Make Mars pools compatible with Python 3.6

wjsi:
## What do these changes do?

Use backported asyncio in Python 3.7 to run Mars pools. Also support shared memory storage for Python 3.6 and 3.7. Tests for Windows not covered.


qinxuye (approved):
LGTM

PR #2114, opened 2021-04-27
Fix pool cases and shared memory cases in Windows

wjsi:
## What do these changes do?

This PR resolves 3 problems:

1. Fix shared memory usage in Windows
2. Fix oscar cases in Windows
3. Add references for plasma reader & writer


qinxuye (approved):
LGTM

PR #2116, opened 2021-04-28
Fix `_get_ports_from_netstat` hang

### LINKED ISSUE #2115: [BUG] mars.utils._get_ports_from_netstat hang
**Describe the bug**
`mars.utils._get_ports_from_netstat` hangs on `subprocess.Popen.wait`. It's documented in [Popen.wait](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait)

> Popen.wait():
> 
> Wait for child process to terminate. Set and return returncode attribute.
> 
> Warning: This will deadlock when using stdout=PIPE and/or stderr=PIPE and the child process generates enough output to a pipe such that it blocks waiting for the OS pipe buffer to accept more data. Use communicate() to avoid that.
> 

**To Reproduce**
1. Your Python version:python 3.8
2. The version of Mars you use: mars master
3. Versions of crucial packages, such as numpy, scipy and pandas: latest
4. Full stack of the error.
image
Call `mars.utils._get_ports_from_netstat` when there're plenty of unused ports

chaokunyang:
## What do these changes do?

fix `mars.utils._get_ports_from_netstat` hang
## Related issue number
closes #2115

### DIFF
--- a/mars/utils.py
+++ b/mars/utils.py
@@ -204,27 +204,25 @@ def _get_ports_from_netstat():
     while True:
         p = subprocess.Popen('netstat -a -n -p tcp'.split(), stdout=subprocess.PIPE)
         try:
-            p.wait(5)
-            break
-        except subprocess.TimeoutExpired:
-            p.terminate()
-            continue
-    occupied = set()
-    for line in p.stdout:
-        line = to_str(line)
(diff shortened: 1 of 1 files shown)


qinxuye (approved):
LGTM

PR #2121, opened 2021-05-12
[ray] Support supervisor exclusive node option

fyrestone:
## What do these changes do?

- Problem
  - The supervisor use one node wastes computing resources.
- Solution
  - Provides an option to decide whether the supervisor should own the node or share the node with the worker.

## Related issue number
Closes https://github.com/mars-project/mars/issues/2122
Closes https://github.com/mars-project/mars/issues/2123
Closes https://github.com/mars-project/mars/issues/2124


mars/oscar/backends/ray/pool.py
```diff
@@ -36,13 +37,17 @@
 
 @_register_message_handler
 class RayMainActorPool(MainActorPoolBase):
+    @classmethod
+    def process_index_gen(cls, address):
```
chaokunyang, line 41: Invoking thie method multiple times with same address should return same generator

chaokunyang:
I'm not sure whether we should allow supervisor share same node with worker. The work load in worker may cause the ndoe oom and processes killed or the node lost.  Sharing node may make supervisor not stable. Systems such as spark/flink all make the master as a standalone node. Or can we make supervisor share node with the driver? Just like spark, it running spark driver in client process or a remote node.

chaokunyang:
In the long term, I think we can allow the supervisor running as a standalone node, or running in the client(ex: notebook kernel). If we running the supervisor in the client, there won't be any computing resources wasted.

chaokunyang:
If we create a placement group which used all cpu resources quata in a node, we can ensure there won't be other actors or placement goup bundles created in the same node. Thus we can ensure the worker will use an exclusive node, but still the supervisor can share node with other supervisor of a different job.

mars/oscar/backends/ray/pool.py
chaokunyang: Invoking thie method multiple times with same address doesn't return same generator, which is not consistent with `AbstractActorPool.process_index_gen`

fyrestone: The implementation of process_index_gen is different between Mars and Ray, it's not a problem.

fyrestone: > Invoking thie method multiple times with same address should return same generator

It returns a new generator by address not the same generator.

fyrestone:
> I'm not sure whether we should allow supervisor share same node with worker. The work load in worker may cause the ndoe oom and processes killed or the node lost. Sharing node will make the supervisor not stable. Systems such as spark/flink all make the master as a standalone node. Or can we make supervisor share node with the driver? Just like spark, it can run the spark driver in the client process or in a remote node.

Currently, sharing nodes is better than exclusive node.
- Ray only support same nodes with same resources.
- The supervisor does not cost too much resources.
- The failover is not ready.

Sharing node will make the supervisor more unstable, but only when the sharing node is down. It's a good idea to start the supervisor in driver, but the connection between supervisor and workers will be out of the cluster. I am not sure the ray client is stable enough for this.

fyrestone:
> If we create a placement group which used all cpu resources quata in a node, we can ensure there won't be other actors or placement goup bundles created in the same node. Thus we can ensure the worker will use an exclusive node, but still the supervisor can share node with other supervisor of a different job.

It works, but many supervisors may be scheduled to the same node. If the node is down, many jobs will be affected.

fyrestone:
> In the long term, I think we can allow the supervisor running as a standalone node, or running in the client(ex: notebook kernel). If we running the supervisor in the client, there won't be any computing resources wasted.

I agree with you, we can enable the supervisor running as a standalone node as needed.

mars/oscar/backends/ray/pool.py
chaokunyang: For the caller `AbstractActorPool`, the inferface behaviour should be consistent. We can return same generator for same address or make `AbstractActorPool.process_index_gen` return a new generator for every call

mars/deploy/oscar/ray.py
```diff
@@ -82,31 +89,46 @@ def __init__(self,
...
+                worker_address = process_placement_to_address(self._cluster_name, worker_index, 0)
+                worker_addresses.append(worker_address)
+                worker_node_address = node_placement_to_address(self._cluster_name, worker_index)
+                address_to_resources[worker_node_address] = {
+                    'CPU': self._worker_cpu,
+                    # 'memory': self._worker_mem
+                }
+        else:
+            for worker_index in range(self._worker_num):
+                worker_address = process_placement_to_address(self._cluster_name, worker_index, int(worker_index == 0))
```
chaokunyang, line 113: supervisor can have some sub pools. The worker address should start after the last supervisor sub pool address instead of `1`. For example, if the node parallelism is 4 and the supervisor pool address is `ray:://cluster/0/0, ray:://cluster/0/1`, the worker address should be `ray:://cluster/0/2, ray:://cluster/0/3`

chaokunyang: Can we add a test with supervisor parallelism of `2` to cover this case?

fyrestone: The `create_supervisor_actor_pool` is called with a fixed n_process=0. There is no sub pool of supervisor actor pool. @qinxuye Will the supervisor actor pool have a sub pool in the future?

qinxuye: Yes, for local it's 0, but for a distributed setting, it must be >= 0

fyrestone: > Yes, for local it's 0, but for a distributed setting, it must be >= 0

Thanks. @chaokunyang I will make the n_process configurable for ray.

qinxuye (approved):
LGTM

PR #2139, opened 2021-06-02
Configure azure pipelines job timeout

wjsi:
## What do these changes do?

Configure azure pipelines job timeout


qinxuye (approved):
LGTM

PR #2147, opened 2021-06-08
Fix race condition when creating actors with IdleLabel strategy in a parallel way

### LINKED ISSUE #2142: [BUG] Race condition when creating actors with parallel tasks
**Describe the bug**
When creating actors in a labeled pool using parallel asyncio tasks, allocation strategy fails because of possible race condition.

**To Reproduce**
```python
import asyncio
import os
from mars import oscar as mo
from mars.oscar.backends.allocate_strategy import IdleLabel

pool = await mo.create_actor_pool('127.0.0.1', n_process=2, labels=[None, 'numa-0', 'numa-0'])

class TestActor(mo.Actor):
    def get_pid(self):
        return os.getpid()

strategy = IdleLabel('numa-0', 'tests')
tasks = [
    mo.create_actor(TestActor, allocate_strategy=strategy, address=pool.external_address),
    mo.create_actor(TestActor, allocate_strategy=strategy, address=pool.external_address),
]
refs = await asyncio.gather(*tasks)
# outputs identical process ids, while the result should be different
print([await ref.get_pid() for ref in refs])
```

**Expected behavior**
``get_pid()`` calls return different process ids.

qinxuye:
## What do these changes do?

This PR fixes race condition when creating actors with IdleLabel strategy in a parallel way.

## Related issue number

Fixes #2142 .

### DIFF
--- a/mars/oscar/backends/mars/tests/test_pool.py
+++ b/mars/oscar/backends/mars/tests/test_pool.py
@@ -574,3 +574,26 @@ async def test_two_pools():
             Router.get_instance().get_internal_address(actor_ref4.address))
 
         assert await actor_ref2.add_other(actor_ref4, 3) == 13
+
+
+@pytest.mark.asyncio
+async def test_parallel_allocate_idle_label():
+    start_method = os.environ.get('POOL_START_METHOD', 'forkserver') \
+        if sys.platform != 'win32' else None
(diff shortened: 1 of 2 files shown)


wjsi (approved):
LGTM

hekaisheng (approved):
LGTM

PR #2161, opened 2021-06-18
Make sure slot usages updated when some task ends

wjsi:
## What do these changes do?

Make sure slot usages are updated when some task ends. This may reduce the ratio when heavy-loaded jobs are loaded onto the same worker.


qinxuye (approved):
LGTM

hekaisheng (approved):
LGTM

PR #2163, opened 2021-06-18
Fault injection subtask processor

fyrestone:
## What do these changes do?
In order to develop the failure recovery feature, we need a fault injection subtask processor.

This PR includes,
- A FaultInjectionSubtaskProcessor to trigger fault when executing an operand.

## Related issue number
N/A


qinxuye:
Actually, now we can replace the `execute` of an operand, see https://github.com/mars-project/mars/blob/57a54e5a88f5fee2530c621f5af96443f8c18ec0/mars/dataframe/indexing/tests/test_indexing_execution.py#L1006-L1008

wjsi:
I don't think now `fault_injection` forms a new service, as

1. `FaultInjectionSubtaskProcessor` is actually part of Subtask service, while services should only communicate with APIs, and
2. what the service is expected to act is just part of the role of worker part of Scheduling service. Just add some extra arg on Subtask may be a better solution.

btw, I wonder if adding a global `_fault_count` will help every subtask recover correctly.

fyrestone:
@qinxuye @wjsi I have changed this PR to create a remote object for the fault injection manager.

qinxuye (approved):
LGTM

wjsi (approved):
LGTM

PR #2170, opened 2021-06-22
[Ray] ray two way hang-detectable channel

chaokunyang:
## What do these changes do?
This PR add support for hang detect in ray communication. Main changes include:
* Add ray two-way channel for actor2actor communication
* Add hang detect for ray object get
* Increase async actor concurrency

By default, 1000 tasks can be running concurrently. So if we has actor calls more than 1000 on an actor and some call rely on next calls, there may be 1000 call waiting in the queue  eventually and next call can never got executed. Thus hang will happen. Two-way channel can solve this kind of hang, since it reduce unfinished calls. Increasing async actor concurrency also mitigate this.

It seems that ray 1.4.0 will [leak worker process](https://github.com/ray-project/ray/issues/16802) when using `async actor` and `exit_actor` simultaneously, this PR also workaround it by killing the processes forcibly.

### DIFF
--- a/.github/workflows/core-ci.yml
+++ b/.github/workflows/core-ci.yml
@@ -48,6 +48,11 @@ jobs:
           pip install numpy\<1.20.0
         fi
 
+        # todo remove this when #2191 is fixed.
+        if [[ "$PYTHON" =~ "3.7" ]]; then
+          pip install pandas\<1.3.0
+        fi
+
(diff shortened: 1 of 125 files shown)


wjsi (approved):
LGTM

qinxuye (approved):
LGTM

PR #2179, opened 2021-06-30
[WIP][ray] Try to fix ray CI

fyrestone:
## What do these changes do?

Ray CI fixture use function scope instead of module scope.


chaokunyang:
Fixed in #2170

PR #2198, opened 2021-07-07
Basic rerun subtask

fyrestone:
## What do these changes do?

- Support rerun subtask when the subtask is failed. (disabled by default)
- Handle `retry_when_fail` for the operands.
- Add an option to config, e.g.

```yaml
scheduling:
  subtask_max_retries: 2
```

- Add an AbstractFaultInjectionManager for test.

## Related issue number
N/A


mars/services/scheduling/worker/execution.py
```diff
@@ -48,16 +50,19 @@ class SubtaskExecutionInfo:
...
+    num_runs: int = 0
     slot_id: Optional[int] = None
     kill_timeout: Optional[int] = None
 
 
 class SubtaskExecutionActor(mo.Actor):
     _subtask_info: Dict[str, SubtaskExecutionInfo]
 
-    def __init__(self):
+    def __init__(self, default_config: Dict):
```
wjsi, line 62: We recommend adding arguments for actor constructors directly. Service config can be parsed and loaded to actors via start() methods.

mars/services/scheduling/worker/execution.py
```diff
@@ -211,34 +215,8 @@ async def internal_run_subtask(self, subtask: Subtask, band_name: str):
...
-            if subtask_info.cancelling:
-                raise asyncio.CancelledError
-
-            subtask_info.result.status = SubtaskStatus.running
-            run_aiotask = asyncio.create_task(
-                subtask_api.run_subtask_in_slot(band_name, slot_id, subtask))
-            yield asyncio.shield(run_aiotask)
-            if subtask_info.cancelling:
-                raise asyncio.CancelledError
+            subtask_info.result = yield self._retry_run_subtask(subtask, band_name, subtask_api)
```
wjsi, line 218: We need to make sure quota is available before every attempt, thus our tasks does not go OOM.

mars/services/scheduling/worker/execution.py
```diff
@@ -274,12 +247,93 @@ async def internal_run_subtask(self, subtask: Subtask, band_name: str):
                                                 subtask.session_id)
             yield task_api.set_subtask_result(subtask_info.result)
 
+    async def _retry_run_subtask(self, subtask: Subtask, band_name: str, subtask_api: SubtaskAPI):
```
wjsi, line 250: Need refactoring. Using nested `asynccontextmanager` is not clear. Handling exceptions with `issubclass` is also not Pythonic. A recommended way is to separate run and retry into two separate methods.

wjsi:
For some operands (esp. operands with user code) we have `retry_when_fail` argument which defines the behavior of operand after fail. Subtask retrial need to take that into consideration.

wjsi:
What's more, we may only retry when `ServerClosed` is raised now and add causes for retry later on. Retry for 99% of operand errors are useless.

mars/services/scheduling/worker/execution.py
fyrestone: The `default_config` is similar with the `WorkerQuotaManagerActor`: https://github.com/mars-project/mars/blob/master/mars/services/scheduling/worker/service.py#L47

mars/services/scheduling/worker/execution.py
fyrestone: Good suggestion. I will fix it.

mars/services/scheduling/worker/execution.py
fyrestone: Can I move the `asynccontextmanager` to the module level? I think it is much clearer than `try ... except ...`, and is much better than the async generator. Using an async generator in a method has different meanings than in the actor method. The latter one will release the actor lock.

mars/services/scheduling/worker/execution.py
fyrestone: > For some operands (esp. operands with user code) we have `retry_when_fail` argument which defines the behavior of operand after fail. Subtask retrial need to take that into consideration.

Thanks. I will take account into  the `retry_when_fail` argument.

fyrestone:
> What's more, we may only retry when `ServerClosed` is raised now and add causes for retry later on. Retry for 99% of operand errors is useless. The data prepare step is easier to fail but is not in the retry scope.

If we only retry when `ServerClosed` is raised, then the subtask exception will fail the task. For example, if the worker process exits, then we retrying the subtask, it may get an ConnectionError if the process fails again. I think we can retry all exceptions here to solve the problems. Then, we can optimize the logic after rerun for the Mars on ray is supported.

I will retry the `data prepare step` to make it more reliable.

wjsi:
> ConnectionError

Then add `OSError` ?

fyrestone:
> ConnectionError

OSError sounds good, it covers many retriable cases. Can we change the mars's `ServerClosed` error to subclass from the `OSError`?

wjsi:
> > ConnectionError
> 
> OSError sounds good, it covers many retriable cases. Can we change the mars's `ServerClosed` error to subclass from the `OSError`?

Just cover two errors as `ServerClosed` may caused by Mars Scheduling itself.

fyrestone:
> > > ConnectionError
> > 
> > 
> > OSError sounds good, it covers many retriable cases. Can we change the mars's `ServerClosed` error to subclass from the `OSError`?
> 
> Just cover two errors as `ServerClosed` may caused by Mars Scheduling itself (or we may cover MarsError?).

Let's retry the subtask when `OSError` and `MarsError` occurred. Can I add a base type `MarsError` for the errors in https://github.com/mars-project/mars/blob/master/mars/oscar/errors.py?

mars/services/scheduling/worker/execution.py
fyrestone: @wjsi Is the `retry_when_fail` implemented? I can't find the retry logic.

wjsi: > The `default_config` is similar with the `WorkerQuotaManagerActor`: https://github.com/mars-project/mars/blob/master/mars/services/scheduling/worker/service.py#L47

`default_config` for `WorkerQuotaManagerActor` is designed to create other actors and can be overwritten. For `SubtaskExecutionActor`, there is no defaults. Hence please use arguments directly.

fyrestone: > > The `default_config` is similar with the `WorkerQuotaManagerActor`: https://github.com/mars-project/mars/blob/master/mars/services/scheduling/worker/service.py#L47
> 
> `default_config` for `WorkerQuotaManagerActor` is designed to create other actors and can be overwritten. For `SubtaskExecutionActor`, there is no defaults. Hence please use arguments directly.

Got it. I will change the `default_config` to the arguments.

mars/services/scheduling/worker/execution.py
```diff
@@ -48,16 +51,73 @@ class SubtaskExecutionInfo:
     supervisor_address: str
     result: SubtaskResult = field(default_factory=SubtaskResult)
     cancelling: bool = False
+    max_retries: int = 0
+    num_retries: int = 0
     slot_id: Optional[int] = None
     kill_timeout: Optional[int] = None
 
 
+class SlotContext:
```
wjsi, line 60: Now that we have stateless actors, context managers may be avoided.

fyrestone: The stateless actor PR https://github.com/mars-project/mars/pull/2218 added a new API `wait_actor_pool_recovered` which is useful to wait the actor pool. But, I am not sure if it is related to the slot context.

fyrestone:
@qinxuye @wjsi The project coverage is very low, i am not sure what happened.

qinxuye:
> @qinxuye @wjsi The project test coverage is very low, i am not sure what happened.

Azure pipelines seem still broken.

fyrestone:
@qinxuye The `test_task_manager.py::test_error_task` often fails, can we mark it as flaky?

qinxuye:
> @qinxuye The `test_task_manager.py::test_error_task` often fails, can we mark it as flaky?

OK, I think it's reasonable.

mars/deploy/oscar/tests/test_fault_injection.py
```diff
@@ -13,55 +13,66 @@
...
-import uuid
 import pytest
 import numpy as np
 
 import mars.tensor as mt
+from mars.remote import spawn
 from mars.deploy.oscar.local import new_cluster
 from mars.deploy.oscar.session import get_default_async_session
+from mars.oscar.errors import FaultInjectionError, ServerClosed
+from ....services.tests.fault_injection_manager import FaultType, AbstractFaultInjectionManager, ExtraConfigKey
```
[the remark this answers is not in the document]
fyrestone: Fixed. Thanks.

mars/services/scheduling/worker/execution.py
```diff
@@ -25,8 +25,8 @@
 from .... import oscar as mo
 from ....core.graph import DAG
 from ....core.operand import Fetch
+from ....oscar.errors import MarsError
```
wjsi, line 28: Put this line in alphabetical order.

wjsi (commented):
LGTM over all, I left 2 comments.

mars/services/scheduling/worker/execution.py
```diff
@@ -22,11 +22,11 @@
 from dataclasses import dataclass, field
 from typing import Dict, Optional, Union
 
+import mars.oscar.errors
```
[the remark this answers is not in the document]
fyrestone: Fixed. Hope we can add some lint rules for the `relative / absolute` imports and the import order.

mars/services/scheduling/worker/execution.py
fyrestone: Fixed. Thanks.

mars/services/scheduling/worker/execution.py
```diff
@@ -250,36 +278,79 @@ async def internal_run_subtask(self, subtask: Subtask, band_name: str):
...
             if self._global_slot_ref is not None:
-                yield (
-                    # make sure slot is released before marking tasks as finished
-                    self._global_slot_ref.release_subtask_slots(
-                        (self.address, band_name), subtask.session_id, subtask.subtask_id),
-                    # make sure new slot usages are uploaded in time
-                    slot_manager_ref.upload_slot_usages(periodical=False)
-                )
+                # make sure slot is released before marking tasks as finished
+                await self._global_slot_ref.release_subtask_slots(
```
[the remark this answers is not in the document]
fyrestone: Fixed. Thanks.

fyrestone:
@wjsi Thanks for merging the latest master.

wjsi (approved):
LGTM

qinxuye (approved):
LGTM

PR #2201, opened 2021-07-07
Restart idle slots when hard memory limit is encountered

wjsi:
## What do these changes do?

Restart idle slots when hard memory limit is encountered. This will relieve task stuck when too much memory used by small allocations caused by `PyAlloc`.

### DIFF
--- a/mars/services/scheduling/worker/quota.py
+++ b/mars/services/scheduling/worker/quota.py
@@ -43,7 +43,7 @@ class QuotaActor(mo.Actor):
     def gen_uid(cls, band_name: str):
         return f'{band_name}_quota'
 
-    def __init__(self, band: BandType, quota_size: int):
+    def __init__(self, band: BandType, quota_size: int, **kw):
         super().__init__()
         self._requests = OrderedDict()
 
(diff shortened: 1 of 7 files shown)


mars/services/scheduling/worker/quota.py
```diff
@@ -269,29 +272,37 @@ def __init__(self, band: BandType, quota_size: int, hard_limit: int = None,
         self._refresh_time = refresh_time or 1
 
         self._stat_refresh_task = None
+        self._slot_manager_ref = None
 
     async def __post_create__(self):
         await super().__post_create__()
         self._stat_refresh_task = self.ref().update_mem_stats.tell_delay(delay=self._refresh_time)
 
+        from .workerslot import BandSlotManagerActor
```
hekaisheng, line 281: Move it to beginning of the function or file.

hekaisheng (commented):
LGTM overall

mars/services/scheduling/worker/quota.py
wjsi: Close to the user location, no need to move.

qinxuye (approved):
LGTM

hekaisheng (approved):
LGTM

PR #2203, opened 2021-07-08
Fix `NoDataToSpill` when multiple storage quota requests happen simultaneously

hekaisheng:
## What do these changes do?

We catch `NoDataToSpill` and create an asyncio event, when put or unpin finishes, we will check spillable size, if size is enough for spilling, call event.set() to wake up spilling task.

### DIFF
--- a/mars/services/storage/core.py
+++ b/mars/services/storage/core.py
@@ -112,17 +112,21 @@ class StorageQuotaActor(mo.Actor):
     def update_quota(self, size: int):
         if self._total_size is not None:
             self._used_size += size
+        logger.debug('Update %s bytes of %s, used size now is %s',
+                     size, self._level, self._used_size)
 
     def request_quota(self, size: int) -> bool:
-        logger.debug('Request %s bytes of %s, used size is %s,'
-                     'total size is %s', size, self.level, self.used_size, self.total_size)
         if self._total_size is not None and size > self._total_size:  # pragma: no cover
             raise RuntimeError(f'Request size {size} is larger '
                                f'than total size {self._total_size}')
         if self._total_size is None:
             self._used_size += size
+            logger.debug('Request %s bytes of %s, used size now is %s,'
+                         'total size is %s', size, self.level, self.used_size, self.total_size)
             return True
         elif self._used_size + size >= self._total_size:
+            logger.debug('Request %s bytes of %s, used size now is %s,'
+                         'space is not enough for the request', size, self.level, self.used_size)
             return False
         else:
             self._used_size += size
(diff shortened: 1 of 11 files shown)


mars/services/storage/core.py
```diff
@@ -244,13 +235,24 @@ def _put_data_info(self,
                 (session_id, data_key), data_info.store_size)
         if isinstance(data_key, tuple):
             self._main_key_to_sub_keys[(session_id, data_key[0])].update([data_key])
+        level = data_info.level
+        # when put finishes, check if spillable size is enough for spilling task
+        if self._spill_events[level] is not None:
```
qinxuye, line 240: Consider to wrap the logic into a SpillHandler or sth like that.

mars/services/storage/handler.py
```diff
@@ -0,0 +1,397 @@
...
+                    level: StorageLevel,
+                    request_size: int,
+                    object_size: int):
+        from .spill import spill
+
+        try:
+            await spill(request_size, level, self._data_manager_ref, self)
+        except NoDataToSpill:
+            logger.warning('No data to spill %s bytes, waiting more space', request_size)
+            await self._spill_manager_refs[level].create_spill_event(object_size)
```
qinxuye, line 387: Can just be one API `wait_for_space(object_size)`.

mars/services/storage/spill.py
```diff
@@ -72,36 +74,91 @@ def unpin_data(self, key):
         if self._pinned_keys[key] <= 0:
             del self._pinned_keys[key]
 
+    @property
+    def spillable_size(self):
```
qinxuye, line 78: Since this is a non-trivial operation, better to create a counterpart function instead of property.

mars/services/storage/spill.py
```diff
@@ -72,36 +74,91 @@ def unpin_data(self, key):
...
+    def gen_uid(cls, level: StorageLevel):
+        return f'spill_manager_{level}'
+
+    async def create_spill_event(self, size):
+        # make sure only one spilling task is waiting the event
+        yield self._lock.acquire()
+        self._event = event = asyncio.Event()
+        event.size = size
+
+    def notify_spill_event(self, spillable_size, quota_left):
```
qinxuye, line 136: Rename to `notify_space_can_spill` or something like that. BTW, type annotation should be added, so do the other funcitons.

qinxuye (commented):
LGTM overall

qinxuye (approved):
LGTM

wjsi (approved):
LGTM

PR #2213, opened 2021-07-14
Support choosing aggregation algorithm at runtime

### LINKED ISSUE #2215: Adaptive aggregation algorithm according to data pattern

For now, groupby+agg uses tree aggregation by default.

image

That is, apply groupby+agg to each chunk, then concatenate them, apply groupby+agg again, until one chunk left.

At the same time, shuffle-based algorithm is supported as well.

image

After groupby+agg on each chunk, shuffle data to different partitions, then groupby+agg on partitioned chunk.

Tree aggregation can potentially cause OOM if group keys are not so much, while shuffle-based algorithm is relatively slow due to barrier sync.

A more intelligent, decide-at-runtime algorithm can be added according to data pattern. groupby+agg can be applied to small part of data, and we would know the data pattern, and tree or shuffle can be decided at that time.

qinxuye:
## What do these changes do?

This PR added support for choosing aggregation algorithm at runtime.

## Related issue number

Resolves #2215 .

### DIFF
--- a/.github/workflows/os-compat-ci.yml
+++ b/.github/workflows/os-compat-ci.yml
@@ -37,6 +37,7 @@ jobs:
 
         source ./ci/rewrite-cov-config.sh
 
+        pip install git+https://github.com/mars-project/pytest-asyncio.git
         pip install numpy scipy cython
         pip install -e ".[dev,distributed]"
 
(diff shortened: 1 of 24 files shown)


wjsi (approved):
LGTM

hekaisheng (approved):
LGTM

PR #2248, opened 2021-07-26
Record mapper meta for shuffle task

qinxuye:
## What do these changes do?

Subtask that contains mapper operand will generate multiple data whose key is consist of main and sub keys, before, only main key is recorded in meta, this will cause potential data leak, what's more, when auto scaling triggered, the data will be not aware of, so data may be lost. This PR tries to record their meta.


wjsi (approved):
LGTM

hekaisheng (approved):
LGTM

PR #2246, opened 2021-07-25
Support PyTorch Dataset for oscar

yuanchongtt:
## What do these changes do?

I have done a demo of pytorch dataset API by inheriting torch.utils.data.Dataset. I also write some testcase in tests subfolder. 

## Related issue number#2224

I have try transfer a torch dataloader object to torch script but I have no idea how to do it. Now I am reading raySGD code to find a way to help mars to run torch or tf code more flexible.

PR #2251, opened 2021-07-27
Support data dependency for `run_script`

qinxuye:
## What do these changes do?

This PR supports data dependency for `run_script`. `data` argument can be specified, and it should be a dict whose key is the variable name and value is tileable object.

Other two modifications:

1. `spawn` mode is removed for `run_script`.
2. `setup_cluster` is added to start a local cluster for test.

e.g.

Driver.

```python
import mars.dataframe as md
import mars.remote as mr

s = md.Series([1, 2, 3])

mr.run_script('my_script.py', data={'my_series': s})
```

`my_script.py`:

```python
s = my_series  # `my_series` can be used directly.
print(s.fetch().sum())  # should be 6
```

This is related to #2246 .

### DIFF
--- a/mars/deploy/oscar/tests/test_ray.py
+++ b/mars/deploy/oscar/tests/test_ray.py
@@ -21,7 +21,6 @@ from mars.deploy.oscar.ray import new_cluster, _load_config
 from mars.deploy.oscar.session import get_default_session, new_session
 from mars.deploy.oscar.tests import test_local
 from mars.serialization.ray import register_ray_serializers
-from mars.tests.conftest import *  # noqa
 from mars.tests.core import require_ray
 from mars.utils import lazy_import
 from .modules.utils import ( # noqa: F401; pylint: disable=unused-variable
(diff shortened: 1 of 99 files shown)


qinxuye:
cc @yuanchongtt

wjsi (approved):
LGTM

hekaisheng (approved):
LGTM

hekaisheng:
@chaokunyang @fyrestone need your review as you are code owners of some change files.

fyrestone (approved):
LGTM

mars/deploy/oscar/tests/test_ray.py
```diff
@@ -21,7 +21,6 @@
 from mars.deploy.oscar.session import get_default_session, new_session
 from mars.deploy.oscar.tests import test_local
 from mars.serialization.ray import register_ray_serializers
-from mars.tests.conftest import *  # noqa
```
chaokunyang, line 24: Why remove this? We still need ray cluster fixture?

qinxuye: I add the fixture to mars/conftest.py，fixture does not need to be imported explicitly any more.

PR #2273, opened 2021-07-31
Allow turning off actor killing

### LINKED ISSUE #2272: Allow suspending slot killing
Sometimes we don't want to stop slots as they cannot be restored as before. An option is needed in scheduling service to keep those slots from stopping.

wjsi:
## What do these changes do?

Add an option in scheduling service to suspend pool killing on task cancel.

## Related issue number

Fixes #2272


qinxuye (approved):
LGTM

hekaisheng (approved):
LGTM

PR #2280, opened 2021-08-02
Add bind method to facilitate extracting batch args

wjsi:
## What do these changes do?

Make it simple to extract arguments for batch methods. For instance,

```python
    class TestClass:
        @extensible
        def method(self, a, b=10):
            pass

        @method.batch
        def method(self, args_list, kwargs_list):
            for args, kwargs in zip(args_list, kwargs_list):
                a, b = self.method.bind(*args, **kwargs)
```

when

```python
test_inst.method.batch(
    test_inst.method.delay(20),
    test_inst.method.delay(30, 5),
)
```

is called, `method.bind` can extract argument `a` and `b` correctly with one line.


hekaisheng (approved):
LGTM

qinxuye (approved):
LGTM

PR #2283, opened 2021-08-03
Reduce memory estimation for specific operands

wjsi:
## What do these changes do?

Reduce memory estimation for specific operands. Also reduce scheduling overhead.


qinxuye (approved):
LGTM

hekaisheng (approved):
LGTM

PR #2286, opened 2021-08-04
Fix hangs when worker main pool has failures.

fyrestone:
## What do these changes do?

- Improve fault injection, support fault position and fault context.
- New `patch_cls` and `patch_super` in mars.tests.core
- Fix hangs when worker main pool has failures, e.g.
  - Unhandled exceptions in SubtaskExecutionActor.
  - Worker main pool crashes.


mars/services/scheduling/worker/execution.py
```diff
@@ -32,6 +32,12 @@
 from ...meta import MetaAPI
 from ...storage import StorageAPI
 from ...subtask import Subtask, SubtaskAPI, SubtaskResult, SubtaskStatus
+from ...tests.fault_injection_manager import (
```
wjsi, line 35: Test modules shall never be used in other part of the code.

fyrestone: Is there better ways to inject fault to test the code?

wjsi: > Is there better ways to inject fault to test the code?

I think it is better to use `patch` module. Some work is needed to pass method patches to different processes.

fyrestone: > > Is there better ways to inject fault to test the code?
> 
> I think it is better to use `patch` module. Some work is needed to pass method patches to different processes.

Can you give more information about the `patch` module?

wjsi (approved):
LGTM

qinxuye (approved):
LGTM

PR #2289, opened 2021-08-04
Add experimental Dask-on-Mars support

loopyme:
## What do these changes do?

Add experimental Dask-on-Mars support, which make it possible to run Dask collections on Mars.


qinxuye:
We don't have the habit to use sort of `experimental` staff, cuz users anyway need to change their code. I recommend to rename `experimental` to `contrib`, like `contrib/dask`.

qinxuye:
What's more, you can create a separate CI test in platform-ci, refer to https://github.com/mars-project/mars/blob/master/.github/workflows/platform-ci.yml to add a new ci in which dask installed.

loopyme:
I am not familiar with CI tests, will this works?

qinxuye:
Seems dask test timeout. Could you please address the issue please?

loopyme:
Working on that

qinxuye:
Is this PR ready for a review? maybe we could mark it as some milestone now?

mars/contrib/dask/converter.py
```diff
@@ -0,0 +1,50 @@
...
+    dc: dask collection
+        Dask collection object to be converted.
+
+    Returns
+    -------
+    Object
+        Mars Object.
+    """
+    if not is_dask_collection(dc):
+        raise ValueError(f"'{type(dc).__name__}' object is not a valid dask collection")
```
qinxuye, line 36: How about raise a TypeError? besides, ut can be added to cover this line.

mars/contrib/dask/tests/test_dask.py
```diff
@@ -0,0 +1,140 @@
...
+def require_dask_installed(func):
+    return pytest.mark.skipif(dask is None, reason='dask not installed')(func)
+
+
+def require_mimesis_installed(func):
+    return pytest.mark.skipif(mimesis is None, reason='mimesis not installed')(func)
+
+
+@require_dask_installed
+def test_delayed():
```
qinxuye, line 47: How about just using fixture `setup_cluster`?

qinxuye (commented):
LGTM overall, coverage could be increased a bit.

hekaisheng:
Will `convert_dask_collection` method be exposed to users? It seems that the recommended way to execute Dask collections is `df.compute(scheduler=mars_scheduler)`, could you add docs in  user guide?

mars/contrib/dask/tests/test_dask.py
loopyme: I used `lazy_import` to keep it consistent with the other test scripts, is that ok?

loopyme:
> Will `convert_dask_collection` method be exposed to users? It seems that the recommended way to execute Dask collections is `df.compute(scheduler=mars_scheduler)`, could you add docs in user guide?

Exposing  `convert_dask_collection` method to user sounds better to me, since it may be more robust than `mars_scheduler`.  Besides, it allows users to modifying the task graph based on the original Dask collections using Remote API or other Mars features. If user only need to run Dask collections, `mars_scheduler` will be better choice.

I am working on doc

mars/contrib/dask/tests/test_dask.py
loopyme: I used [`lazy_import` way](https://github.com/mars-project/mars/pull/2289/commits/908f1cef0700b231e1e92068d7e2ccaa14778ad5) to make it consistent with the other test scripts, is that ok?

qinxuye: I mean the fixture: https://github.com/mars-project/mars/blob/322c187842a0ca99cea2a5f311d10e681969bd78/mars/conftest.py#L134

`setup_cluster` will be called only once across the entire CI, it will significantly save the time.

For the usage, just use it as the argument of a test function.

```python
def test_delayed(setup_cluster):
```

No import is required, because pytest knows how to handle it.

Hance, if `setup_cluster` is used, `setup_function` and `teardown_function` can be removed.

loopyme: Not sure if I get this, are you talking about [this way](https://github.com/mars-project/mars/pull/2289/commits/717b213d260c046feb77f96c604acf9a42b22e6f)?

qinxuye: Exactly.

qinxuye:
Spelling check failed: https://github.com/mars-project/mars/pull/2289/checks?check_run_id=3342458497

loopyme:
Sorry about the typo

qinxuye (approved):
LGTM

wjsi (approved):
LGTM

qinxuye:
Congrats, look forward to seeing more contributions from you.

PR #2325, opened 2021-08-11
Optimize mars supervisor scheduling

chaokunyang:
## What do these changes do?
Fix supervisor hang:
* Before this pr, if we have 1000 chunks need to `set_subtask_result` and each take 1 second, we'll need to wait 1000 seconds before we can schedule next subtask graph.
* Big sutask graph generation may take more than 3 minutes, this will block event loop 3 minutes. This PR moved the graph preparation into another thread to avoid event loop hang.


chaokunyang:
Remove chunks time:
image

chaokunyang:
generate subtask graph time:
image
image

chaokunyang:
The duration for graph generation by this pr:
image

qinxuye (approved):
LGTM

wjsi (approved):
LGTM

PR #2328, opened 2021-08-13
Improve wait_actor_pool_recovered

### LINKED ISSUE #2320: [BUG] wait_actor_pool_recovered can never return in some cases
When handling the message of wait_pool_recovered, it retrieves the event in `self._recover_events` and waits for it. After sub pool is recovered, the event is set. 

There exist some problems, for example when handling stop messages at
https://github.com/mars-project/mars/blob/00332d63cf1d469bbaba950e09aaf522cef0b93a/mars/oscar/backends/pool.py#L859-L870

If at the time `stop_sub_pool` returns, the monitor have done recovering, the waiting will last forever and it just hangs.

Also when handling wait_pool_recovered message, in some extreme cases, if the `await` returns after recovering have been done, the waiting never ends.
https://github.com/mars-project/mars/blob/00332d63cf1d469bbaba950e09aaf522cef0b93a/mars/oscar/backends/pool.py#L872-L875

We should fix it carefully and make it more reliable to avoid potential problems.

keyile:
## What do these changes do?

This PR enhances reliability of wait_actor_pool_recovered and avoid waiting forever. The changes are small and the ideas are below:

1. The message handler creates the event and waits for it, so it can always retrieve the event (will not get None).
2. The monitor pops and sets the recover event at every end of iterations, so the message handler will not wait forever.
3. The monitor checks the existence at the start of iteration, to ensure the event will not be set too early.

This PR also adds tests for auto-recover and server-closed for ray as mars pool does.

## Related issue number

Resolve #2320

### DIFF
--- a/mars/oscar/backends/pool.py
+++ b/mars/oscar/backends/pool.py
@@ -864,15 +864,11 @@ class MainActorPoolBase(ActorPoolBase):
                     self.sub_processes[message.address],
                     timeout=timeout,
                     force=force)
-                if self._auto_recover:
-                    self._recover_events[message.address] = asyncio.Event()
                 processor.result = ResultMessage(message.message_id, True,
                                                  protocol=message.protocol)
             elif message.control_message_type == ControlMessageType.wait_pool_recovered:
-                # check the aliveness of sub pool first, in case monitor task haven't found it.
-                if not await self.is_sub_pool_alive(self.sub_processes[message.address]):
-                    if self._auto_recover and message.address not in self._recover_events:
-                        self._recover_events[message.address] = asyncio.Event()
+                if self._auto_recover and message.address not in self._recover_events:
+                    self._recover_events[message.address] = asyncio.Event()
 
                 event = self._recover_events.get(message.address, None)
                 if event is not None:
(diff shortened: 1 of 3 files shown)


mars/oscar/backends/pool.py
```diff
@@ -864,15 +864,11 @@ async def handle_control_command(self,
...
-                    self._recover_events[message.address] = asyncio.Event()
                 processor.result = ResultMessage(message.message_id, True,
                                                  protocol=message.protocol)
             elif message.control_message_type == ControlMessageType.wait_pool_recovered:
-                # check the aliveness of sub pool first, in case monitor task haven't found it.
-                if not await self.is_sub_pool_alive(self.sub_processes[message.address]):
-                    if self._auto_recover and message.address not in self._recover_events:
-                        self._recover_events[message.address] = asyncio.Event()
+                if self._auto_recover and message.address not in self._recover_events:
+                    self._recover_events[message.address] = asyncio.Event()
```
wjsi, line 871: I think one single event object is enough. Recover event is created during pool initialization and replaced with a new event every time when a monitor loop finishes.

qinxuye (approved):
LGTM

wjsi (approved):
LGTM

PR #2339, opened 2021-08-14
Remove bokeh from package requirements

wjsi:
## What do these changes do?

Remove bokeh from package requirements. We may use [ApexCharts.js](https://github.com/apexcharts/apexcharts.js) to achieve similar visual effect.


qinxuye (approved):
LGTM

hekaisheng (approved):
LGTM

PR #2343, opened 2021-08-16
[Ray] Fix ray ci

keyile:
## What do these changes do?

1. Clear the instances when stopping services. It guarantees to create new instances in the subsequent test case.
2. Eliminate the different behavior of ray backend.
3. Parametrize test_load_third_party_modules for better setup and teardown.


qinxuye (approved):
LGTM

hekaisheng (approved):
LGTM

Issue #2342, opened 2021-08-16
Tileable Progress can be illustrated on Task Detail DAG


RandomY-2:
**Is your feature request related to a problem? Please describe.**
On the Task Detail DAG, we can paint the color of each tileable based on its progress. This will help us to oversee the status of the task. 

Right now, all the tileables on the DAG are filled with a yellow color:
![Picture1](https://user-images.githubusercontent.com/69881587/129529255-1dde9071-e7ee-490f-9583-8abd1c8423ca.png)

 but if we paint the color based on the percentage finished, the graph will be more informative. A tileable that has been executed can have a filled color, while a tileable that is 70% executed can have 70% of it filled with yellow and the other 30% remain white.  This gives us a more straightforward understanding of the progress of the task overall. 

**Describe the solution you'd like**
We can create another API endpoint that takes a task id and returns a list/dict that contains all tileables in the DAG and their percentage finished as a float between 0 and 1, where 1 means this tileable has been executed and 0 means this tileable has not been executed, and a float in between means how much of this tileable has been executed.

A sample JSON returned may look like this: 

`
{
    [
        { tileable_id_1: 1 },
        { tileable_id_2: 0.7 },
        ...
    ]
}
`

Then on the frontend, we can change the color for tileables based on the corresponded float value.

wjsi:
I'm now considering whether we need to make this API more informative. For instance, we also need to display status information on the DAG. Hence the result of API may like

```json
{
  "tileable1": {
    "progress": 0.7,
    "status": "running",
  },
  "tileable2": {
    "progress": 1.0,
    "status": "finished",
  },
  "tileable3": {
    "status": "cancelled",
  },
  "tileable3": {
    "status": "errored",
  },
  ...
}
```

PR #2357, opened 2021-08-19
Add API to retrieve progress and status of tileables

wjsi:
## What do these changes do?

Add API support for #2342. Subtasks are binded to tileables to get progresses. If chunks of a tileable is fused with chunks of another tileable, progresses are shared between different tileables.

## Related issue number

Deal with API part of #2342.

### DIFF
--- a/.github/workflows/core-ci.yml
+++ b/.github/workflows/core-ci.yml
@@ -63,20 +63,7 @@ jobs:
           fi
 
           conda install -n test --quiet --yes -c pkgs/main python=$PYTHON certifi
-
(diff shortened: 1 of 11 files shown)


qinxuye (approved):
LGTM

hekaisheng (approved):
LGTM

RandomY-2:
Sometimes this API doesn't return all tileables in the tileable graph. For instance, when executing these operations, the API only returns 11 tileables while the tileable graph has 12 tileables. 

141629447700_ pic_hd

Issue #1826, opened 2020-12-25
Support `Series.between`


wjsi:
Support `Series.between`

suchendra-h:
Hi, How can I help with this issue?

qinxuye:
> Hi, How can I help with this issue?

You're welcome.

gowrijsuria:
Hi, 

I have created a PR #2368 for this issue.

PR #2366, opened 2021-08-22
Added Equals for Series and Dataframe

gowrijsuria:


wjsi:
We keep `package-lock.json` in source control. The reason can be found at https://stackoverflow.com/a/44210813.

Mars DataFrame objects does not have `_mgr` attribute, thus this implementation may not work. If you think it works, please add tests and make them work.

gowrijsuria:
> We keep `package-lock.json` in source control. The reason can be found at https://stackoverflow.com/a/44210813.
> 
> Mars DataFrame objects does not have `_mgr` attribute, thus this implementation may not work. If you think it works, please add tests and make them work.

Could you suggest what I can use instead of _mgr attribute?

gowrijsuria:
Should I add the tests to the file "mars/dataframe/tests/test_core.py" ? And for checking whether the tests are running perfectly , am I supposed to run "pytest mars"?

mars/dataframe/core.py
```diff
@@ -1028,6 +1028,14 @@ def empty(self):
             raise ValueError('Tileable object must be executed first')
         return shape == (0,)
 
+    @property
+    def equals(self, other):
+        if not (isinstance(other, type(self)) or isinstance(self, type(other))):
+            return False
+        from .datasource.from_tensor import series_from_tensor
+        other = series_from_tensor(other)
+        return self._mgr.equals(other._mgr)
```
qinxuye, line 1037: There is no `_mgr` properties for Mars DataFrame, you need other way to implement this function.

mars/dataframe/core.py
```diff
@@ -1483,6 +1491,14 @@ def empty(self):
             raise ValueError('Tileable object must be executed first')
         return (0 in shape)
 
+    @property
+    def equals(self, other):
+        if not (isinstance(other, type(self)) or isinstance(self, type(other))):
+            return False
+        from .datasource.from_tensor import dataframe_from_tensor
+        other = dataframe_from_tensor(other)
+        return self._mgr.equals(other._mgr)
```
qinxuye, line 1500: Ditto.

qinxuye (commented):
You cannot just copy the implementation of pandas, cuz Mars has no block manager inside, you need to seek out other way to implement this, and I think maybe it's not as simple as we expected.

qinxuye:
> Should I add the tests to the file "mars/dataframe/tests/test_core.py" ? And for checking whether the tests are running perfectly , am I supposed to run "pytest mars"?

Yes, you can run `pytest mars/dataframe/tests/test_core.py` for this file, and if you are using IDE like pycharm, you can run specified test as you wish.

qinxuye:
To be honest, `equals` is not as simple as we expected, and for the purpose of solving a good first issue, I think https://github.com/mars-project/mars/issues/1826 may be more suitable.

PR #2385, opened 2021-08-24
Add array pad function

JuntaoMa:
## What do these changes do?
Add array pad function, see [here](https://numpy.org/doc/stable/reference/generated/numpy.pad.html) for details.

Pad function has several padding modes, and currently four of them can work properly, i.e. 'constant', 'edge', 'linear_ramp', 'empty'. Since I apply `np.pad` to each edge input chunks independently, those modes which requiring values from other chunks, cannot produce correct results. This needs to be fixed later.


hekaisheng:
Does this PR support all modes for pad function? If not, please raise `NotImplementedError` for unsupported modes and add tests for all supported modes. In addition, please remove the commented code in test file.

PR #2398, opened 2021-08-28
Add checks for file mode changes and absolute imports

wjsi:
## What do these changes do?

Add checks below:
1. File mode changes. This forces commiters to ignore fileMode in Windows.
2. Absolute imports in non-test source code.

Also add a `.gitattributes` to avoid CRLF line endings in Windows.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,18 @@
+* text=auto
+# enforce text on certain files
+*.py text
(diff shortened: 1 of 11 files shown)


qinxuye (approved):
LGTM

hekaisheng (approved):
LGTM

PR #2406, opened 2021-08-30
[Ray] Fix ray version

chaokunyang:
## What do these changes do?
Ray master version is not stable, which makes mars on ray test unstable too. The PR fix the version to a specified commit and will change it to ray>=1.7.0 when it got released.


qinxuye (approved):
LGTM

wjsi (approved):

Issue #2384, opened 2021-08-24
Tileable subtask structure visualization


RandomY-2:
**Is your feature request related to a problem? Please describe.**
When visualizing the tileable graph, we can also include a dag visualization for the subtask structure of each tileable, which can help us better control the progress of tileable. 

**Describe the solution you'd like**
We can create an API endpoint to return the subtask dag structure based on the tileable id. The json returned can be something similar to the json returned from the tileable graph endpoint. For example: 

```
{
     dependencies: [
          { from_subtask_id, to_subtask_id }, 
          ...
     ],

     subtasks: [
          { subtask_id, subtask_name, subtask_progress, subtask_status },
          ...
     ],
}
```

Then on the web, when user clicks on a tileable, the selected tileable's subtask graph can be displayed on the right side tileable detail panel, and the progress and color will be updated periodically just like the progress of the dag.

wjsi:
As number of subtasks may be large, it is better to display them as small circles without text.

wjsi:
Already done.

PR #2426, opened 2021-09-04
Support visualizing subtask graphs on Mars Web

RandomY-2:
## What do these changes do?

The task detail page now can also allow users to visualize the subtask dag for tileables:

Screen Shot 2021-09-04 at 9 39 44 AM

When users click on one of the tileables in the tileable graph, the subtask graph for the selected tileable will be displayed on the right panel. The graph will show the dag structure of the subtasks for the selected tileable and the dependencies to other subtasks that don't belong to the selected tileable. 

Since a dag chart is needed for both tileable graph and subtask graph, the dag creation logistics are extracted to a new React component.

The subtask progress and status part is also separated from the original subtask api to allow easier request and update of subtask progress.

## Related issue number
#2384


wjsi:
Two suggestions / questions:

1. Add input / output indicators if exist.
2. Have you ever tested the performance of rendering, for instance, the time to render 1000 nodes?

RandomY-2:
> Two suggestions / questions before code review:
> 
> 1. Add input / output indicators if exist.
> 2. Have you ever tested the performance of rendering, for instance, the time to render 1000 nodes?

I tested rendering 900 nodes, and it takes about 4 seconds to render them. Would this be an acceptable time?

I also realized another potential issue. If the dependencies are too much, since the subtask graph only is rendered on the bottom part of the right panel, the dependencies are too close to each other. I will see if there is a way to make the graph looks better.

Screen Shot 2021-09-04 at 10 52 30 AM


Screen Shot 2021-09-04 at 10 50 52 AM

qinxuye:
> > Two suggestions / questions before code review:
> > 
> > 1. Add input / output indicators if exist.
> > 2. Have you ever tested the performance of rendering, for instance, the time to render 1000 nodes?
> 
> Sorry what do you mean by input / output indicators?
> 
> I tested rendering 900 nodes, and it takes about 4 seconds to render them. Would this be an acceptable time?
> 
> I also realized another potential issue. If the dependencies are too much, since the subtask graph only is rendered on the bottom part of the right panel, the dependencies are too close to each other. I will see if there is a way to make the graph to look better.
> 
> Screen Shot 2021-09-04 at 10 52 30 AM
> 
> Screen Shot 2021-09-04 at 10 50 52 AM

Is it possible to add a mask or some hint to tell user that it's rendering, sometimes I entered the page and found nothing there, after a while, a graph showed up. I think maybe we can tell the user that it's rendering the graph.

wjsi:
> > Two suggestions / questions before code review:
> > 
> > 1. Add input / output indicators if exist.
> > 2. Have you ever tested the performance of rendering, for instance, the time to render 1000 nodes?
> 
> Sorry what do you mean by input / output indicators?
> 
> I tested rendering 900 nodes, and it takes about 4 seconds to render them. Would this be an acceptable time?
> 
> I also realized another potential issue. If the dependencies are too much, since the subtask graph only is rendered on the bottom part of the right panel, the dependencies are too close to each other. I will see if there is a way to make the graph to look better.

Currently inputs and outputs are missing in subtask dags. Can you add them?

image

200 nodes per second is a little slow. Is it possible to accelerate it?

RandomY-2:
OK, I will see what I can do.

RandomY-2:
The new DAGCanvasChart component uses a library that employs canvas based graph rendering on the frontend, which, unlike the previous dagre-d3, doesn't need to interact with the browser dom. So the original 900 nodes dag now only takes about 0.5 seconds to render. I still need some time to finalize all the code and rendering.

wjsi:
Excellent work! Several proposals:

1. Add tooltips to subtask nodes. For calculation nodes, their names can be displayed. For input / output nodes, names of nodes from / to these nodes can be displayed.
2. It seems that colors of running nodes are not rendered correctly and are not refreshing when progress updates. Please fix it.
image
3. When number of chunks are large (for instance, 2500), the browser get stuck. Is it possible to give alerts when number of chunks are too large?
4. As new NPM libraries are introduced, please add their licenses in LICENSE file in the root dir. Both cytoscape.js and cytoscape.js-dagre use MIT license. Also remember to add license files at the head of source files you created.
5. Some packed JS files are a little too large after `npm run bundle`. Please try split them further.
image

wjsi:
By talking about `tooltips`, I mean something like this, not a new panel under existing ones

img

Browser stuck is not handled either.

mars/services/task/tests/test_service.py
```diff
@@ -459,30 +457,93 @@ def f1(count: int):
...
+
+    task_id = await task_api.submit_tileable_graph(graph, fuse_enabled=True)
+
+    tileable_graph = await task_api.get_tileable_graph_as_json(task_id)
+    for tileable in tileable_graph.get('tileables'):
+        subtask_details = await task_api.get_tileable_subtask_detail(task_id, tileable.get('tileableId'))
+
+        subtask_ids = set()
+        for subtask_id, subtask_detail in subtask_details.items():
+            assert ((subtask_detail.get('status') >= 0 and subtask_detail.get('status') <= 4) or (subtask_detail.get('status') == -1 and subtask_detail.get('progress') == -1) or ((subtask_detail.get('status') == -2 and subtask_detail.get('progress') == -2)))
```
wjsi, line 498: This line is tooo loong. Try split this in multiple lines. It is also confusing using values of enumerations instead of values themselves, may use `SubtaskStatus.cancelled.value` instead.

wjsi (dismissed):

wjsi:
Please rebase your code and refactor given co mments.

wjsi (changes requested):

wjsi (changes requested):

wjsi (approved):
LGTM

qinxuye (approved):
LGTM

Issue #2195, opened 2021-07-06
Mutable tensor/dataframe on oscar


sighingnow:
**Is your feature request related to a problem? Please describe.**

We previous implemented "mutable tensor" in #415 (see also linked pull requests), by,

+ when creating a mutable tensor, create a corresponding actor on the scheduler node
+ intermediate writes happens on this `MutableTensorActor`
+ when sealing the mutable tensor, this actor split the tensor to a set of chunks and distribute them to workers, as an executed `Tensor`. 

The previous design has many limitation, as the tensor is stored in scheduler, it is not very scalable, especially the tensor is large and not sparse.

**Describe the solution you'd like**

In this task we are going to revisit the mutable tensor/mutable dataframe design, as,

(please feel free to correct me if there are any mistakes/misunderstandings)

1. add an API to client/session to create mutable tensor/dataframe with given shape (or given columns and indexes for dataframe).
2. starts a "MutableActor" for every mutable tensors/dataframes, which, will first create a set of chunks according to the shape, and distribute those tiled chunks to all workers, and, on each worker, create a actor to manage write/read requests on the chunks being placed on that worker.
3. maintain a `chunk -> actor worker` mapping as well.
3. when read/write to the mutable tensor/dataframe, first figure out which chunk to read/write, and forward the request to that worker.
4. when sealing the mutable tensor/dataframe, put those intermediate chunks on workers to the storage backends (and may be to some other storage endpoints), register metadata for those chunks, destroy the actors on workers, register the metadata for the tensor/dataframe itself, and finally destroy the manage actor.

**Additional context**

See also #415 for previous design and see related pull requests for details about previous implementation.

sighingnow:
We previous implemented "mutable tensor" in #415 (see also linked pull requests), by,

+ when creating a mutable tensor, create a corresponding actor on the scheduler node
+ intermediate writes happens on this `MutableTensorActor`
+ when sealing the mutable tensor, this actor split the tensor to a set of chunks and distribute them to workers, as an executed `Tensor`. 

The previous design has many limitation, as the tensor is stored in scheduler, it is not very scalable.

PR #2432, opened 2021-09-06
Support mutable tensor on oscar

Coco58323:
## What do these changes do?

1. add an API to the client/session to create a mutable tensor with given shape
2. starts a "MutableActor" for every mutable tensor, which, will first create a set of chunks according to the shape, and distribute those tiled chunks to all workers, and, on each worker, create an actor to manage to write/read requests on the chunks being placed on that worker.
3. maintain a chunk -> actor worker mapping.
4. when read/write to the mutable tensor, first figure out which chunk to read/write, and forward the request to that worker. Now it can read/write a single point of a tensor.

## Related issue number

#2195

### DIFF
--- a/mars/deploy/oscar/session.py
+++ b/mars/deploy/oscar/session.py
@@ -345,6 +345,16 @@ class AbstractAsyncSession(AbstractSession, metaclass=ABCMeta):
         name : str
         """
 
+    @abstractmethod
+    async def create_mutable_tensor(self,
+                                    shape: tuple,
+                                    dtype: str,
+                                    chunksize,
+                                    name: str = None):
+        """
+        Create a mutable tensor.
+        """
+    
     async def stop_server(self):
         """
         Stop server.
@@ -493,6 +503,16 @@ class AbstractSyncSession(AbstractSession, metaclass=ABCMeta):
         -------
         web_endpoint : str
             web endpoint
+        """
+
+    @abstractmethod
+    def create_mutable_tensor(self,
+                              shape: tuple,
+                              dtype: str,
+                              chunksize,
+                              name: str = None):
+        """
+        Create a mutable tensor.
         """
 
     def fetch_log(self,
@@ -901,7 +921,7 @@ class _IsolatedSession(AbstractAsyncSession):
         await self._session_api.delete_session(self._session_id)
         if self._asyncio_task_timeout_detector_task:  # pragma: no cover
             self._asyncio_task_timeout_detector_task.cancel()
-
+    
     async def create_remote_object(self,
                                    session_id: str,
                                    name: str,
@@ -919,6 +939,13 @@ class _IsolatedSession(AbstractAsyncSession):
                                     session_id: str,
                                     name: str):
         return await self._session_api.destroy_remote_object(session_id, name)
+
+    async def create_mutable_tensor(self,
+                                    shape: tuple,
+                                    dtype: str,
+                                    chunksize,
+                                    name: str = None):
+        return await self._session_api.create_mutable_tensor(self._session_id, shape, dtype, chunksize, name)
 
     async def stop_server(self):
         if self.client:
@@ -1088,6 +1115,15 @@ class AsyncSession(AbstractAsyncSession):
                                     name: str):
         pass  # pragma: no cover
 
+    @implements(AbstractAsyncSession.create_mutable_tensor)
+    @_delegate_to_isolated_session
+    async def create_mutable_tensor(self,
+                                    shape: tuple,
+                                    dtype: str,
+                                    chunksize,
+                                    name: str = None):
+        pass  # pragma: no cover
+    
     @implements(AbstractAsyncSession.get_web_endpoint)
     @_delegate_to_isolated_session
     async def get_web_endpoint(self) -> Optional[str]:
@@ -1274,6 +1310,14 @@ class SyncSession(AbstractSyncSession):
     @implements(AbstractSyncSession.get_cluster_versions)
     @_delegate_to_isolated_session
     def get_cluster_versions(self) -> List[str]:
+        pass  # pragma: no cover
+
+    @implements(AbstractSyncSession.create_mutable_tensor)
+    @_delegate_to_isolated_session
+    def create_mutable_tensor(self,
+                                    shape: tuple,
+                                    chunksize,
+                                    dtype: str):
         pass  # pragma: no cover
 
     def destroy(self):
--- a/mars/services/mutable/supervisor/service.py
+++ b/mars/services/mutable/supervisor/service.py
@@ -1,57 +1,10 @@
-# Copyright 1999-2021 Alibaba Group Holding Ltd.
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#      http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
 import asyncio
 import sys
 import itertools
 from typing import OrderedDict
 from mars.tensor.utils import split_indexes_into_chunks,decide_chunk_sizes
-import numpy as np
-from ... import oscar as mo
-
-
-class Chunk:
-    def __init__(self, shape,value=None) -> None:
-        self._shape = shape
-        self._tensor = np.ones(shape)
-
-    def write(self, index, value):
-        self._tensor[index] = value
-    
-    def read(self, index):
-        return self._tensor[index]
-
-
-class MutableTensorChunkActor(mo.Actor):
-    def __init__(self, chunklist: OrderedDict, value=None) -> None:
-        self.idx_chunk = OrderedDict()
-        for k,v in chunklist.items():
-            self.idx_chunk[k] = Chunk(v,value)
-
-    async def __post_create__(self):
-        pass
-
-    async def __on_receive__(self, message):
-        return await super().__on_receive__(message)
-
-    async def write(self, index,relatepos,value):
-        chunk:Chunk = self.idx_chunk[index]
-        chunk.write(tuple(relatepos),value)
-
-    async def read(self, index, relatepos):
-        chunk:Chunk = self.idx_chunk[index]
-        return chunk.read(tuple(relatepos))
+from .... import oscar as mo
+from ..worker.service import MutableTensorChunkActor
 
 class MutableTensorActor(mo.Actor):
     def __init__(self, shape: tuple, dtype: str, chunksize, name: str = None):
@@ -122,6 +75,7 @@ class MutableTensorActor(mo.Actor):
                 v = v.T
                 for nidx in v:
                     val = await chunk_actor.read(idx,nidx)
+                    print(idx,nidx,val)
                     ans.append(val)
         return ans
         
@@ -136,21 +90,3 @@ class MutableTensorActor(mo.Actor):
     @property
     def name(self):
         return self._name
-
-
-class MutableTensor:
-    def __init__(self,
-                 ref: mo.ActorRef,
-                 loop: asyncio.AbstractEventLoop):
-        self._ref = ref
-        self._loop = loop
-
-    def __getattr__(self, attr):
-        func = getattr(self._ref, attr)
-
-        def wrap(*args, **kwargs):
-            coro = func(*args, **kwargs)
-            fut = asyncio.run_coroutine_threadsafe(coro, loop=self._loop)
-            return fut.result()
-
-        return wrap
--- a/mars/services/mutable/worker/service.py
+++ b/mars/services/mutable/worker/service.py
@@ -1,0 +1,38 @@
+# Copyright 1999-2021 Alibaba Group Holding Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import asyncio
+from typing import OrderedDict
+from .... import oscar as mo
+from .core import Chunk
+
+class MutableTensorChunkActor(mo.Actor):
+    def __init__(self, chunklist: OrderedDict, value=None) -> None:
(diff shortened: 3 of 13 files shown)


qinxuye:
Just merge the master branch.

mars/services/mutable/supervisor/service.py
```diff
@@ -0,0 +1,153 @@
...
+        for k in self._work_pools.items():
+            worker_address.append(str(k[0][0]))
+        chunknumber = 1; num = 0
+        for nsplit in self._nsplits:
+            chunknumber *= len(nsplit)
+        leftchunk = chunknumber
+        chunk_list = OrderedDict()
+        for idx in itertools.product(*(range(len(nsplit)) for nsplit in self._nsplits)):
+            chunk_list[idx] = [self._nsplits[i][idx[i]] for i in range(len(idx))]
+            num += 1; leftchunk -= 1
```
qinxuye, line 53: Split into multiple lines of code.

mars/services/mutable/supervisor/service.py
```diff
@@ -0,0 +1,153 @@
...
+            if (num == chunknumber//workernumer and  leftworker != 1  or leftworker == 1 and leftchunk == 0):
+                chunk_ref = await mo.create_actor(MutableTensorChunkActor, chunk_list, self.default_value, address=worker_address[leftworker-1])
+                num = 0; chunk_list = OrderedDict(); leftworker -= 1
```
qinxuye, line 56: Ditto.

mars/services/mutable/supervisor/service.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def calc_index(self, idx):
+        pos = 0; acc = 1
+        for it, nsplit in zip(itertools.count(0), reversed(self._nsplits)):
+            it = len(idx) - it-1
+            pos += acc*(idx[it])
+            acc *= len(nsplit)
+        return pos
+
+    async def write(self, index, value):
```
qinxuye, line 69: Writing and reading data should be a behavior of worker, it will be too slow to write data to supervisor and then workers.

mars/services/mutable/supervisor/service.py
```diff
@@ -0,0 +1,153 @@
...
+                for actor_index, lastindex in zip(itertools.count(0), self._chunkactors_lastindex):
+                    if lastindex >= pos:
+                        target_index = actor_index
+                        break
+                chunk_actor = self._chunk_to_actors[target_index]
+                v = v.T
+                for nidx in v:
+                    await chunk_actor.write(idx, nidx, value)
+
+    async def read(self, index):
```
qinxuye, line 84: Ditto.

mars/services/mutable/supervisor/service.py
```diff
@@ -0,0 +1,153 @@
...
+            chunk_list[idx] = [self._nsplits[i][idx[i]] for i in range(len(idx))]
+            num += 1; leftchunk -= 1
+            if (num == chunknumber//workernumer and  leftworker != 1  or leftworker == 1 and leftchunk == 0):
+                chunk_ref = await mo.create_actor(MutableTensorChunkActor, chunk_list, self.default_value, address=worker_address[leftworker-1])
+                num = 0; chunk_list = OrderedDict(); leftworker -= 1
+                self._chunk_to_actors.append(chunk_ref)
+                pos = self.calc_index(idx)
+                self._chunkactors_lastindex.append(pos)
+
+    def calc_index(self, idx):
```
qinxuye, line 61: Type annotation.

mars/services/mutable/supervisor/service.py
```diff
@@ -0,0 +1,153 @@
...
+                chunk_actor = self._chunk_to_actors[target_index]
+                v = v.T
+                for nidx in v:
+                    val = await chunk_actor.read(idx, nidx)
+                    print(idx, nidx, val)
+                    ans_list.append(val)
+        return ans_list
+
+
+class MutableTensor:
```
qinxuye, line 104: Add this class to APIs or somewhere else.

mars/services/mutable/worker/service.py
```diff
@@ -0,0 +1,38 @@
...
+            self.idx_chunk[k] = Chunk(v, default_value)
+
+    async def __post_create__(self):
+        pass
+
+    async def __on_receive__(self, message):
+        return await super().__on_receive__(message)
+
+    async def write(self, index, relatepos, value):
+        chunk: Chunk = self.idx_chunk[index]
```
qinxuye, line 33: Is it possible to record index to data inside with a version, so that the user can read some specified version data. The real data can be re-assembled in read.

mars/services/session/api/oscar.py
```diff
@@ -109,6 +109,20 @@ async def destroy_remote_object(self,
         session = await self._get_session_ref(session_id)
         return await session.destroy_remote_object(name)
 
+    async def create_mutable_tensor(self,
```
qinxuye, line 112: I recommend to add this API to mutable API instead of session API.

mars/services/session/supervisor/core.py
```diff
@@ -166,6 +173,22 @@ async def get_remote_object(self, name: str):
...
+            MutableTensorActor, shape, dtype, chunk_size, worker_pools, name, default_value, address=self.address, uid=to_binary(name))
+        wrapper = MutableTensor(ref)
+        self.mtensor_dict[name] = wrapper
+        return wrapper
+
+    async def get_mutable_tensor(self, name: str):
+        if name in self.mtensor_dict.keys():
+            return self.mtensor_dict[name]
+        else:
+            raise Exception('invalid name!')
```
qinxuye, line 190: Raise a ValueError instead a Exception.

qinxuye (commented):
I left some comments

qinxuye:
Besides, now, the data stored in worker is just a numpy ndarray, please use storage API to store data. And `seal_mutable_data` can be added in order to convert a mutable tensor with a given version into a vanilla mars tensor.

mars/services/session/supervisor/core.py
```diff
@@ -166,6 +173,22 @@ async def get_remote_object(self, name: str):
     async def destroy_remote_object(self, name: str):
         return await mo.destroy_actor(mo.ActorRef(self.address, to_binary(name)))
 
+    async def create_mutable_tensor(self, shape: tuple, dtype: str, chunk_size, name: str = None, default_value=0):
+        worker_pools: dict = await self._cluster_api.get_all_bands()
```
sighingnow, line 177: Use `role=NodeRole.WORKER`.

sighingnow:
@Coco58323 please address comments above.

About the "seal" operation, I would like to implement in this way,

1. put chunks to storage backend on every worker, returns a set of chunks in form `(chunk index, shape, dtype, data info, worker address)`, the data info is the result of putting chunks to the backend
2. have a special op, taking these data infos as arguments, and returns the data infos directly in its `execute()` method. The op would be tiled respected to `worker address` of each chunk
3. in `SubtaskProcessor._store_data`, if the chunk result is a `DataInfo` instance, just skip the "put" operation.

I think it is more concise and easy to implement than operating on metas of chunks and graphs than before.

Any comments? @qinxuye

qinxuye:
> @Coco58323 please address comments above.
> 
> About the "seal" operation, I would like to implement in this way,
> 
> 1. put chunks to storage backend on every worker, returns a set of chunks in form `(chunk index, shape, dtype, data info, worker address)`, the data info is the result of putting chunks to the backend
> 2. have a special op, taking these data infos as arguments, and returns the data infos directly in its `execute()` method. The op would be tiled respected to `worker address` of each chunk
> 3. in `SubtaskProcessor._store_data`, if the chunk result is a `DataInfo` instance, just skip the "put" operation.
> 
> I think it is more concise and easy to implement than operating on metas of chunks and graphs than before.
> 
> Any comments? @qinxuye

Actually, I think fetch op is enough, but be aware that the sealed data's meta has been recorded.

sighingnow:
> 
> Actually, I think fetch op is enough, but be aware that the sealed data's meta has been recorded.

Our previous implementation is returning a `TensorFetch` tensor when sealing, but that is quite tricky. As we need to add the tensor to existing graphs. And there was a `SealerActor` to register the metadata of chunks as well.

I think add a new op and a speical case in `SubtaskProcessor._store_data` would make things more cleaner, as we don't need to touch the core of scheduling and metadata management code to archieve the goal, just a new op and several lines of modifcation in the core.

sighingnow:
For more about previous complex logic, see also

1. https://github.com/sighingnow/mars/blob/c7897fbd144d230fff5edabc1494fb3ff44aa0d2/mars/scheduler/session.py#L130
2. https://github.com/sighingnow/mars/blob/master/mars/worker/seal.py#L49

qinxuye:
> Our previous implementation is returning a `TensorFetch` tensor when sealing, but that is quite tricky. As we need to add the tensor to existing graphs. And there was a `SealerActor` to register the metadata of chunks as well.

For oscar, graph is not required, actually, if meta is registered correctly, it should work.

sighingnow:
> > Our previous implementation is returning a `TensorFetch` tensor when sealing, but that is quite tricky. As we need to add the tensor to existing graphs. And there was a `SealerActor` to register the metadata of chunks as well.
> 
> For oscar, graph is not required, actually, if meta is registered correctly, it should work.

Do we not record the tileable graph any more? And do we still need to put keys of those chunks into the chunk graph?

qinxuye:
> 
> Do we not record the tileable graph any more? And do we still need to put keys of those chunks into the chunk graph?

Tileable graph is not required, tieable with fetch chunks is OK.

mars/deploy/oscar/session.py
```diff
@@ -16,6 +16,7 @@
 import concurrent.futures
 import itertools
 import logging
+from mars.services.mutable.api.oscar import MutableAPI
```
sighingnow, line 19: Put the import to `import ..`  group.

wjsi: Please follow instructions given by [flake8](https://dev.azure.com/mars-project/mars/_build/results?buildId=1548&view=logs&jobId=435ca956-9126-505f-566f-fff31072e2ba&j=435ca956-9126-505f-566f-fff31072e2ba&t=456eced4-5f9f-56bd-1d85-873f530e72f3) to fix your code.

mars/services/session/api/web.py
```diff
@@ -1,3 +1,4 @@
+
```
wjsi, line 1: Why add this empty line?

mars/services/session/supervisor/core.py
```diff
@@ -124,12 +128,16 @@ def __init__(self, session_id: str,
...
 
+        self.mtensor_dict = OrderedDict()
+
     @classmethod
     def gen_uid(cls, session_id):
         return f'{session_id}_session_actor'
 
     async def __post_create__(self):
         from .custom_log import CustomLogMetaActor
+        print('before clustapi', self.address)
```
wjsi, line 139: Remove prints added for debug purpose.

mars/services/mutable/supervisor/service.py
```diff
@@ -0,0 +1,103 @@
...
+        num = 0
+        for nsplit in self._nsplits:
+            chunknumber *= len(nsplit)
+        leftchunk = chunknumber
+        chunk_list = OrderedDict()
+        for idx in itertools.product(*(range(len(nsplit)) for nsplit in self._nsplits)):
+            chunk_list[idx] = [self._nsplits[i][idx[i]] for i in range(len(idx))]
+            num += 1
+            leftchunk -= 1
+            if (num == chunknumber//workernumer and  leftworker != 1  or leftworker == 1 and leftchunk == 0):
```
wjsi, line 54: Leave exactly one space before and after the operator.

mars/services/mutable/supervisor/service.py
```diff
@@ -0,0 +1,103 @@
...
+        worker_address = []
+        for k in self._work_pools.items():
+            worker_address.append(str(k[0][0]))
+        chunknumber = 1
+        num = 0
+        for nsplit in self._nsplits:
+            chunknumber *= len(nsplit)
+        leftchunk = chunknumber
+        chunk_list = OrderedDict()
+        for idx in itertools.product(*(range(len(nsplit)) for nsplit in self._nsplits)):
```
wjsi, line 50: Ditto.

mars/deploy/oscar/session.py
```diff
@@ -41,6 +41,7 @@
 from ...services.lifecycle import AbstractLifecycleAPI, LifecycleAPI
 from ...services.meta import MetaAPI, AbstractMetaAPI
 from ...services.session import AbstractSessionAPI, SessionAPI
+from ...services.mutable.api.oscar import MutableAPI
```
qinxuye, line 44: You can import the `MutableAPI` for the `services.mutable` module, just like the other lines.

mars/deploy/oscar/session.py
```diff
@@ -1030,6 +1074,7 @@ async def _init(cls,
         from ...services.meta import WebMetaAPI
         from ...services.task import WebTaskAPI
         from ...services.cluster import WebClusterAPI
+        from ...services.mutable.api.web import WebMutableAPI
```
qinxuye, line 1077: Ditto.

mars/services/mutable/api/oscar.py
```diff
@@ -0,0 +1,53 @@
...
+            address, SessionManagerActor.default_uid())
+        return MutableAPI(address, session_manager)
+
+    @alru_cache(cache_exceptions=False)
+    async def _get_session_ref(self, session_id: str) -> Union[SessionActor, mo.ActorRef]:
+        return await self._session_manager_ref.get_session_ref(session_id)
+
+    async def create_mutable_tensor(self, session_id: str, shape: tuple, dtype: str, chunk_size, name: str=None, default_value=0):
+        session = await self._get_session_ref(session_id)
+        return await session.create_mutable_tensor(shape, dtype, chunk_size, name, default_value)
```
qinxuye, line 49: Is it possible to implement this function without touching session manager?

mars/services/mutable/api/web.py
```diff
@@ -0,0 +1,28 @@
...
+
+#TODO: implements web
+
+
+class WebMutableAPI(AbstractMutableAPI):
+    def __init__(self, address: str):
+        self._address = address.rstrip('/')
+
+    async def create_mutable_tensor(self, session_id: str, shape: tuple, dtype: str, chunk_size, name: str = None, default_value=0):
+        pass
```
qinxuye, line 25: The web API is absent?

mars/services/mutable/supervisor/core.py
```diff
@@ -0,0 +1,121 @@
...
+                target_index = 0
+                pos = self.calc_index(idx)
+                for actor_index, lastindex in zip(itertools.count(0), self._chunkactors_lastindex):
+                    if lastindex >= pos:
+                        target_index = actor_index
+                        break
+                chunk_actor = self._chunk_to_actors[target_index]
+                v = v.T
+                for nidx in v:
+                    val = await chunk_actor.read(idx, nidx, time.time())
```
qinxuye, line 87: It could be really slow if we read one by one, is it possible to divide all indexes into actors, and fetch them together to accelerate?

qinxuye (commented):
I left some comments

qinxuye:
Thanks, look forward to seeing more contributions from you.

PR #2439, opened 2021-09-08
Fix missing DAGs when initializing with empty API results

RandomY-2:
## What do these changes do?

The current tileable graph compares prevProp's tileables and dependencies with the current tileables and dependencies, and only if they are different, the dag will be rendered. However, there is also a check that states that the length of tileable details must match the length of tileables, and if the React setState for tileable detail takes some time and finishes setting the state after the tileables and dependencies have been passed to the graph component, the prevProp's tileables and dependencies have already become the new tileables and dependencies. As such, the dag will never be rendered. 

This PR removed the length check for tileable details. Instead, if the rendered tileable's detail is not in the passed nodesStatus prop, the tileable will just be rendered as a pending tileable. So now when the tileables and dependencies change, the dag will always be rendered. Then when the detail prop changes, the new status and progress will be changed accordingly.

### DIFF
--- a/mars/services/web/ui/src/task_info/TaskTileableGraph.js
+++ b/mars/services/web/ui/src/task_info/TaskTileableGraph.js
@@ -17,14 +17,10 @@
 import React from 'react';
 import { select as d3Select } from 'd3-selection';
 import {
-    zoom as d3Zoom,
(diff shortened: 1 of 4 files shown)


wjsi (approved):
LGTM

PR #2452, opened 2021-09-14
Allow specifying multiple supervisor processes

### LINKED ISSUE #2451: [BUG] Mars web got stuck when chunk graph is large
**Describe the bug**
After a tileable graph with small chunk size is submitted and after a while, Mars Web got stuck.

**To Reproduce**
Use code below to reproduce:

```python
import mars
import mars.tensor as mt
sess = mars.new_session('http://localhost:13042/')
a = mt.ones((1000, 1000), chunk_size=10)
b = (a + 1) * 2 + (a - 3) ** 2
b.execute()
```

wjsi:
As loops of supervisors may be blocked by GIL, it is possible to use another process to handle sessions in separate processes. As it may bring serialization overhead, we offer it as an option.

Fixes #2451

(though as a partial solution)

### DIFF
--- a/mars/oscar/backends/pool.py
+++ b/mars/oscar/backends/pool.py
@@ -450,7 +450,7 @@ class ActorPoolBase(AbstractActorPool, metaclass=ABCMeta):
     @implements(AbstractActorPool.create_actor)
     async def create_actor(self, message: CreateActorMessage) -> result_message_type:
         with _ErrorProcessor(message.message_id, message.protocol) as processor:
-            actor_id = to_binary(message.actor_id)
+            actor_id = message.actor_id
             if actor_id in self._actors:
                 raise ActorAlreadyExist(
                     f"Actor {actor_id} already exist, " f"cannot create"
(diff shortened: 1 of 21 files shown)


mars/services/session/supervisor/core.py
```diff
@@ -42,14 +42,24 @@ async def create_session(self, session_id: str, create_services: bool = True):
...
+        try:
+            session_actor_ref = await mo.create_actor(
+                SessionActor,
+                session_id,
+                self._service_config,
+                address=address,
+                uid=SessionActor.gen_uid(session_id),
+                allocate_strategy=mo.allocate_strategy.RandomSubPool(),
+            )
+        except IndexError:
```
hekaisheng, line 54: Add comment about when `IndexError` happens.

mars/oscar/backends/pool.py
```diff
@@ -450,7 +450,7 @@ class ActorPoolBase(AbstractActorPool, metaclass=ABCMeta):
     @implements(AbstractActorPool.create_actor)
     async def create_actor(self, message: CreateActorMessage) -> result_message_type:
         with _ErrorProcessor(message.message_id, message.protocol) as processor:
-            actor_id = message.actor_id
+            actor_id = to_binary(message.actor_id)
```
hekaisheng, line 453: when would actor_id not be binary?

hekaisheng (approved):
LGTM

qinxuye (approved):
LGTM

PR #2466, opened 2021-09-17
Implements `glm.LogisticRegression`

Fernadoo:
## What do these changes do?

Implements logistic regression as well as SGD optimizer.


Fernadoo:
Could you guys take a look at the implementation? The algorithm is mathematically correct, but I am not sure about the style, like the interfaces.

If it is ok with you, I am gonna add pytests. Also, just let me know if you got some suggestions for testing.😁

PR #2470, opened 2021-09-18
Fixed a align related bug.

qxzhou1010:
Before fixed, if we run this test case, the error will occur.
```
def test_dataframe(setup):
    t = np.random.rand(10, 3)
    pdf = pd.DataFrame(t)

    df = md.DataFrame(pdf, chunk_size=(5, 3))
    r = df[0] != df.sort_index()[0].shift(-1)
    expected = pdf[0] != pdf.sort_index()[0].shift(-1)
    pd.testing.assert_series_equal(r.execute().fetch(), expected)
```
Run this test case, an error will occur.
`TypeError: Cannot interpret '0' as a data type`
Only need to change this code:
` if kw.get('name', None) is None and getattr(inputs[0], 'name', None) is not None:
                kw['name'] = inputs[0].name`
    in align.py,we can fix this bug


wjsi:
@qxzhou1010 Please do not open multiple PRs for one single issue. Pushing to https://github.com/qxzhou1010/mars/tree/fix-bug will automatically update your code in PR.

qxzhou1010:
A repeated pull request.

PR #2480, opened 2021-09-21
Split tileable information and subtask graph into two tabs

RandomY-2:
## What do these changes do?

The subtask graph and tileable details are now separated into two tabs, so there are more space to include information for both.

Screen Shot 2021-09-21 at 1 56 34 PM

Screen Shot 2021-09-21 at 1 57 32 PM


qinxuye (approved):
LGTM

PR #2481, opened 2021-09-22
Support specified vineyard socket and skip the launching vineyardd process

sighingnow:
## What do these changes do?

The vineyardd might be prelaunched in a `DaemonSet` and shared with other compute engines.

## Related issue number

N/A


sighingnow:
The failure is not caused by this pull request.

The patch coverage seems incorrect, as lines in the newly-added "else" branch are all covered: https://app.codecov.io/gh/mars-project/mars/compare/2481/tree/mars/storage/vineyard.py#D1L130

qinxuye (approved):
LGTM

PR #2493, opened 2021-09-30
Include tileable property in detail api

RandomY-2:
## What do these changes do?

Tileable properties are included in the detail api, and we can later display these properties on web.

### DIFF
--- a/mars/services/task/supervisor/processor.py
+++ b/mars/services/task/supervisor/processor.py
@@ -599,11 +599,26 @@ class TaskProcessorActor(mo.Actor):
             else:
                 status = SubtaskStatus.running
 
+            # get properties for tileable
+            all_propertes = {'slot': getattr(tileable, slot, None) for slot in tileable.__slots__}
+
+            # Only int/float/string type values for now
+            displayed_properties = dict()
+            for property_key, property_value in all_propertes['slot'].items():
+                # Omit key property
+                if (property_key == 'key'):
+                    continue
+
+                if isinstance(property_value, int) or isinstance(property_value, float) or isinstance(property_value, str):
+                    displayed_properties[property_key] = property_value
+
             tileable_infos[tileable.key] = {
                 'progress': progress,
                 'subtaskCount': len(results),
                 'status': status.value,
+                'properties': displayed_properties,
             }
+
         return tileable_infos
 
     def get_tileable_subtasks(self, tileable_id: str, with_input_output: bool):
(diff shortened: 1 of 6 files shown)


mars/services/task/supervisor/processor.py
```diff
@@ -599,11 +599,27 @@ def get_tileable_details(self):
             else:
                 status = SubtaskStatus.running
 
+            # get properties for tileable
+            all_propertes = {'slot': getattr(tileable, slot, None) for slot in tileable.__slots__}
```
wjsi, line 603: I wonder if this list comprehension expression will work. Please also try adding tests for your code.

mars/services/task/supervisor/processor.py
```diff
@@ -599,11 +599,29 @@ def get_tileable_details(self):
...
+            # Only int/float/string type values for now
+            displayed_properties = dict()
+            for property_key, property_value in all_propertes.items():
+                # Omit key property
+                if (property_key == 'key'):
+                    continue
+
+                if isinstance(property_value, int) or isinstance(property_value, float) \
+                    or isinstance(property_value, str):
+                    displayed_properties[property_key] = property_value
```
wjsi, line 616: Too complicated. two lines will cover all the needs.

```python
props = {slot: getattr(tileable, slot, None) for slot in tileable.__slots__}
props = {k: v for k, v in props.items() if k != 'key' and isinstance(v, (int, float, str)}
```

mars/services/task/tests/test_service.py
```diff
@@ -384,6 +384,41 @@ def _get_fields(details, field, wrapper=None):
     details = await task_api.get_tileable_details(task_id)
     assert details[r7.key]['status'] == SubtaskStatus.errored.value
 
+    def a():
```
wjsi, line 387: Is it possible to test with test cases generated previously?

mars/services/task/tests/test_service.py
```diff
@@ -384,6 +384,20 @@ def _get_fields(details, field, wrapper=None):
     details = await task_api.get_tileable_details(task_id)
     assert details[r7.key]['status'] == SubtaskStatus.errored.value
 
+    def is_valid_type(value):
+        return isinstance(value, int) or isinstance(value, float) or isinstance(value, str)
```
wjsi, line 388: Multiple `isinstance`s can be simplified with a single call on tuple of types, for inst., `isinstance(value, (int, float, str))`.

Issue #2112, opened 2021-04-27
Mars on ray


chaokunyang:
This issue is used to track the progress of mars on ray. All mars-on -ray related issue should linked with this issue.

chaokunyang:
#2972

PR #2576, opened 2021-11-06
[scheduling] A speculative execution based task scheduler

### LINKED ISSUE #2575: [Proposal] A speculative execution based task scheduler
For big cluster and distributed system, subtask may got hang or stale for many reasons on different nodes. In those cases, the whole subtask graph may never got finsihed or got pretty slowly by some subtasks.
Task speculative execution mode is a way we can inspect part of homogeneous subtaks, and based on the stats of those subtasks, to determine whether a subtask got stale/slow/hang, then we can speculatively schedule it to other workers at the same time. The subtask execution finished earlier will be used as the result for this subtask and its result chunks.

Main part:
* subtask unique and determinate id for its compute logic
* group subtask by compute logic so that we can base on finished homogeneous subtak to determine whether a running subtask got stale/slow/hang.
* concurent and duplicate run and isolation for subtask at the same time.
* Slow/stale node detection and remove
* Support submit subtask by excluding some workers
* Speculative execution for subtasks.

# Related issues
#2112

chaokunyang:
## What do these changes do?
This pr implements a speculative task execution based scheduler. For big cluster and distributed system, subtask may got hang or stale for many reasons on different nodes. In those cases, the whole subtask graph may never got finsihed or got pretty slowly by some subtasks. Task speculative execution mode will detect those subtasks and try to schedule it to other workers at the same time. In this way , the whole system will execute much more quickly and stable.

### Main changes includes:
* subtask unique and determinate id for its compute logic
* group subtask by compute logic so that we can base on finished homogeneous subtaks to determine whether a running subtask got stale/slow/hang.
* Slow/stale node detection and remove
* Support submit subtask by excluding some workers
* concurent and duplicate run and isolation for subtask at the same time.
* Speculative execution for subtask

### How to use
```yaml
scheduling:
  speculation:
    # Enables (yes) or disables (no) speculative execution of subtasks
    enabled: no
    # Don't submit subtasks actually for slow subtasks
    dry: no
    # The time interval seconds to use before checking for speculative subtasks.
    interval: 5
    # The percentage of subtasks that has not finished yet at which to start speculation.
    threshold: 75%
    # Minimum amount of time seconds a task runs before being considered for speculation.
    # This can be used to avoid launching speculative copies of tasks that are very short.
    min_task_runtime: 3
    # How many times slower a task is than the median to be considered for speculation.
    multiplier: 1.5
    # Max number of concurrent speculative run for a subtask.
    max_concurrent_run: 3
    enable_kill_slot: yes
```

## Related issue number
#2112 
Closes #2575 
## Check code requirements


mars/core/operand/base.py
```diff
@@ -142,6 +142,12 @@ class Operand(Base, metaclass=OperandMetaclass):
     _inputs = ListField("inputs", FieldTypes.reference(EntityData))
     _outputs = ListField("outputs")
     _output_types = ListField("output_type", FieldTypes.reference(OutputType))
+    # An unique and deterministic key for subtask compute logic. It should be even the same
+    # for different run if the compute logic doesn't change. This id will be used in subtask speculative
+    # execution and hbo scheduling and so on.
+    logic_key = StringField('logic_key')
```
wjsi, line 148: Speculation is part of execution strategy, and it is better to separate them from core data structure.

mars/oscar/core.pyx
```diff
@@ -30,14 +30,14 @@ CALL_METHOD_BATCH = 1
 logger = logging.getLogger(__name__)
 
 cdef:
-    bint _log_unhandled_errors = False
-    bint _log_cycle_send = False
+    bint _log_unhandled_errors = True
+    bint _log_cycle_send = True
```
wjsi, line 34: Make these to False to avoid expensive error detection.

mars/services/meta/store/dictionary.py
```diff
@@ -47,16 +48,24 @@ async def create(cls, config) -> Dict:
...
-        self._store[object_id] = meta
-        if isinstance(meta, _ChunkMeta):
-            for band in meta.bands:
-                self._band_chunks[band].add(object_id)
+    def _set_meta(self,
+                  object_id: str,
+                  meta: _CommonMeta,
+                  timestamp: int = None):
+        if object_id not in self._store:
+            self._store_timestamp[object_id] = timestamp
```
wjsi, line 56: Data not recycled for `_store_timestamp`.

mars/services/meta/api/oscar.py
```diff
@@ -138,20 +138,22 @@ async def set_chunk_meta(
         memory_size: int = None,
         store_size: int = None,
         bands: List[BandType] = None,
+        timestamp: int = None,
```
wjsi, line 141: The design purpose for `timestamp` field need to be illustrated in detail.

mars/services/scheduling/supervisor/assigner.py
```diff
@@ -164,7 +167,14 @@ async def assign_subtasks(self, subtasks: List[Subtask]):
                         max_size = size
                     elif size == max_size:
                         bands.append(band)
-            assigns.append(random.choice(bands))
+            band = random.choice(bands)
+            if band in exclude_bands and exclude_bands_force:
+                raise NoBandAvailable(f'No bands available for subtask {subtask.subtask_id} after '
+                                      f'excluded {exclude_bands}')
+            if subtask.bands_specified and band not in subtask.expect_bands:
+                raise Exception(f'No bands available for subtask {subtask.subtask_id} on bands {subtask.expect_bands} '
```
wjsi, line 175: Why not use `NoBandAvailable`?

mars/services/scheduling/supervisor/assigner.py
```diff
@@ -85,32 +85,36 @@ def _get_device_bands(self, is_gpu: bool):
...
+    def _get_random_band(self,
+                         is_gpu: bool,
+                         exclude_bands: Set[BandType] = None,
+                         exclude_bands_force: bool = False):
+        if exclude_bands:
+            avail_bands = [band for band in self._get_device_bands(is_gpu) if band not in exclude_bands]
+            if avail_bands:
+                return random.choice(avail_bands)
+            elif exclude_bands_force:
+                raise Exception(f'No bands available after excluding bands {exclude_bands}')
```
wjsi, line 97: Ditto.

mars/services/scheduling/supervisor/manager.py
```diff
@@ -56,35 +68,150 @@ def to_summary(self, **kwargs) -> SubtaskScheduleSummary:
...
-    ):
+    def __init__(self,
+                 session_id: str,
+                 subtask_max_reschedules: int = DEFAULT_SUBTASK_MAX_RESCHEDULES,
+                 subtask_cancel_timeout: int = 5,
+                 speculation_config: Dict[str, object] = None):
         self._session_id = session_id
         self._subtask_infos = dict()
         self._subtask_summaries = dict()
+        self._grouped_unfinished_subtasks = defaultdict(dict)
```
wjsi, line 87: Putting all speculation strategy code in `SubtaskManagerActor` directly seems too heavy for me any this may not be easy to write unit tests for these strategies. Is it possible to create some abstractions?

mars/services/scheduling/supervisor/manager.py
```diff
@@ -129,18 +255,47 @@ async def _get_execution_ref(self, band: BandType):
...
-            subtask_info = self._subtask_infos.pop(subtask_id, None)
+        bands = bands or [None] * len(subtask_ids)
+        for subtask_id, subtask_band in zip(subtask_ids, bands):
+            subtask_info = self._subtask_infos.get(subtask_id, None)
             if subtask_info is not None:
-                self._subtask_summaries[subtask_id] = subtask_info.to_summary(
-                    is_finished=True
-                )
+                self._subtask_summaries[subtask_id] = subtask_info.to_summary(is_finished=True)
+                subtask_info.end_time = time.time()
```
wjsi, line 269: Ditto. Making speculation strategy scattered in actors may not be a good practice.

wjsi (changes requested):

mars/core/operand/base.py
chaokunyang: It's also used in hbo(history based optimization) for match history subtasks.

mars/oscar/core.pyx
chaokunyang: reverted

mars/services/meta/api/oscar.py
chaokunyang: It's a version which can be an optimistic lock for set/del meta for same object. If  version is not None and not same, delete chunk meta will be skipped

mars/services/scheduling/worker/execution.py
```diff
@@ -381,11 +386,19 @@ async def _run_subtask_once():
                     await mo.wait_actor_pool_recovered(sub_pool_address, self.address)
                 raise ex
             finally:
-                logger.debug("Subtask running ended, slot_id=%r", slot_id)
+                # make sure allocated slots are traced
+                if slot_id is None:  # pragma: no cover
+                    slot_id = await slot_manager_ref.get_subtask_slot_by_ts(timestamp)
```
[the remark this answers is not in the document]
chaokunyang: `acquire_free_slot` may got failed, this the slot _id is None,  but the status in `worker_slot_manager_ref` may get updated. In such case, we need a way to get the acquired slot id, so we can release the slot. timestamp is used for this kind of case, so we can know which slot is the slot acquired by us.

mars/services/task/supervisor/processor.py
```diff
@@ -755,14 +769,49 @@ async def _decref_input_subtasks(
...
+            prev_result.status == SubtaskStatus.succeeded
+            or prev_result.progress > subtask_result.progress
+        ):
+            logger.info(
+                "Skip set subtask %s with result %s, previous result is %s.",
+                subtask.subtask_id,
+                subtask_result,
+                prev_result,
+            )
+            # For duplicate run of subtasks, if the progress or s
```
Catch-Bull, line 806: forget delete？

mars/services/scheduling/supervisor/assigner.py
```diff
@@ -164,7 +191,18 @@ async def assign_subtasks(self, subtasks: List[Subtask]):
...
                         bands.append(band)
-            assigns.append(random.choice(bands))
+            band = random.choice(bands)
+            if band in exclude_bands and exclude_bands_force:
+                raise NoAvailableBand(
+                    f"No bands available for subtask {subtask.subtask_id} after "
+                    f"excluded {exclude_bands}"
+                )
+            if subtask.bands_specified and band not in subtask.expect_bands:
+                raise Exception(
```
fyrestone, line 201: Why use Exception instead of NoAvailableBand?

mars/services/scheduling/supervisor/assigner.py
```diff
@@ -164,7 +191,18 @@ async def assign_subtasks(self, subtasks: List[Subtask]):
                         max_size = size
                     elif size == max_size:
                         bands.append(band)
-            assigns.append(random.choice(bands))
+            band = random.choice(bands)
+            if band in exclude_bands and exclude_bands_force:
```
fyrestone, line 195: Cheap eval first, exclude_bands_force and band in exclude_bands is better.

mars/services/scheduling/supervisor/manager.py
```diff
@@ -86,6 +102,14 @@ async def __post_create__(self):
         self._global_slot_ref = await mo.actor_ref(
             GlobalSlotManagerActor.default_uid(), address=self.address
         )
+        self._speculation_execution_scheduler = SpeculativeScheduler(
+            self._queueing_ref, self._global_slot_ref, self._speculation_config
+        )
+        await self._speculation_execution_scheduler.start()
+
+    async def __pre_destroy__(self):
+        if self._subtask_speculation_enabled:
```
fyrestone, line 111: Why `if self._subtask_speculation_enabled`? If `self._speculation_execution_scheduler` exists, we should stop it.

mars/services/scheduling/supervisor/queueing.py
```diff
@@ -222,12 +238,17 @@ async def submit_subtasks(self, band: Tuple = None, limit: Optional[int] = None)
...
     def _ensure_top_item_valid(self, task_queue):
-        """Clean invalid subtask item from queue to ensure that"""
+        """Clean invalid subtask item from the queue to ensure that when the queue is not empty,
+        there is always some subtasks waiting being scheduled."""
         while (
             task_queue and task_queue[0].subtask.subtask_id not in self._stid_to_items
         ):
             #  skip removed items (as they may be re-pushed into the queue)
             heapq.heappop(task_queue)
+        if task_queue:
```
fyrestone, line 248: return bool(task_queue)

mars/services/scheduling/supervisor/assigner.py
chaokunyang: fixed

mars/services/scheduling/supervisor/assigner.py
chaokunyang: Good catch

fyrestone (dismissed):
This feature has been tested in practice. LGTM, I just leave some comments.

mars/deploy/oscar/session.py
```diff
@@ -916,7 +916,7 @@ async def _run_in_background(
                 tileable.params = fetch_tileable.params
 
     async def execute(self, *tileables, **kwargs) -> ExecutionInfo:
-        fuse_enabled: bool = kwargs.pop("fuse_enabled", True)
+        fuse_enabled: bool = kwargs.pop("fuse_enabled", None)
```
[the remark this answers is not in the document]
chaokunyang: If set default value to `True`, the `fuse_enabled ` config specified in cluster creation won't take effect

mars/dataframe/base/apply.py
```diff
@@ -84,6 +84,25 @@ def __init__(
...
+            self._elementwise,
+        ]
+        if self.func:
+            if hasattr(self.func, "__code__"):
+                return token_values + [self.func.__code__]
+            else:  # pragma: no cover
+                return token_values + [
+                    self.func.__class__.__module__,
+                    self.func.__class__.__name__,
+                ]
```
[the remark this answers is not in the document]
chaokunyang: cython functions doesn't have `co_code` properties, but we do should use `co_code` and `cell_contents` for normal function.

mars/core/operand/base.py
```diff
@@ -153,6 +157,30 @@ def __init__(self: OperandType, *args, **kwargs):
         self._extract_scheduling_hint(kwargs)
         super().__init__(*args, **kwargs)
 
+    def _update_key(self):
+        super()._update_key()
+        if not hasattr(self, "logic_key") or not self.logic_key:
+            self._obj_set("logic_key", self._get_logic_key())
```
qinxuye, line 163: Do we need to generate such a `logic_key` every time creating an operand? How about generating it when a subtask is created?

mars/core/operand/fuse.py
```diff
@@ -25,6 +25,11 @@ class Fuse(Operand):
 
     fuse_graph = ReferenceField("fuse_graph", ChunkGraph)
 
+    def _get_logic_key_token_values(self):
```
qinxuye, line 28: Now the fuse op will no longer be generated during fusing when coloring algorithm is introduced.

qinxuye (commented):
I haven't dig into the speculation logic, but for such a `logic_key`, I have some comments.

First, I don't think it's a good idea to add such a property to operand directly, because it's actually bound to the scheduling, thus I recommend to add it to `SchedulingHint`, and when scheduling subtasks, the property can be used for speculative scheduling.

Second, about generating logic keys, I don't think it's necessary to generate it instantly when operands are created, I'd prefer to generate it during subtask generation. And, I guess the generating logic can be separated from operand. 

For instance, 

```
op_logic_key_generators: Dict[Type[Operand], Callable] = {}

class OperandLogicKeyGenerator:
    def gen_logic_key(self) -> str:
        pass

op_logic_key_generators[Operand] = OperandLogicKeyGenerator
op_logic_key_generators[OtherOperand] = OtherGenerator
```

The generator logic can be accompany with or without operand itself, but for those developer, they don't have to care about it.

mars/core/operand/base.py
```diff
@@ -153,6 +157,30 @@ def __init__(self: OperandType, *args, **kwargs):
...
+
+    def _get_logic_key(self):
+        """The subclass may need to override this method to ensure unique and deterministic."""
+        fields = self._get_logic_key_token_values()
+        try:
+            return tokenize(*fields)
+        except Exception as e:  # pragma: no cover
+            raise Exception(
+                f"Cannot generate logic key for operator {self} with fields {fields}"
+            ) from e
```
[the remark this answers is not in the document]
chaokunyang: Is there a way to keep the exception?

mars/conftest.py
```diff
@@ -63,7 +65,16 @@ def ray_large_cluster(request):  # pragma: no cover
             cluster.add_node(num_cpus=num_cpus, memory=num_cpus * 2 * 1024**3)
         )
         if len(remote_nodes) == 1:
-            ray.init(address=cluster.address)
+            sig = inspect.signature(ray.job_config.JobConfig)
```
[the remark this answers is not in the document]
chaokunyang: Good suggestion, I'll fix it

mars/core/operand/base.py
```diff
@@ -153,6 +157,30 @@ def __init__(self: OperandType, *args, **kwargs):
         self._extract_scheduling_hint(kwargs)
         super().__init__(*args, **kwargs)
 
+    def _update_key(self):
+        super()._update_key()
+        if not hasattr(self, "logic_key") or not self.logic_key:
```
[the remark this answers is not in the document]
chaokunyang: Fixed

chaokunyang:
> 

The `logic_key` represents the compute logic for an operator. If we separate the `logic_key` generation code from the  operator, it may introduce extra maintanance cost, diverge from the real operator logic. For example, for `DataFrameToSQLTable` operand, the `logic_key` generation code is 
```
    def _get_logic_key_token_values(self):
        fields_to_tokenize = [
            getattr(self, k, None)
            for k in [
                "_table_name",
                "_schema",
                "_if_exists",
                "_index",
                "_index_label",
                "_chunksize",
                "_dtype",
                "_method",
            ]
        ]
        return super()._get_logic_key_token_values() + fields_to_tokenize
```
which is highly corelated to the inner inplementation of the operator. It't very easy that key generation code diverge from the `DataFrameToSQLTable` inplementation if we put the code in another place.

mars/core/operand/fuse.py
chaokunyang: So we should remove the fuse op?

mars/core/operand/base.py
chaokunyang: `logic_key` generation can be made lazy. How about:
```
    @functools.lru_cache
    def get_logic_key(self):
        """The subclass may need to override this method to ensure unique and deterministic."""
        fields = self._get_logic_key_token_values()
        try:
            return tokenize(*fields)
        except Exception as e:  # pragma: no cover
            raise Exception(
                f"Cannot generate logic key for operator {self} with fields {fields}"
            ) from e
```

mars/core/operand/base.py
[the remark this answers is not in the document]
fyrestone: Currently, `with_traceback` works better than raise ... from ...

mars/services/scheduling/supervisor/manager.py
```diff
@@ -327,3 +404,242 @@ def get_schedule_summaries(self, task_id: Optional[str] = None):
...
+            self._grouped_unfinished_subtasks.pop(subtask.logic_id, None)
+            logger.info(
+                "Subtask group with logic id %s parallelism %s finished.",
+                subtask.logic_id,
+                subtask.parallelism,
+            )
+
+    async def _speculative_execution(self):
+        while True:
+            await asyncio.sleep(self._subtask_speculation_interval)
```
wjsi, line 498: Add comments to describe what is going on in code below.

mars/services/scheduling/supervisor/manager.py
```diff
@@ -103,6 +126,8 @@ async def add_subtasks(self, subtasks: List[Subtask], priorities: List[Tuple]):
                 )
                 if subtask_max_reschedules is None:
                     subtask_max_reschedules = self._subtask_max_reschedules
+                if subtask.subtask_id in self._subtask_infos:
+                    raise Exception(f"Subtask {subtask.subtask_id} already added.")
```
wjsi, line 130: Use a detailed error type, for instance, `KeyError`.

wjsi (changes requested):

mars/deploy/oscar/session.py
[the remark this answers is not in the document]
chaokunyang: There are other default values configured in function params, this change doesn't break fusions

.github/workflows/core-ci.yml
```diff
@@ -64,7 +64,6 @@ jobs:
           conda install -n test --quiet --yes -c pkgs/main python=$PYTHON certifi
         fi
         conda list -n test
-
```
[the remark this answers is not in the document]
chaokunyang: reverted

mars/deploy/oscar/tests/test_ray_advanced.py
```diff
@@ -0,0 +1,63 @@
+# Copyright 1999-2021 Alibaba Group Holding Ltd.
```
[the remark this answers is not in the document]
chaokunyang: changed to `test_ray_scheduling.py`

mars/services/subtask/worker/processor.py
```diff
@@ -401,37 +402,23 @@ async def _store_meta(
...
-                    stored_keys,
-                    self.subtask.subtask_id,
-                )
-                raise
+            try:
+                # Since we don't delete chunk data on this worker, we need to ensure chunk meta are recorded
+                # in meta service, so that `processor.decref_stage` can delete the chunk data finally.
+                await asyncio.shield(set_chunks_meta())
+            except asyncio.CancelledError:  # pragma: no cover
+                await f
```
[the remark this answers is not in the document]
chaokunyang: Good catch, thanks a lot

mars/services/scheduling/supervisor/assigner.py
```diff
@@ -85,11 +85,34 @@ def _get_device_bands(self, is_gpu: bool):
...
         return filtered_bands
 
-    def _get_random_band(self, is_gpu: bool):
-        avail_bands = self._get_device_bands(is_gpu)
-        return random.choice(avail_bands)
+    def _get_random_band(
+        self,
+        is_gpu: bool,
+        exclude_bands: Set[BandType] = None,
+        exclude_bands_force: bool = False,
```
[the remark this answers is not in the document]
chaokunyang: Actually if `exclude_bands_force` is `False`, we return a random which is not empty, so maybe we should keep using `exclude_bands_force` or something else? @wjsi

mars/oscar/core.pyx
```diff
@@ -425,9 +425,9 @@ cdef class _BaseActor:
         try:
             res = None
             while True:
-                with debug_async_timeout('actor_lock_timeout',
-                                         'async_generator %r hold lock timeout', gen):
-                    async with self._lock:
+                async with self._lock:
```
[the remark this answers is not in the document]
chaokunyang: It's used to detect actor lock timeout, should I remove it in this PR?

mars/oscar/debug.py
```diff
@@ -175,4 +177,32 @@ def set_message_trace(message_trace):
...
+    frame = inspect.currentframe()
+    if frame and frame.f_back:
+        call_site = frame.f_back.f_code
+    else:
+        call_site = None
+    return _create_task(task_with_ex_logged(coro, call_site), *args, **kwargs)
+
+
 reload_debug_opts_from_env()
+sys.setrecursionlimit(100000)
```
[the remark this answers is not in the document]
chaokunyang: This is used to avoid `stack` overflow error which we met in out inner case because there are too many async calls. Do I need to remove it from this PR? Or just add some comments here? @wjsi

mars/oscar/debug.py
```diff
@@ -175,4 +177,32 @@ def set_message_trace(message_trace):
...
+        logger.exception(
+            "Coroutine %r at call_site %s execution got exception %s.",
+            coro,
+            call_site,
+            e,
+        )
+        raise
+
+
+def create_task_with_ex_logged(coro, *args, **kwargs):  # pragma: no cover
```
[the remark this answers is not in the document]
chaokunyang: Moved to `mars.utils`

mars/services/scheduling/worker/workerslot.py
```diff
@@ -165,9 +183,12 @@ def release_free_slot(self, slot_id: int, session_stid: Tuple[str, str]):
                 f"the acquired session_stid: {acquired_session_stid}, "
                 f"the releasing session_stid: {session_stid}"
             )
-        acquired_slot_id = self._session_stid_to_slot.pop(acquired_session_stid)
-        assert acquired_slot_id == slot_id
-
+        slots = self._session_stid_to_slots[acquired_session_stid]
+        assert slot_id in slots
+        slots.remove(slot_id)
+        self._timestamp_to_slot.pop(self._slot_to_timestamp.pop(slot_id), None)
```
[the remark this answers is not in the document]
chaokunyang: If two subtasks of same `subtask_id` running in the same worker, we can't distinguish which slot by subtask_id to cancel.

mars/services/subtask/core.py
```diff
@@ -61,6 +61,15 @@ class Subtask(Serializable):
     priority: Tuple[int, int] = TupleField("priority", FieldTypes.int32)
     rerun_time: int = Int32Field("rerun_time")
     extra_config: dict = DictField("extra_config")
+    stage_id: str = StringField("stage_id")
+    # An unique and deterministic key for subtask compute logic. See logic_key in operator.py.
+    logic_id: str = StringField("logic_id")
+    # index for subtask with same compute logic.
+    index: int = Int32Field("index")
+    # parallelism for subtask with same compute logic.
+    parallelism: int = Int32Field("parallelism")
```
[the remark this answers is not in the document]
chaokunyang: Ths `logic_parallelism` won't change after a subtask graph is generated. So there won't be synchoronizing issue if I understand correctly?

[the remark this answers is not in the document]
chaokunyang: `parallelism` represents actual parallelism of subtasks with same logic, naming it as `logic_parallelism` seems a little wired for me? Do we need to change to `logic_parallelism`?

mars/services/subtask/core.py
```diff
@@ -61,6 +61,15 @@ class Subtask(Serializable):
     priority: Tuple[int, int] = TupleField("priority", FieldTypes.int32)
     rerun_time: int = Int32Field("rerun_time")
     extra_config: dict = DictField("extra_config")
+    stage_id: str = StringField("stage_id")
+    # An unique and deterministic key for subtask compute logic. See logic_key in operator.py.
+    logic_id: str = StringField("logic_id")
+    # index for subtask with same compute logic.
+    index: int = Int32Field("index")
```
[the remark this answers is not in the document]
chaokunyang: Same as `logic_parallelism `

mars/oscar/core.pyx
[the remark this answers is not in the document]
wjsi: It's Okay.

mars/services/scheduling/supervisor/assigner.py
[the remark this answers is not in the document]
wjsi: The name `force` itself is not clear enough as what is called `force`. As the behavior actually is that when the flag is on and no bands available, the function should raise instead of returning a random band instead.

mars/oscar/debug.py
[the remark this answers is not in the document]
wjsi: Then you need to diagnose the reason why there is such a huge stack instead of adding a flag to suppress the error. I propose removing it.

mars/services/scheduling/worker/workerslot.py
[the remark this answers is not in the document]
wjsi: > If two subtasks of same `subtask_id` running in the same worker, we can't distinguish which slot by subtask_id to cancel.

This scenario need to be avoided at least in the same band. That is, we need to assume that a band is homogeneous inside itself and running multiple instances inside the same band shall be avoided.

mars/services/subtask/core.py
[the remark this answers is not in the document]
wjsi: > `parallelism` represents actual parallelism of subtasks with same logic, naming it as `logic_parallelism` seems a little wired for me? Do we need to change to `logic_parallelism`?

`parallelism` is actually a variable for subtask groups, not subtasks. Thus `parallelism` is not parallelism inside the subtask itself but a task-level definition. Thus clarification is needed.

> Ths `logic_parallelism` won't change after a subtask graph is generated. So there won't be synchoronizing issue if I understand correctly?

This assertion might be valid for now, but it may limit future implementations of dynamic splitting esp. when handling skewness in join operands, as you need to synchronize between different subtasks.

mars/services/subtask/core.py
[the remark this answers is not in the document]
wjsi: > Same as `logic_parallelism `

Might be confuising as there is also a `index` with output chunks.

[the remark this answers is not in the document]
chaokunyang: Make sense. changed to `logic_index`

mars/services/subtask/core.py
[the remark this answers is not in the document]
chaokunyang: data skew and dynamic splitting up is a complex topic. for example, in spark, `data skew` will be handled using following ways:
* Dynamically changes sort merge join into broadcast hash join.
* Dynamically coalesces partitions (combine small partitions into reasonably sized partitions) after shuffle exchange. Very small tasks have worse I/O throughput and tend to suffer more from scheduling overhead and task setup overhead. Combining small tasks saves resources and improves cluster throughput.
* Dynamically handles skew in sort merge join and shuffle hash join by splitting (and replicating if needed) skewed tasks into roughly evenly sized tasks.
* Dynamically detects and propagates empty relations.
This is a complex topic, I think we can keep current design, and deal with it in later PR when we do some dynamic re-scheduling?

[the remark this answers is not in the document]
chaokunyang: `parallelism` changed to `logic_parallelism `

mars/oscar/debug.py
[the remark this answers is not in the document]
chaokunyang: removed

mars/services/scheduling/supervisor/assigner.py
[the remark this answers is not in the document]
chaokunyang: `allow_empty ` gives me a sense this method will return None if there is no available bands after excluded `exclude_bands` when `allow_empty` is True, but it return a ramdom band actually.

mars/services/scheduling/supervisor/speculation.py
```diff
@@ -0,0 +1,267 @@
...
+                subtask.logic_parallelism,
+            )
+
+    async def _speculative_execution(self):
+        while True:
+            # check subtasks in the same group which has same logic key periodically, if some subtasks hasn't been
+            # finished in a considerably longer duration, then those subtasks maybe slow/hang subtasks, try resubmit
+            # it to other bands too.
+            await asyncio.sleep(self._subtask_speculation_interval)
+            for logic_id, subtask_infos_dict in dict(
```
wjsi, line 121:
```suggestion
            for logic_key, subtask_infos_dict in dict(
```

mars/services/scheduling/supervisor/speculation.py
```diff
@@ -0,0 +1,267 @@
...
+                self._grouped_finished_subtasks
+            ).items():
+                if subtask_infos_dict:
```
wjsi, line 124: To reduce indentation,

```suggestion
                if not subtask_infos_dict:
                    continue
```

mars/services/scheduling/supervisor/speculation.py
```diff
@@ -0,0 +1,267 @@
...
+                    one_subtask = next(iter(subtask_infos)).subtask
+                    parallelism = one_subtask.logic_parallelism
+                    spec_threshold = max(
+                        1, int(self._subtask_speculation_threshold * parallelism)
+                    )
+                    if parallelism > len(subtask_infos) >= spec_threshold:
+                        unfinished_subtask_infos = self._grouped_unfinished_subtasks[
+                            logic_id
+                        ].values()
+                        duration_array = np.sort(
```
wjsi, line 135: Write simple comments about what is calculated here.

mars/services/scheduling/supervisor/speculation.py
```diff
@@ -0,0 +1,267 @@
...
+                            self._subtask_speculation_min_task_runtime,
+                        )
+                        now = time.time()
+                        unfinished_subtask_infos = [
+                            info
+                            for info in unfinished_subtask_infos
+                            if info not in subtask_infos
+                            and now - info.start_time > duration_threshold
+                        ]
+                        if unfinished_subtask_infos:
```
wjsi, line 155: Reduce indentation by `continue` when `unfinished_subtask_infos` is False.

mars/services/scheduling/supervisor/speculation.py
```diff
@@ -0,0 +1,267 @@
...
+                            exclude_bands = set()
+                            for info in unfinished_subtask_infos:
+                                for band in info.band_futures.keys():
+                                    exclude_bands.add(band)
```
wjsi, line 159:
```suggestion
                                exclude_bands.update(info.band_futures.keys())
```

mars/services/scheduling/supervisor/speculation.py
```diff
@@ -0,0 +1,267 @@
...
+        if (
+            subtask_info.num_speculative_concurrent_run
+            == self._subtask_speculation_max_concurrent_run
+        ):
+            logger.debug(
+                "Subtask %s speculative run has reached max limit %s, "
+                "won't submit another speculative run.",
+                subtask.subtask_id,
+                self._subtask_speculation_max_concurrent_run,
+            )
```
wjsi, line 214: Return directly.

mars/services/scheduling/supervisor/speculation.py
```diff
@@ -0,0 +1,267 @@
...
+                    < subtask_info.num_speculative_concurrent_run + 1
+                ):
+                    # ensure same subtask won't be submitted to same worker.
+                    logger.info(
+                        "Speculative execution for subtask %s has not been submitted to worker,"
+                        "waiting for being submitted to worker."
+                        "Cluster resources may be not enough after excluded %s",
+                        subtask.subtask_id,
+                        exclude_bands,
+                    )
```
wjsi, line 228: Ditto.

mars/services/scheduling/supervisor/speculation.py
chaokunyang: Good suggestion, I'll fix it

chaokunyang:
There is a vinyard error:
```
================================================================================== FAILURES ===================================================================================
______________________________________________________________________ test_vineyard_operators[vineyard] ______________________________________________________________________

create_cluster = (<mars.deploy.oscar.local.LocalClient object at 0x7f9d79b82910>, 'vineyard')

    @pytest.mark.asyncio
    async def test_vineyard_operators(create_cluster):
        param = create_cluster[1]
        if param != "vineyard":
            pytest.skip("Vineyard is not enabled")
    
        session = get_default_async_session()
    
        # tensor
        raw = np.random.RandomState(0).rand(55, 55)
        a = mt.tensor(raw, chunk_size=15)
        info = await session.execute(a)  # n.b.: pre-execute
        await info
    
        b = mt.to_vineyard(a)
        info = await session.execute(b)
        await info
        object_id = (await session.fetch(b))[0]
    
        c = mt.from_vineyard(object_id)
        info = await session.execute(c)
        await info
>       tensor = await session.fetch(c)

mars/deploy/oscar/tests/test_local.py:161: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
mars/deploy/oscar/session.py:1410: in fetch
    return await asyncio.wrap_future(
mars/deploy/oscar/session.py:1822: in _fetch
    data = await session.fetch(tileable, *tileables, **kwargs)
mars/deploy/oscar/session.py:1104: in fetch
    merged = merge_chunks(index_to_data)
mars/utils.py:601: in merge_chunks
    (idx, xp.concatenate([c[1] for c in cs], axis=ndim - i - 1))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

args = ([ndarray([[0.5488135 , 0.71518937, 0.60276338, 0.54488318, 0.4236548 ,
          0.64589411, 0.43758721, 0.891773  , ...01, 1.49941390e-01, 7.54090366e-01, 7.19077024e-01,
          5.59570534e-01, 5.84644598e-01, 9.12707209e-02]]), ...],)
kwargs = {'axis': 1}
relevant_args = [ndarray([[0.5488135 , 0.71518937, 0.60276338, 0.54488318, 0.4236548 ,
          0.64589411, 0.43758721, 0.891773  , 0...e-01, 1.49941390e-01, 7.54090366e-01, 7.19077024e-01,
          5.59570534e-01, 5.84644598e-01, 9.12707209e-02]]), ...]

>   ???
E   ValueError: all the input array dimensions for the concatenation axis must match exactly, but along dimension 0, the array at index 0 has size 15 and the array at index 6 has size 10

<__array_function__ internals>:180: ValueError
```
@sighingnow Could you help take a look at it?

sighingnow:
> There is a vinyard error:
> ....
> @sighingnow Could you help take a look at it?

I can reproduce the failure and are working on it.

hekaisheng:
@sighingnow Could you open a separate PR to solve the issues about vineyard? This PR won't be backport.

sighingnow:
> 

Sure. Once the CI on this PR becomes green I will submit a new pull request for that patch.

sighingnow:
> @sighingnow Could you open a separate PR to solve the issues about vineyard? This PR won't be backport.

FYI: submitted in https://github.com/mars-project/mars/pull/2817

qinxuye (approved):
LGTM

wjsi (approved):
LGTM

PR #2589, opened 2021-12-01
Supports adding Mars extensions via setup entrypoints

fyrestone:
## What do these changes do?
This PR defines an entrypoint for mars extensions. So, some packages can implement the entrypoint to make mars auto load the package as extension. Just the same as SDC is an extension of numba. Please refer to SDC for more details: https://github.com/IntelPython/sdc/blob/master/setup.py#L440

There are 4 load points:
- LocalCluster
- RayCluster
- AbstractActorPool
- new_session

The entrypoints will be loaded only once in a process.

### Why not load entrypoints in mars.__init__?
- Entrypoints may import mars for extension, this will cause recursive import error.

## Related issue number
N/A

## Check code requirements

### DIFF
--- a/mars/core/entrypoints.py
+++ b/mars/core/entrypoints.py
@@ -1,0 +1,42 @@
+# Copyright 1999-2021 Alibaba Group Holding Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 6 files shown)


mars/core/entrypoints.py
```diff
@@ -0,0 +1,42 @@
...
+
+from pkg_resources import iter_entry_points
+
+logger = logging.getLogger(__name__)
+
+
+# from https://github.com/numba/numba/blob/master/numba/core/entrypoints.py
+# Must put this here to avoid extensions re-triggering initialization
+@functools.lru_cache(maxsize=None)
+def init_all():
```
wjsi, line 27: Use a more meaningful method name, for instance, `init_extension_entrypints`?

wjsi (commented):
LGTM overall, I left a comment.

mars/core/entrypoints.py
fyrestone: Yes, `init_extension_entrypoints` is better. Thanks.

wjsi (approved):
LGTM

qinxuye (approved):
LGTM

Issue #2590, opened 2021-12-01
[BUG] Progress always is 0 or 100%


fyrestone:
**Describe the bug**
![image](https://user-images.githubusercontent.com/6308809/144194999-c6b0bcfd-2824-479d-8a05-6cb31784ee8f.png)
The progress in notebook always is 0 or 100%

**To Reproduce**
To help us reproducing this bug, please provide information below:
1. Your Python version 3.7.7
2. The version of Mars you use latest master 14b9569c4f4a63a8386ab69a9d98dcefadc5c98f
3. Versions of crucial packages, such as numpy, scipy and pandas
4. Full stack of the error.
5. Minimized code to reproduce the error.

**Expected behavior**
Smooth progress update while computing.

**Additional context**
Notebook use web session as mars client to connect to a remote mars cluster.

PR #2591, opened 2021-12-01
Fix progress always is 0 or 100%

fyrestone:
## What do these changes do?
The web session use same timeout for client and server. Server timeout + RPC cost > client timeout, so the client call will always timeout, then the progress will not be updated.

- Always get progress even the previous call timeout.
- Set max(timeout / 2.0, timeout -1) for server.

## Related issue number

Fixes https://github.com/mars-project/mars/issues/2590

## Check code requirements


wjsi (approved):
LGTM

qinxuye (approved):
LGTM

PR #2599, opened 2021-12-03
Make Proxima work with latest Mars

yuyiming:
* The first arg of `TensorSlice` should be a set instead of a list;
* Most chunks in `ProximaBuilder` are bundled to specified workers, but only start-OPs are assigned during analyzing. I assign other chunks;

### DIFF
--- a/mars/services/task/analyzer/analyzer.py
+++ b/mars/services/task/analyzer/analyzer.py
@@ -303,6 +303,12 @@ class GraphAnalyzer:
             "Assigned %s start chunks for task %s", len(start_ops), self._task.task_id
         )
 
+        # assign expect workers for successors
+        for chunk in self._chunk_graph.topological_iter():
+            if chunk not in start_ops:
+                if chunk.op.expect_worker is not None:
+                    chunk_to_bands[chunk] = self._to_band(chunk.op.expect_worker)
+
         # fuse node
         if self._fuse_enabled:
             logger.debug("Start to fuse chunks for task %s", self._task.task_id)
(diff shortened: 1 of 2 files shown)


mars/services/task/analyzer/analyzer.py
```diff
@@ -303,6 +303,12 @@ def gen_subtask_graph(self) -> SubtaskGraph:
             "Assigned %s start chunks for task %s", len(start_ops), self._task.task_id
         )
 
+        # assign expect workers for successors
+        for chunk in self._chunk_graph.topological_iter():
```
hekaisheng, line 307: No need to use `topological_iter`, `for chunk in self._chunk_graph` is more efficient。

mars/services/task/analyzer/analyzer.py
```diff
@@ -303,6 +303,12 @@ def gen_subtask_graph(self) -> SubtaskGraph:
             "Assigned %s start chunks for task %s", len(start_ops), self._task.task_id
         )
 
+        # assign expect workers for successors
```
hekaisheng, line 306: "assign expect workers for those specified with expect_worker" seems clearer.

mars/services/task/analyzer/analyzer.py
chaokunyang, line 307: topological_iter is unnecessary

yuyiming: I've fixed this.

mars/services/task/analyzer/analyzer.py
yuyiming: good idea

mars/services/task/analyzer/analyzer.py
yuyiming: fixed

wjsi (approved):
LGTM

hekaisheng (approved):
LGTM

PR #2602, opened 2021-12-03
Add `get_cluster_info` method for ClusterAPI

zhongchun:
## What do these changes do?
There is no method to get worker node resources when you want to know the resources in the cluster. Especially when mars on ray, there is some confuse. Because we have a `RayActorDriver` which keeps initialized cluster info, but when we use `ClusterStateActor` to request or release worker, cluster info will change.  

So we should maintain cluster info like worker node resources in one place and ClusterAPI is just appropriate.

## Check code requirements


hekaisheng:
Could you add some cases to cover your code?

zhongchun:
> Could you add some cases to cover your code?

Ok.

hekaisheng:
This PR seems only for Ray backend, could you add the implementation for default backend?

zhongchun:
> This PR seems only for Ray backend, could you add the implementation for default backend?

How about  implement it in later pr? I opens an issue https://github.com/mars-project/mars/issues/2617 and I will implement it in the near future.

PR #2611, opened 2021-12-09
Fix duplicate decref of subtask input chunk

### LINKED ISSUE #2610: [BUG] race condition: duplicate decref of subtask input chunk
**Describe the bug**
Let us suppose stage A have two Subtask S1 and S2, and they have the same input chunk C

1. S1 got an error, and stage_processor.done has been set.
2. S2 call set_subtask_result, it already reduces reference count of C  but not set `stage_processor.results[C.key]`
3. `TaskProcessorActor` find `stage_processor` got an error and call `self._cur_processor.incref_stage(stage_processor)` in function `TaskProcessorActor.start`, it will also reduce the reference count of C which is input of S2.

**To Reproduce**
1. Your Python version: python 3.7.9
3. Minimized code to reproduce the error:

you can run this bash: `pytest mars/services/task/supervisor/tests/test_task_manager.py::test_error_task -s` in this branch: https://github.com/Catch-Bull/mars/tree/race_condition_case, which base latest mars master,and add some sleep to make reproducing easily

**Expected behavior**

**Additional context**

Catch-Bull:
* fix decrease subtask input chunk

This is a race condition.

Let us suppose stage A have two Subtask S1 and S2, and they have the same input chunk C

1. S1 got an error, and stage_processor.done has been set.
2. S2 call set_subtask_result, it already reduces reference count of C  but not set `stage_processor.results[C.key]`
3. `TaskProcessorActor` find `stage_processor` got an error and call `self._cur_processor.incref_stage(stage_processor)` in function `TaskProcessorActor.start`, it will also reduce the reference count of C which is input of S2.

## Related issue number

Fixes #2610

## Check code requirements

### DIFF
--- a/mars/services/task/supervisor/processor.py
+++ b/mars/services/task/supervisor/processor.py
@@ -215,8 +215,9 @@ class TaskProcessor:
             if error_or_cancelled:
                 # error or cancel, rollback incref for subtask results
                 for subtask in subtask_graph:
-                    if stage_processor.subtask_results.get(subtask):
+                    if subtask.subtask_id in stage_processor.decref_subtask:
                         continue
+                    stage_processor.decref_subtask.add(subtask.subtask_id)
                     # if subtask not executed, rollback incref of predecessors
                     for inp_subtask in subtask_graph.predecessors(subtask):
                         for result_chunk in inp_subtask.chunk_graph.results:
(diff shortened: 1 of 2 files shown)


wjsi (approved):
LGTM

hekaisheng (approved):
LGTM

PR #2690, opened 2022-02-09
[WIP] use direct async call for in-process rpc

### LINKED ISSUE #2691: Use direct async call for in-process rpc
Mars oscar actor uses at least one `asyncio.Task` for every actor call, which is expensive when there are thousands of tasks. `asyncio.Task` is about 100~200 slower than a normal async call, and the cost grows when there are more tasks in event loop. When there are hundreds of subtasks being schduled by the supervisor, there will be thousands of `asyncio.Task` in the event loop of supervisor, which make the supervisor as the bottleneck of the whole system.

For in-process actor call, it's possible to use direct async function call. Instead of creating an new `asyncio.Task` to process the message, we can just use the `asyncio.Task` of the caller to handle the message, thus reduced all extra `asyncio.Task`.

**Additional context**
For `TellMessage`, it's executed in the background and there is not an easy way to avoid the `asyncio.Task`, so`TellMessage` should be used sparely.

chaokunyang:
## What do these changes do?
This PR use the direct async function call to speed up in-progress actor call, and gives a speed up about 2.6 times.

**Benchmark code**:
```python
import asyncio
from contextlib import asynccontextmanager
import datetime
import os
import sys
import time

import mars.oscar as mo


class BenchmarkActor(mo.Actor):

    def __init__(self, value):
        super().__init__()
        self.value = value

    async def send(self, uid, method, iternum, *args):
        actor_ref = await mo.actor_ref(uid, address=self.address)
        value = None
        for _ in range(iternum):
            value = await getattr(actor_ref, method)(*args)
        return value

    async def inc(self, delta):
        self.value += delta
        return self.value

    def get_value(self):
        return self.value


@asynccontextmanager
async def actor_pool_context():
    start_method = (
        os.environ.get("POOL_START_METHOD", "forkserver")
        if sys.platform != "win32"
        else None
    )
    pool = await mo.create_actor_pool(
        "127.0.0.1", n_process=2, subprocess_start_method=start_method
    )
    await pool.start()
    yield pool
    await pool.stop()


async def test_dummy_call_benchmark(pool):
    ref1 = await mo.create_actor(BenchmarkActor, 1, address=pool.external_address)
    ref2 = await mo.create_actor(BenchmarkActor, 2, address=pool.external_address)
    await ref1.send(ref2, "inc", 2, 2)
    iternum = 100000
    expect = await ref2.get_value() + iternum * 2
    start = time.time()
    print(f"Start with iternum {iternum} at {datetime.datetime.now()}")
    assert await ref1.send(ref2, "inc", iternum, 2) == expect
    print(f"End with iternum {iternum} at {datetime.datetime.now()}, took {time.time() - start} seconds.")


async def main():
    async with actor_pool_context() as ctx:
        await test_dummy_call_benchmark(ctx)


if __name__ == '__main__':
    asyncio.run(main())
```

**Benchmark Env**
* 2.2 GHz 6-Core Intel Core i7
* 16 GB 2400 MHz DDR4
* OS: macOS BigSur 11.5.2 (20G95)

**Benchmark result**
* Without this PR: 100000 actor calls in 21.88 seconds
* WIth this PR: 100000 actor calls in 8.19 seconds

## Related issue number
Closes #2691 

## Check code requirements

### DIFF
--- a/.github/workflows/platform-ci.yml
+++ b/.github/workflows/platform-ci.yml
@@ -132,7 +132,7 @@ jobs:
             coverage combine build/ && coverage report
           fi
           if [ -n "$WITH_RAY" ]; then
-            pytest $PYTEST_CONFIG --timeout=120 -m ray
+            pytest $PYTEST_CONFIG --timeout=300 -m ray
             coverage report
           fi
           if [ -n "$RUN_DASK" ]; then
--- a/mars/oscar/backends/communication/__init__.py
+++ b/mars/oscar/backends/communication/__init__.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from .base import Client, Server, Channel
-from .core import get_client_type, get_server_type, gen_local_address
+from .core import get_client_type, get_server_type, gen_local_address, is_local_address
 from .dummy import DummyClient, DummyServer, DummyChannel
 from .socket import (
     SocketClient,
--- a/mars/oscar/backends/communication/__init__.py
+++ b/mars/oscar/backends/communication/__init__.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from .base import Client, Server, Channel
-from .core import get_client_type, get_server_type, gen_local_address, is_local_address
+from .core import get_client_type, get_server_type, gen_local_address, use_direct_call
 from .dummy import DummyClient, DummyServer, DummyChannel
 from .socket import (
     SocketClient,
--- a/mars/oscar/backends/communication/base.py
+++ b/mars/oscar/backends/communication/base.py
@@ -254,7 +254,7 @@ class Client(ABC):
         """
 
     @implements(Channel.send)
-    async def send(self, message):
+    async def send(self, message, wait_response: bool = False):
         return await self.channel.send(message)
 
     @implements(Channel.recv)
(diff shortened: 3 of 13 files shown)


wjsi:
Please open a new issue to illustrate your changes. What's more, any benchmarks for that?

chaokunyang:
> Please open a new issue to illustrate your changes. What's more, any benchmarks for that?

@wjsi I will open an issue after I finished the benchmark.

fyrestone:
Remove in-process actor will get more performance boost. But, that will change lots of code.

chaokunyang:
> Remove in-process actor will get more performance boost. But, that will change lots of code.

Yes, removing in-process actor will get more performance boost. Actually, after using direct call for in-process call, the main cost are the oscar itself. The cost is obvious from the dumped flame graph:
profile

wjsi:
> > Remove in-process actor will get more performance boost. But, that will change lots of code.
> 
> Yes, removing in-process actor will get more performance boost. Actually, after using direct call for in-process call, the main cost are the oscar itself. The cost is obvious from the dumped flame graph: (graph removed for simplicity)

I used to evaluate call performance with script below

```python
import mars
import mars.remote as mr
import time
from random import random

def f(i):
    pass
    # time.sleep(0.1)

if __name__ == '__main__':
    r = [mr.spawn(f, args=i) for i in range(1000)]
    _ = mars.execute(*r)

    mars.stop_server()
```

May run scripts like this and compare time cost before and after the change.

wjsi:
@chaokunyang any updates about this PR?

chaokunyang:
[](url)

PR #2719, opened 2022-02-16
Introduce coloring based fusion algorithm

### LINKED ISSUE #2435: [PROPOSAL] New fusion strategy based on coloring algorithm
# New fusion strategy based on coloring algorithm

## Background

From perspective of execution, Mars schedules subtask graph to workers to achieve parallelism. Each subtask is consist of one or more chunks except for fetch chunk.

Graph below illustrates the process:

image-20210906164438904

Till this proposal written, current fusion strategy is relatively simple, that is, try to fuse the straight line that has at least 2 nodes. So as illustrated in the graph, only chunk 3 and 4 can be fused, and the final subtask graph will be consist of 5 subtasks.

## New coloring based algorithm

This proposal tries to propose a new coloring based algorithm. Steps include:

1. Coloring the initial nodes according to the bands that assigned by assigner.
2. Propagate color in the topological order, if the input nodes have same color, color it with the same color; otherwise, color with a new color.
3. Propagate with reversed topological order, check a node with its inputs, if all inputs have different color with itself, skip; otherwise, if some of inputs have the same color, but some others have different color, color the input nodes with same one  with a new color, and propagate to its inputs and so on.

Below will show some examples.

### Example 1

image-20210907181827625

The ops of 1 and 2 colored with 1 and ops of 5 and 6 colored with 2, then color will propagate as topological order, op of 3 will notice only one kind of color received, then it will be colored with 1 too, so does op of 4. 

Propagate in a reverse topological order will not notice any inconsistency. Just skip.

Finally, the entire graph will only be colored with two colors, the nodes with same color can be fused, hence it will be separated into 2 subtasks.

### Example 2

image-20210907192948585

The ops of 1 and 2 colored with 1, and ops of 9 and 10 colored with 2. Propagate with topological order; 3, 4, 5 with color 1; 11, 12 with color 2; 6, 7 with color 3; 8 with color 4; 13 with color 5.

Then check in a reversed topological order, 3 will find that 4 has same color 1 but 5 has different color 3, conflict happens, so node 4 will be colored with a new color 6, propagate to node 5 with same color. So for the node 12, it will be colored with 7.

 the entire chunk graph will be consist of 7 subtasks.

## Configuration

Some configuration could be added.

1. Max size of same color for start ops. For the ops that allocated to same band, we should limit the size of same color, fuse too much may decrease the concurrency.
2. Max successors that can be propagated with same color. For those op comes up with plenty of successors, we should limit the size that can be propagated.

qinxuye:
## What do these changes do?

This PR introduced coloring based fusion algorithm described in #2435 .

## Related issue number

Fixes #2435 

## Check code requirements


hekaisheng (approved):
LGTM

mars/services/task/analyzer/fusion.py
```diff
@@ -12,101 +12,165 @@
...
+            )
+        ):
+            return False
+        return True
+
+    def _color_init_nodes(self) -> Dict[OperandType, int]:
+        # for initial op with same band but different priority
+        # we color them w/ different colors,
+        # to prevent from wrong fusion.
+        band_priority_to_colors = dict()
```
wjsi, line 76: A little strange using priority to do grouping here. Better to use distance measuring over undirected version of DAG to group initials.

mars/services/task/analyzer/fusion.py
```diff
@@ -12,101 +12,165 @@
...
-                selected.append(cur_node)
-                if (
-                    self._graph.count_successors(cur_node) != 1
-                    or cur_node in result_chunk_set
-                ):
-                    break
+
+            predecessors = self.chunk_graph.predecessors(chunk)
+            pred_colors = {op_to_colors[pred.op] for pred in predecessors}
+            if len(predecessors) == 1 and chunk_to_is_broadcaster.get(predecessors[0]):
```
wjsi, line 120: It is possible that a successor accepts a non-broadcaster input with a broadcast one. Under this scenario it is possible to get unexpected color result. Also need to consider if allocating a new color directly is oversimplified. Is it possible to use some grouping strategy instead or add some TODO items?

mars/services/task/analyzer/fusion.py
```diff
@@ -12,101 +12,165 @@
...
+    def color(self) -> Dict[ChunkType, int]:
+        chunk_to_colors = dict()
+
+        # step 1: Coloring the initial nodes according to the bands that assigned by assigner
+        op_to_colors = self._color_init_nodes()
+
+        # step2: Propagate color in the topological order,
+        # if the input nodes have same color, color it with the same color;
+        # otherwise, color with a new color.
+        chunk_to_is_broadcaster = dict()
```
wjsi, line 107: Use a `set` instead?

mars/services/task/analyzer/fusion.py
```diff
@@ -12,101 +12,165 @@
...
+                # conflict
+                # color the successors with new colors
+                stack = []
+                for succ in self.chunk_graph.iter_successors(chunk):
+                    if chunk_to_colors[succ] == chunk_color:
+                        chunk_to_colors[succ] = op_to_colors[
+                            succ.op
+                        ] = self._next_color()
+                        stack.extend(self.chunk_graph.successors(succ))
+                # color the descendants with same color to the new one
```
wjsi, line 156: I'm thinking over this iteration-in-iteration implementation and get an idea without proof that one node may only be iterated at least twice. Maybe some comments needed here to show that the time complexity does not degrade.

wjsi (commented):
Some clarifications needed.

wjsi (approved):
LGTM

hekaisheng (approved):
LGTM

PR #2730, opened 2022-02-21
Add address and pid prefix to the mars exception message

fyrestone:
## What do these changes do?

Developers need the address and pid for debug in a large distributed cluster, so this PR,

- [x] Add address and pid prefix to the mars exception message.

An example exception,

``` python
  File "mars/oscar/core.pyx", line 226, in _handle_actor_result
    result = await result
  File "/Users/admin/mars/services/tests/fault_injection_patch.py", line 86, in run
    return await super().run()
  File "/Users/admin/mars/services/subtask/worker/processor.py", line 457, in run
    await self._execute_graph(chunk_graph)
  File "/Users/admin/mars/services/subtask/worker/processor.py", line 195, in _execute_graph
    await self._async_execute_operand(self._datastore, chunk.op)
  File "/Users/admin/mars/services/tests/fault_injection_patch.py", line 94, in _async_execute_operand
    handle_fault(fault)
  File "/Users/admin/mars/services/tests/fault_injection_manager.py", line 50, in handle_fault
    raise FaultInjectionError("Fault Injection")
types._MarsError: [address=ray://test_cluster/1/2, pid=35418] Fault Injection
```

## Related issue number

Fixes #xxxx

## Check code requirements

### DIFF
--- a/mars/oscar/backends/message.py
+++ b/mars/oscar/backends/message.py
@@ -170,11 +170,22 @@ class ResultMessage(_MessageBase):
 
 
 class ErrorMessage(_MessageBase):
-    __slots__ = "error_type", "error", "traceback"
+    __slots__ = "address", "pid", "error_type", "error", "traceback"
+
+    # Check the as_instanceof_cause is not recursive.
+    #
+    # e.g. SubtaskRunnerActor.run_subtask will reraise the exception raised
+    # from SubtaskProcessorActor.run. But these two actors are in the same
+    # process, so we don't want to append duplicated address and pid in the
+    # error message.
+    class AsCauseBase:
+        pass
 
     def __init__(
         self,
         message_id: bytes,
+        address: str,
+        pid: int,
         error_type: Type[BaseException],
         error: BaseException,
         traceback: TracebackType,
@@ -182,6 +193,8 @@ class ErrorMessage(_MessageBase):
         message_trace: List[MessageTraceItem] = None,
     ):
         super().__init__(message_id, protocol=protocol, message_trace=message_trace)
+        self.address = address
+        self.pid = pid
         self.error_type = error_type
         self.error = error
         self.traceback = traceback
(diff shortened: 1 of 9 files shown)


mars/oscar/backends/message.py
```diff
@@ -191,6 +204,17 @@ def __init__(
...
         return MessageType.error
 
+    def as_instanceof_cause(self):
+        if issubclass(self.error_type, ErrorMessage.AsCauseBase):
+            return self.error.with_traceback(self.traceback)
+
+        class _MarsError(ErrorMessage.AsCauseBase, type(self.error)):
+            pass
+
+        return _MarsError(
```
wjsi, line 214: Wrapping existing errors has some side effects. For instance, some exception types have custom attributes, for instance, HTTP error codes. Wrapping with error messages hide these attributes and may lead to unexpected errors. Simply try adding attributes to these exceptions and log locations when handling exceptions at receiver end may be a better solution.

mars/oscar/backends/pool.py
```diff
@@ -59,8 +59,9 @@
 
 
 class _ErrorProcessor:
-    def __init__(self, message_id: bytes, protocol):
+    def __init__(self, message_id: bytes, address: str, protocol):
```
wjsi, line 62: Simply use `pool` as the argument?

mars/oscar/backends/message.py
fyrestone: Fixed. Thanks. I added a test case `test_as_instanceof_cause` for testing `as_instanceof_cause`.

mars/oscar/backends/pool.py
fyrestone: Currently, the `_ErrorProcessor` needs some message info and the address of current pool. Replacing `address` to `pool` does not simplify the constructor parameters.

wjsi (approved):
LGTM

hekaisheng (approved):
LGTM

PR #2741, opened 2022-02-22
[Bug] Fix duplicate dec object ref

Catch-Bull:
## What do these changes do?

Since the `report progress` will also set the subtask result, when the supervisor load is relatively high, the arrival time of the two set subtask results will be longer, resulting in the object ref being decremented one more time

## Related issue number

Fixes #xxxx

## Check code requirements


qinxuye (approved):
LGTM

hekaisheng (approved):
LGTM

PR #2759, opened 2022-02-27
Stop inferring outputs when args provided

### LINKED ISSUE #2758: make mars type inference optional
Mars type inferrence will generate mock data, then feed the data into user provided function, if the user function is time-consuming  which take minutes or hours, there will be much time wasting in type inference, and the real computing didn't even happen which is unacceptable in some cases.

It would be great if we can make dtypes lazy and type inferrence optional. If so, the expression call cost would be minimal.

wjsi:
## What do these changes do?

Stop inferring outputs when args provided

## Related issue number

Fixes #2758

## Check code requirements


qinxuye (approved):
LGTM

hekaisheng (approved):
LGTM

PR #2833, opened 2022-03-17
Implements `LocElimination` optimization rule

hekaisheng:
## What do these changes do?

This PR implements a new optimization rule, it will remove loc operand when it follows by merge operand. For pattern like this:
``` Python
indexed = df1[['col1', 'col2']]
merged = indexd.merge(df2, on='col1')
```
Coloring algorithm won't color Index node and merge node as one color,  so `indexed` data will be put into storage and brings overhead for storing, loading and transferring. This rule removes these indexing operands and move indexing into merge's execution.

## Related issue number

Fixes #xxxx

## Check code requirements

### DIFF
--- a/mars/core/graph/entity.py
+++ b/mars/core/graph/entity.py
@@ -34,6 +34,21 @@ class EntityGraph(DAG, metaclass=ABCMeta):
         results
         """
 
+    @results.setter
+    @abstractmethod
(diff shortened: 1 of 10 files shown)


hekaisheng:
Close as it has no improvements.

PR #2845, opened 2022-03-21
Fix store duplicate chunk and meta per subtask

fyrestone:
## What do these changes do?

There are duplicated data puts (same data key multiple puts),

1.  among different subtasks. _`A & B > 0`_
2. in one subtask. _`len(A) > len(set(A))`_

Mars put duplicate data due to duplicated result chunks. This PR dedup the result chunks by chunk key before put, so the  `#2` can be fixed.

The example code,
```python
import numpy as np
import pandas as pd
import mars
import mars.dataframe as md

if __name__ == "__main__":
    with mars.new_session():
        s = np.random.RandomState(0)
        raw = pd.DataFrame(s.rand(100, 4), columns=list("abcd"))
        df = md.DataFrame(raw, chunk_size=1)

        r = df.describe().execute(extra_config={"enable_profiling": True})
        print(r)
```

This PR can reduce,

- Data put count from 3070 to 2532.

## Related issue number

Fixes #xxxx

## Check code requirements

### DIFF
--- a/mars/services/subtask/worker/processor.py
+++ b/mars/services/subtask/worker/processor.py
@@ -284,20 +284,20 @@ class SubtaskProcessor:
         ]
 
         # store data into storage
-        puts = []
-        stored_keys = []
+        data_key_to_puts = {}
         for result_chunk in result_chunks:
             data_key = result_chunk.key
             if data_key in self._datastore:
                 # non shuffle op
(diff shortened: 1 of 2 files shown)


wjsi (approved):
LGTM

qinxuye (approved):
LGTM

PR #2862, opened 2022-03-24
Upgrade azure-pipelines to Python 3.9

### LINKED ISSUE #2781: [BUG] Potential memory leak in mo.extensible
**Describe the bug**
There is a function cache per descriptor created by mo.extensible. The function cache has 1000 capacity and will reference the decorated actor instance.

ActorType
    |
    |-foo (mo.extensible descriptor with capacity 1000)
    |_bar (mo.extensible descriptor with capacity 1000)

``` python
a1 = ActorType()
a1.foo  # The a1 is referenced in foo descriptor, and the descriptor is referenced by the ActorType.
a2 = ActorType()
a2.bar # The a2 is referenced in far descriptor, and the descriptor is referenced by the ActorType.
```

An example to reproduce this problem

```python
import asyncio
import mars.oscar as mo


class Actor(mo.Actor):
    @mo.extensible
    async def foo(self):
        return "foo"

    def __del__(self):
        print("destroy")


async def main():
    pool = await mo.create_actor_pool("127.0.0.1", n_process=0)
    ref = await mo.create_actor(Actor, address=pool.external_address)
    # Access extensible method will make the actor instance referenced in memory.
    print(await ref.foo())
    await mo.destroy_actor(ref)
    print(Actor.__dict__["foo"].__get__.__func__.cache_info())


loop = asyncio.get_event_loop()
loop.run_until_complete(main())
```
Expect result is
```
foo
destroy
```

**To Reproduce**
1. Your Python version 3.7.7
2. The version of Mars you use latest master
3. Versions of crucial packages, such as numpy, scipy and pandas
4. Full stack of the error.

qinxuye:
## What do these changes do?

Upgrade azure-pipelines to Python 3.9.

## Related issue number

Fixes #2781

## Check code requirements

### DIFF
--- a/.github/workflows/docker-cd.yml
+++ b/.github/workflows/docker-cd.yml
@@ -26,6 +26,7 @@ jobs:
         shell: bash
         env:
           DOCKER_ORG: ${{ secrets.DOCKERHUB_USERNAME }}
+        if: ${{ github.repository == 'mars-project/mars' }}
         run: |
           source ./ci/reload-env.sh
 
--- a/.github/workflows/pypi-cd.yml
+++ b/.github/workflows/pypi-cd.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Build wheels
         uses: pypa/cibuildwheel@v2.3.1
         env:
-          CIBW_BEFORE_BUILD: pip install -r ci/requirements-wheel.txt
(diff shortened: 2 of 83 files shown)


wjsi (approved):
LGTM

hekaisheng (approved):
LGTM

PR #2874, opened 2022-03-25
Optimized iterative tiling by pruning unrelated chunks

### LINKED ISSUE #2868: Optimize iterative tiling via pruning intermediate chunks
## Background

Iterative tiling is used during `tile` of an operand when the input tileable has unknown shape, or the operand want to trigger part of input chunks to perform an algorithm based on the input data. For instance, `df.groupby().agg()` will decide to use `tree` or `shuffle` according to the aggregated size of the first 4 chunks.

However, iterative tiling may trigger the execution of all inputs tileables and the memory usage may increase significantly. Still take the `df.groupby().agg()` as an example.

```python
import mars.dataframe as md
import mars.tensor as mt
df = md.DataFrame(mt.random.rand(1000, 10, chunk_size=100))
df.groupby(0).agg('sum').execute(extra_config={'dump_subtask_graph': True})
```

The graph shows:

image

Other than the 4 aggregated chunks that yielded inside `tile`, all the other 6 `DataFrameFromTensor` chunks are submitted as well, this may cause huge memory occupation.

## Solution

We suggest solutions as below:

1. If a `tile` triggered iterative tiling, the yielded chunks should be included in result chunks, if no chunks yielded, the input chunks are treated as result chunks.
2. If a tileable is in result tileables, its chunks are included in result chunks.
3. Otherwise, chunks that are not used by result chunks will be pruned.

qinxuye:
## What do these changes do?

This PR optimized iterative tiling by pruning unrelated chunks.

After this PR, the example in #2868 shows as below.

image

## Related issue number

Fixes #2868 .

## Check code requirements

### DIFF
--- a/mars/core/entity/tests/test_utils.py
+++ b/mars/core/entity/tests/test_utils.py
@@ -25,6 +25,8 @@ class _TestOperand(TensorOperand, TensorOperandMixin):
 
         data1 = mt.sort(data1)
         data2 = mt.sort(data2)
+        data1, data2 = yield from recursive_tile(data1, data2)
+        yield data1.chunks + data2.chunks + [data1, data2]
         data_all = mt.concatenate([data1, data2])
         s1 = mt.searchsorted(data1, data_all)
         s2 = mt.searchsorted(data2, data_all)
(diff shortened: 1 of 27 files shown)


wjsi (approved):
LGTM

hekaisheng (approved):
LGTM

PR #2895, opened 2022-04-01
Use bloom filter to optimize df.merge execution

hekaisheng:
## What do these changes do?

Use small DataFrame to build a bloom filter and filter the other DataFrame before merge, it will extremely reduce the number of data transfers when shuffling them. 

It's now a rough implementation, some details need to be confirmed.

## Related issue number

Fixes #xxxx

## Check code requirements


mars/dataframe/base/bloom_filter.py
```diff
@@ -0,0 +1,255 @@
...
+        return bloom_filter
+
+    @classmethod
+    def _build_dataframe_filter(cls, in_data, op):
+        try:
+            bloom_filter = BloomFilter(
+                max_elements=op.max_elements, error_rate=op.error_rate
+            )
+            in_data[op.on].apply(lambda v: bloom_filter.add(cls._get_value(v)), axis=1)
+        except TypeError:  # pragma: no cover
```
qinxuye, line 104: Better to add tests instead of skipping.

mars/dataframe/merge/merge.py
```diff
@@ -904,6 +982,9 @@ def join(
         When how is "inner", merged result could be much smaller than original DataFrame,
         if the number of chunks is greater than the threshold,
         it will merge small chunks automatically.
+    bloom_filter: bool or dict, default True
```
qinxuye, line 985: Add doc about if dict, which keys can have.

qinxuye (commented):
Coverage can still be improved.

mars/dataframe/merge/merge.py
qinxuye: Besides keys, the default values can be mentioned.

mars/dataframe/base/bloom_filter.py
```diff
@@ -0,0 +1,255 @@
...
+    def __init__(self, execution_stage=None, **kwargs):
+        if execution_stage in ["build", "union"]:
+            output_types = [OutputType.object]
+        else:
+            output_types = [OutputType.dataframe]
+        kwargs["_output_types"] = output_types
+        super().__init__(execution_stage=execution_stage, **kwargs)
+
+    @classmethod
+    def _get_value(cls, value):
```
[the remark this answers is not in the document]
qinxuye: Still no type annotation, so does the other functions like `_filter_on_index` etc.

mars/dataframe/base/bloom_filter.py
```diff
@@ -0,0 +1,257 @@
...
+                            cls._convert_to_hashable_dtypes(in_data.dtypes)
+                        )
+                        filtered = in_data[converted_data[on].apply(row_func, axis=1)]
+                    ctx[op.outputs[0].key] = filtered
+
+        else:  # pragma: no cover
+            raise ValueError(f"Unknown execution stage: {op.execution_stage}")
+
+
+def filter_by_bloom_filter(
```
qinxuye, line 181: Maybe this function could be tileable level, and the chunks related stuff could be moved to the `tile` function, I guess this function could be exposed to user some day perhaps, not just be taken as a utility function.

mars/dataframe/merge/merge.py
```diff
@@ -245,6 +255,32 @@ def _gen_shuffle_chunks(
             )
         return reduce_chunks
 
+    @classmethod
+    def _gen_bloom_filter_chunks(cls, left, right, left_on, right_on, op):
```
qinxuye, line 259: Type annotation, and seems like this function returns two tileables instead of `chunks` in the function name.

qinxuye (commented):
All new functions should be typed annotated, besides, I left some other comments.

qinxuye (commented):
All new functions should be typed annotated, besides, I left some other comments.

mars/dataframe/base/bloom_filter.py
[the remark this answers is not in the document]
hekaisheng: value could be a scalar like string, int, float, etc. I have added comments about the value.

qinxuye (approved):
LGTM

wjsi (approved):
LGTM

Issue #2916, opened 2022-04-13
[WIP]Proposal - Large scale Shuffle For Mars


chaokunyang:
# Background
Shuffle is performance critical part of distributed systems. In large scale cluster, shuffle will be bottle-neck for scaling, and the cause of cluster inefficiency. Current mars shuffle has scheduling and execution performance inefficiency which needs to be addressed:
- Scheduling cost
  - Serialization cost is very high. A mapper subtask contains thounds of chunks which is dependencies of `ShuffleProxy` if there are thounds of mapper chunks; A reducer subtask contains three groups of values which is part of `FetchShuffle` if there are thounds of mapper chunks. For large shuffle which has thounds of chunks, serializing a subtask will take about 1~20S which is obviously unacceptable. 
  - Scheduling gap. Mars will start schedule reducer subtasks only all mapper subtasks finish. For large scale cluster/shuffle, node jitter and mapper straggler is very common, the reducer will not be scheduled though mostly mapper subtasks has been finished.
- IO cost
  - For large scale shuffle, the shuffle block will be small, thus IO cost and meta stoe cost will be high. Actually sometimes the block be less than 10kb. In the end, there be too much small objects slow down the whole system.

# Optimization
The optimization can be splitted into following steps:
## Optimize mars serialization for shuffle
* We need do reduce the data needing sent to the workers
* Avoid duplicate serialization for `ShuffleProxy` and `FetchShuffle`
* Speed up serialization performance.
## Broadcast Join
If one of the table is pretty small, we can broadcast the smaller table to all mars workers, then execute the mapper join. In this way, there will be no shuffle. 
## Sort-Merge Join
For large scale shuffle, if we sort data in mapper, then all the things that reducer need to do is juse merge all mappers data. The reducer will be faster. And it there are data skew, hash-join may get OOM too.
##  Push-based Shuffle
![image](https://user-images.githubusercontent.com/12445254/163120088-dd0dfb59-6c1c-4b7c-85fb-d44c544f9902.png)
Push-based Shuffle has been proved the art of the state for shuffle algrithms. It's proposed by LinkedIn and spark3.2 has used [push-based  shuffle](https://issues.apache.org/jira/browse/SPARK-30602) to improve shuffle efficiency. Mars should use Push-based Shuffle too. 

Implement push-based shuffle useing distributed future provided by ray is easy, only need less 300 lines code. But mars doesn't provided such abstraction and mars shuffle is based on chunk abstraction.  So we propose to two solutions:
[...]

PR #2914, opened 2022-04-12
Mars serialization performance optimization

chaokunyang:
## What do these changes do?
This PR try to optimize mars serialization performance. In a large scale setup, mars serializaion took almost all cpu times of the supervisor and the computation can't proceed.

The optimization contains four parts:
*  Cache `FetchShuffle` serialization since itt data is large and  will be serialized thounds of times and sent to different workers.
* Optimize `Serializable` serialization by separate pickleable fields from other fields
* Optimize collection serialization by embedding basice types in header.
* Reduce header sizes

Before this PR, serialization for subtask/serializable/list/tuple/dict took:
```
6.770135926
11.172584438
1.8735732240000011
1.8050089589999985
3.7955724459999978
```

After this PR, it only took:
```
2.470553245999999
3.2098875259999993
0.2601947710000001
0.25069336799999853
0.5507458110000005
```

## Related issue number

#2916
Fixes #2907

## Check code requirements

- [ ] tests added / passed (if needed)
- [ ] Ensure all linting tests pass, see [here](https://docs.pymars.org/en/latest/development/contributing.html#check-code-styles) for how to run them

PR #2920, opened 2022-04-15
Optimize message serialization

chaokunyang:
## What do these changes do?
Optimize message serialization

## Related issue number

#2914
#2916
#2907

## Check code requirements


wjsi:
As nums of message types are relatively small, while construction of messages also requires acceleration, I'd like to rewrite the whole module with cython instead of playing with slot tricks.

PR #2922, opened 2022-04-15
Skip adding band in meta when fetch shuffle data

hekaisheng:
## What do these changes do?

Skip adding band in meta when fetch shuffle data as shuffle produce fragmented data.

## Related issue number

Fixes #xxxx

## Check code requirements


hekaisheng:
Actually we can only remove mapper data fetched from remote workers, if they are local, it means that the mapper was executed in same worker, they are tracked by lifecycle service and may be used by other subtask, thus they can't be deleted. So I moved deletions from `SubtaskProcessor` to `SubtaskExecutionActor`, return the remote mapper keys from `storage.fetch` and remove them after execution.

qinxuye (approved):
LGTM

wjsi (approved):
LGTM

Issue #2791, opened 2022-03-07
Improve performance of building expr, subtask graph etc


qinxuye:
# Motivation

Now, when tileables have large number of chunks, building expr, subtask graph and so forth could become the bottleneck, we need to try to reduce the overhead.

# Solutions

## Support generating dtypes, index_value, columns_value lazily for DataFrame chunks

Now, generating meta of DataFrame chunks will occupy quite a non-negligible, since the tileable have some important meta like dtypes, nsplits, index_value etc, we can actually calculate the chunk's meta, we don't have to generate the chunk meta intermediately unless we need them.

So we can add support for generating those meta lazily unless they are used. This job is POC in #2756 .

## Stop including dtypes, index_value etc in chunks of subtasks

When subtask graph is generated, meta of chunks will be include, we can skip those parts in subtask graph, hence the cost of serialization will be expected to reduce. However, for now, not a large number of ops are using these meta in their execution process, so we may need some work to correct them.

## Reduce cost of updating chunk meta

Updating chunk meta may be another bottleneck that can be reduced.

# Item Sheet

- [ ] Support generating dtypes, index_value, columns_value lazily for DataFrame chunks(#2756)
- [ ] Stop including dtypes, index_value etc in chunks of subtasks
- [ ] Reduce cost of updating chunk meta

PR #2923, opened 2022-04-16
Stop including `dtypes` etc in chunk graph of subtasks to reduce serialization

qinxuye:
## What do these changes do?

This PR tries to skip including `dtypes` etc in chunk graph of subtasks to reduce serialization.

## Related issue number

#2791 .

## Check code requirements


qinxuye:
`dtypes` etc is used in many ops, it would be a really effort to truly remove them from execution, and it's not very straightforward as well, so close this PR.

PR #2929, opened 2022-04-18
Ignore fetch operands when assign initial nodes

hekaisheng:
## What do these changes do?

When trigger iterative tiling, subtask's graph may has fetch operand, and we treat them as initial nodes and assign them in `GraphAnalyzer`, it will result in losing locality and bad performance. This PR try to solve this issue by ignoring fetch chunks when assign chunk graph.

## Related issue number

Fixes #xxxx

## Check code requirements

### DIFF
--- a/mars/services/task/execution/api.py
+++ b/mars/services/task/execution/api.py
@@ -68,10 +68,6 @@ class TaskExecutor(ABC):
         """Get available band slots."""
 
     @abstractmethod
-    async def get_chunk_bands(self, chunk_keys: List[str]) -> Dict[str, BandType]:
-        """Get bands that store these chunks"""
-
-    @abstractmethod
     async def get_progress(self) -> float:
         """Get the execution progress."""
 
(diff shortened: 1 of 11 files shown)


mars/services/task/execution/api.py
```diff
@@ -67,6 +67,10 @@ async def __aexit__(self, exc_type, exc_val, exc_tb):
     async def get_available_band_slots(self) -> Dict[BandType, int]:
         """Get available band slots."""
 
+    @abstractmethod
+    async def get_chunk_bands(self, chunk_keys: List[str]) -> Dict[str, BandType]:
```
qinxuye, line 71: About if it is reasonable to add this API to execution, what's your opinion? @fyrestone

fyrestone: The API is a bit wired for execution. From the title of the PR, I think may be the changes are related to graph construction not the execution?

fyrestone: Can you use the chunk to bands info for the graph assigner? https://github.com/mars-project/mars/blob/master/mars/services/task/supervisor/processor.py#L244

qinxuye (approved):
LGTM

wjsi (approved):
LGTM

PR #2949, opened 2022-04-22
Remove `should_be_monotonic` property

### LINKED ISSUE #2939: Remove `should_be_monotonic` property in `IndexValue`
Now `IndexValue`'s  should_be_monotonic is used to sort data after fetching from cluster when it is `True`, it's quite tricky and has some side effects. We need to remove it and accept different behavior with Pandas now, to keep with Pandas, we should implement it in distributed manner in the future.

hekaisheng:
## What do these changes do?

Remove `should_be_monotonic` property in `IndexValue`, it may cause incompatible results for some operands.

## Related issue number

Fixes #2939 

## Check code requirements


qinxuye (approved):
LGTM

wjsi (approved):
LGTM

PR #2957, opened 2022-04-24
[Ray] optimize ray fetcher by query in remote node

### LINKED ISSUE #2956: Submit query condition to remote node instead of fetch to local then query
Curently if ray fetcher gets objects with condition, it will fetch objects to local, then filter local objects with conditions. This will incur much objects transfer cost and high memory footsprint in local node. 

We should submit query to objects' node then fetch queried result instead.

chaokunyang:
## Related issue number

Fixes #2956 

## Check code requirements

### DIFF
--- a/mars/services/task/execution/ray/fetcher.py
+++ b/mars/services/task/execution/ray/fetcher.py
@@ -15,9 +15,10 @@
 import asyncio
 from collections import namedtuple
 from typing import Dict, List
+from .....utils import lazy_import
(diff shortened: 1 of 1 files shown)


fyrestone:
LGTM. It schedules the condition task to the object location to get the filtered result.

fyrestone (approved):

qinxuye (approved):
LGTM

PR #2959, opened 2022-04-25
Support `sort=True` for Groupby

### LINKED ISSUE #2915: Proposal - Group By Sort
The current implementation of the group by aggregation API in Mars uses the shuffle implementation for large dataframes. In the shuffle algorithm, the final order of the grouped and aggregated dataframe is not guaranteed. 

In this proposal we attempt to use the Parallel Sorting by Regular Sampling algorithm [PSRS] for sorting the group by result. 

### Pandas API
This sort feature is important to provide consistency with the Pandas API. In Pandas, the group by API provides the sort flag which can be used to sort the resulting output. 

> sort : bool, default True
> Sort group keys.  Get better performance by turning this off.  Note this does not influence the order of observations within each group.  Groupby preserves the order of rows within each group.

## Groupby Sort
For the groupby sort support, we propose leveraging the PSRS Algorithm to sort the dataframe in a distributed manner. We try to superimpose the different phases in the PSRS algorithm with the existing group by algorithm in Mars. We then add any missing components to facilitate the PSRS algorithm. 
Below we try to describe each Phase of the PSRS algorithm for groupby in detail.
- PHASE 1: In Phase 1, we sort each aggregated dataframe and pick n samples per mapper where n is the number of reducers (chunks).
- PHASE 2:  In Phase 2, we introduce a new operand marked as "Sample and Pivot" that takes in the samples generated in the map stage in Phase 1. We then sort the samples (and remove duplicates if necessary) It then generates the pivot elements depending on the number of reducers. So for n reducers, we pick n-1 pivot elements distributed across the samples.
- PHASE 3: In Phase 3 we have each Map assign reducer key using the pivot information (generated in Phase 2) instead of using hashes. This ensures that all groups G such that pivot_i < G < pivot_(i+1) are mapped to reducer_i. For reducer key, we will need to leverage the existing indexes for each chunk. I.e. Reducer with reducer key reducer_i will be the reducer with index i.
- PHASE 4: In Phase 4 we use the existing proxy shuffle method to have each reducer take keys belonging to this reducer. Then in the aggregation phase, the final dataframe is also sorted to ensure sorting within each chunk. As the final chunks are then sorted by index, we get the final dataframe in a sorted manner by groups.

Flowchart (2)

## Assumptions
- Distribution: Although the initial chunk data can be assumed to a normal distribution, we cannot assume the data to be uniformly distributed after the individual groupby aggregates on each chunk. The PSRS Algorithm however claims to achieve good load balancing even when the data is arbitrarily distributed.
- Duplicates: For a dataframe with n elements split across p chunks, we may have duplicates in the data. However, given the nature of group by, we can assume that no chunk will have any duplicates. It is only during Phase 2, where we gather regular samples from each chunk, can we get duplicates.

sak2002:
## Related issue number

Fixes #2915

## Check code requirements


mars/dataframe/groupby/aggregation.py
```diff
@@ -432,9 +619,15 @@ def _perform_shuffle(
         in_df: TileableType,
         out_df: TileableType,
         func_infos: ReductionSteps,
+        pivot_chunk,
```
qinxuye, line 622: Could you add a type hint here, it should be a `ChunkType`.

mars/dataframe/groupby/aggregation.py
```diff
@@ -293,6 +298,102 @@ def __call__(self, groupby):
         else:
             return self._call_series(groupby, df)
 
+    @classmethod
+    def partition_merge_data(cls, op, partition_chunks, proxy_chunk, in_df):
```
qinxuye, line 302: Could you add type hints for arguments? Only add hints for the new functions that you add. Due to historical reason,  some functions do not have type hints, but we hope that all the new functions should have.

qinxuye: So do the other new functions.

mars/dataframe/groupby/aggregation.py
```diff
@@ -432,9 +619,15 @@ def _perform_shuffle(
         in_df: TileableType,
         out_df: TileableType,
         func_infos: ReductionSteps,
+        pivot_chunk,
     ):
         # Shuffle the aggregation chunk.
-        reduce_chunks = cls._gen_shuffle_chunks(op, in_df, agg_chunks)
+        if op.groupby_params["sort"] and pivot_chunk is not None:
```
qinxuye, line 625: Maybe check pivot_chunk is OK? line 824 gens pivot_chunk only when some conditions meet.

mars/dataframe/groupby/sort.py
```diff
@@ -0,0 +1,184 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import numpy as np
+import pandas as pd
+
+from ..operands import DataFrameOperandMixin
```
qinxuye, line 18: I keep the order of import as from outer to inner. like

```
from .... import xx
from ... import xx
from .. import xx
```

mars/dataframe/groupby/sort.py
```diff
@@ -0,0 +1,184 @@
...
+            assert out.shape[1] == 1
+            out = out.iloc[:, 0]
+        ctx[op.outputs[-1].key] = out
+
+
+class DataFrameGroupbySortShuffle(MapReduceOperand, DataFrameOperandMixin):
+    _op_type_ = OperandDef.GROUPBY_SORT_SHUFFLE
+
+    # for shuffle map
+    _by = ListField("by")
```
qinxuye, line 110:
```suggestion
    by = ListField("by")
```

Now we prefer using public property instead of private one, and the `@property` stuffs are not required any more.

qinxuye (commented):
Really nice work, LGTM overall, some comments are mainly about type hints and code style.

mars/dataframe/groupby/aggregation.py
sak2002: Can you elaborate on this point? 
pivot chunk is None by default. If `op.groupby_params["sort"] and len(in_df.chunks) > 1` only then pivot chunk is not None

mars/dataframe/groupby/aggregation.py
sak2002: Fixed @qinxuye. Please confirm.

mars/dataframe/groupby/aggregation.py
sak2002: Fixed @qinxuye. Please confirm.

mars/dataframe/groupby/sort.py
sak2002: Fixed @qinxuye. Please confirm.

mars/dataframe/groupby/sort.py
sak2002: Fixed @qinxuye. Please confirm.

mars/dataframe/groupby/aggregation.py
qinxuye: Yeah, I mean we only need check `if pivot_chunk is not None:`, because if it's not None, `op.groupby_params["sort"]` is definitely True.

sak2002: understood

mars/dataframe/groupby/sort.py
```diff
@@ -0,0 +1,173 @@
...
+        ctx[op.outputs[-1].key] = out
+
+
+class DataFrameGroupbySortShuffle(MapReduceOperand, DataFrameOperandMixin):
+    _op_type_ = OperandDef.GROUPBY_SORT_SHUFFLE
+
+    # for shuffle map
+    by = ListField("by")
+    n_partition = Int32Field("n_partition")
+    output_types = ListField(
```
qinxuye, line 113: `_output_types` is defined already, so this property is not needed. you can redirect the setting in init.

```
def __init__(self, *args, output_types=None, **kw):
    super().__init__(self, *args, _output_types=output_types, **kw)
```

hekaisheng (approved):
LGTM, really a great work.

wjsi (approved):
LGTM

qinxuye (approved):
LGTM

qinxuye:
Really nice work.

Issue #2787, opened 2022-03-04
[Proposal] Expand slot management to resource management


zhongchun:
## Motivation

Currently Mars use slot for resource management and bands allocation which just consider cpu/gpu but no memory. Mars always allocate one slot which represents one core cpu or gpu for a subtask. It works well most time. But there are some shortcomings like:

* Subtasks need less cpu but assigned more which results in low cpu utilization and long execution time
* Subtasks need more memory and less cpu which leads node OOM

So we could develop more granular resource management and allocation to increase resource utilization, improve scheduling efficiency, and avoid OOM.

## Design

We propose a more common resource management which includes not only cpu/gpu but also memory, and even estimated time of a subtask.
A subtask of Mars needs one slot but no other resource by default. We could add more different types of resources to management.
Obviusly we can involve memory first as follows:

```
class Resource:
    num_cpus: float
    num_gpus: float
    num_mem_bytes: float
```

With this we can expand slot management to resource management. And bands allocation needs to consider both cpu/gpu and memory.
So we should develop a more complex resource management from a simple resource(cpu/gpu) to multiple resources.

In addition, we can easily implement hbo if we have an external system which can recommend resources for subtasks by history information.
If no external system, we can set memory resource to 0 which degenerates to the original slot scheduler or set a value through configuration to avoid OOM.

And later we can estimated execution time of subtasks if the external HBO system can recommend subtask execution time.

## Plan

In order to implement this proposal, we plan to do:
* Add physical resource management which has been in #2731 
* Add a logic id for subtask which represents a unique subtask and in different submits the same subtask has same logic id which has been in #2575 
* Add a logic id for tileable graph which just like subtask logic id and this is for HBO
* Introduce resource management and bands allocation #2825

wjsi:
As slots and CPU / memory combination are both resources, if we unify slot and resource strategy (for instance, always specify mem requirement as 0), do we really need strategy itself?

zhongchun:
> As slots and CPU / memory combination are both resources, if we unify slot and resource strategy (for instance, always specify mem requirement as 0), do we really need strategy itself?

Agree with you, we can unify them. I'll optimize the workflow.

PR #2961, opened 2022-04-25
Add logic key for tileable graph

zhongchun:
## What do these changes do?
 Logic key is a unique and deterministic key for `TileableGraph`. For multiple runs the logic key will remain same if the computational logic doesn't change. And it can be used to some optimization when running a same `execute`, like HBO.

## Related issue number
#2787 

## Check code requirements

### DIFF
--- a/mars/core/graph/entity.py
+++ b/mars/core/graph/entity.py
@@ -12,8 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import logging
-
 from abc import ABCMeta, abstractmethod
 from typing import List, Dict, Tuple, Union, Iterable
 
@@ -23,8 +21,6 @@ from ...serialization.serializables import Serializable, DictField, ListField, B
 from ...serialization.serializables.core import SerializableSerializer
 from ...utils import tokenize
 from .core import DAG
-
-logger = logging.getLogger(__name__)
 
 
 class EntityGraph(DAG, metaclass=ABCMeta):
@@ -88,14 +84,13 @@ class TileableGraph(EntityGraph, Iterable[Tileable]):
     def logic_key(self):
         if not hasattr(self, "_logic_key") or self._logic_key is None:
             token_keys = []
-            for node in self.topological_iter():
+            for node in self.bfs():
                 token_keys.append(
                     tokenize(node.op.get_logic_key(), **node.extra_params)
+                    if node.extra_params
+                    else node.op.get_logic_key()
                 )
-            self._logic_key = tokenize(*sorted(token_keys))
-            logger.debug(
-                "Got logic key: %s, token_keys: %s", self._logic_key, token_keys
-            )
+            self._logic_key = tokenize(*token_keys)
         return self._logic_key
 
 
--- a/mars/core/graph/entity.py
+++ b/mars/core/graph/entity.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
+
 from abc import ABCMeta, abstractmethod
 from typing import List, Dict, Tuple, Union, Iterable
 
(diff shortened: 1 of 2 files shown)


mars/core/graph/entity.py
```diff
@@ -74,6 +84,20 @@ def results(self):
...
+    def logic_key(self):
+        if not hasattr(self, "_logic_key") or self._logic_key is None:
+            token_keys = []
+            for node in self.topological_iter():
+                token_keys.append(
+                    tokenize(node.op.get_logic_key(), **node.extra_params)
+                )
+            self._logic_key = tokenize(*sorted(token_keys))
+            logger.debug(
+                "Got logic key: %s, token_keys: %s", self._logic_key, token_keys
```
wjsi, line 97: May just need to dump keys for tasks. Keys for all nodes bring too much information which may not be helpful for debug.

mars/core/graph/entity.py
```diff
@@ -74,6 +84,20 @@ def results(self):
     def results(self, new_results):
         self._result_tileables = new_results
 
+    @property
+    def logic_key(self):
+        if not hasattr(self, "_logic_key") or self._logic_key is None:
+            token_keys = []
+            for node in self.topological_iter():
```
wjsi, line 91: `topological_iter` does not guarantee order for nodes.

mars/core/graph/entity.py
```diff
@@ -74,6 +84,20 @@ def results(self):
...
+                token_keys.append(
+                    tokenize(node.op.get_logic_key(), **node.extra_params)
```
wjsi, line 93: As keys already determinded for nodes, is tehre any extra information brought by extra_params? Remember that `tokenize` is an expensive call.

mars/core/graph/entity.py
zhongchun: Year, so sort the `token_keys` in line 93.

mars/core/graph/entity.py
zhongchun: Thanks for your reminder. There are `raw_chunk_size` and other info in extra_params which may make the chunk graph is different. This logic key is used when communicating with external service and its cost is comparable to the time consumed by requesting external service.

wjsi: I would like `tokenize` not performend when there is no `extra_params`.

mars/core/graph/entity.py
wjsi: Then `topological_iter` can even be replaced with a BFS to save time.

mars/core/graph/entity.py
zhongchun: Good suggestion.

mars/core/graph/entity.py
zhongchun: Good suggestion. Thanks a lot.

mars/core/graph/tests/test_graph.py
```diff
@@ -105,3 +108,75 @@ def test_to_dot():
...
     dot = to_str(graph.to_dot(trunc_key=5))
     assert all(to_str(n.key)[:5] in dot for n in graph) is True
+
+
+def test_tileable_graph_logic_key():
+    # Tensor
+    t1 = mt.random.randint(10, size=(10, 8), chunk_size=4)
+    t2 = mt.random.randint(10, size=(10, 8), chunk_size=5)
+    graph1 = (t1 + t2).build_graph(tile=False)
+    graph2 = (t1 + t2).build_graph(tile=False)
```
fyrestone, line 118: Is the logic key remains the same even the tileable instances are different?
```python
graph1 = (t1 + t2).build_graph(tile=False)
tt1 = mt.random.randint(10, size=(10, 8), chunk_size=4)
tt2 = mt.random.randint(10, size=(10, 8), chunk_size=5)
graph2 = (tt1 + tt2).build_graph(tile=False)
assert graph1.logic_key == graph2.logic_key
```

zhongchun: Yes. Your writing is better, I'll optimize this.

fyrestone: Good to know this.

wjsi (approved):
LGTM

qinxuye (approved):
LGTM

Issue #2893, opened 2022-04-01
[PROPOSAL] The Execution API


fyrestone:
# The Execution API

## Background

Mars provides some interfaces for third party backends, but it is difficult to implement them and can not take the advantages of third party engine. For example, the Ray backend has to implement the Mars actor pool for Ray, the channel for Ray, the storage backend for Ray. However, the Mars on Ray can not use the great features of Ray: data-locality scheduling, data prefetching, distributed object management, ...

This proposal introduces an Execution API to split the graph construction and execution, it gives great freedom for the third party execution backends. The execution backend is able to choose the best way to execute each subtask graph and does not need to care about how the graph is constructed or optimized.

## Execution API

A classic execution logic of Mars is mixing tiling and execution, that is the iterative tiling. So, the execution API should handle this case.
![image](https://user-images.githubusercontent.com/6308809/161222445-31a7ae99-154a-4d23-98ae-d334cf561e1d.png)

Then, we can get the Execution API.

```python
from abc import ABC, abstractmethod
from dataclasses import dataclass
from typing import List, Dict, Any, Type

from ..core import ChunkGraph
from ..typing import BandType
from ..services.task import Task
from ..services.subtask import SubtaskGraph


@dataclass
class ExecutionChunkResult:
    key: str  # The chunk key for fetching the result.
    meta: Dict  # The chunk meta for iterative tiling.
    context: Any  # The context info, e.g. ray.ObjectRef.


class TaskExecutor(ABC):
    name = None

    @classmethod
    @abstractmethod
    async def create(
        cls, config: Dict, *, session_id: str, address: str, task: Task, **kwargs
    ) -> "TaskExecutor":
        """Create TaskExecutor instance from config.
        
        An example of config:
        {
            "backend": "mars",
            "mars": {...},
            "ray": {...},
        }
        """

    async def __aenter__(self):
        """Called when begin to execute the task."""

    @abstractmethod
    async def execute_subtask_graph(
        self,
        stage_id: str,
        subtask_graph: SubtaskGraph,
        chunk_graph: ChunkGraph,
        context: Any = None,
    ) -> List[ExecutionChunkResult]:
        """Execute a subtask graph and returns result."""

    async def __aexit__(self, exc_type, exc_val, exc_tb):
        """Called when finish the task."""

    @abstractmethod
    async def get_available_band_slots(self) -> Dict[BandType, int]:
        """Get available band slots."""

    @abstractmethod
    async def get_progress(self) -> float:
[...]

PR #2968, opened 2022-04-26
New ExecutionConfig API

fyrestone:
## What do these changes do?

This PR introduces the `ExecutionConfig` API to execution. With the `ExecutionConfig`, we can pass the cluster params to the execution backend.

- [x] New `ExecutionConfig`, the executor accepts an `ExecutionConfig` instance instead of a bare dict.
- [x] Rename the `task_executor_config` to `execution_config`.
- [x] Remove the `ClusterAPI` usage from Ray executor.
- [x] Move the logic of building band resources to `mars/services/task/execution/utils.py`.
- [x] When using Ray backend, the Mars cluster does not start the workers.

### How to add an API for the task processor (the parts other than execution) ?

The API can be put into the `ExecutionConfig`, some backends can have a different implementation.

### How to add an API for Ray executor (only used in the execution) ?

First, we can put the default value to the `config.yml`:

```yml
task:
  execution_config:
    backend: mars
    ray:
        ray_executor_specific_option: 5
```

Then, we can add an API `get_specific_opiton()` to the `RayExecutionConfig` to read the value of `ray_executor_specific_option` from the `self._execution_config`:

```json
{
  "backend": "ray",
  "ray": {
    "ray_executor_specific_option": 5
  }
}
```

## Related issue number

## Check code requirements

### DIFF
--- a/mars/deploy/oscar/tests/test_local.py
+++ b/mars/deploy/oscar/tests/test_local.py
@@ -275,6 +275,20 @@ async def test_execute(create_cluster, config):
 
     del a, b
 
+    if (
+        not isinstance(session._isolated_session, _IsolatedWebSession)
+        and session.client
+    ):
+        worker_pools = session.client._cluster._worker_pools
+        await session.destroy()
+        for worker_pool in worker_pools:
+            if hasattr(worker_pool, "external_address"):
+                _assert_storage_cleaned(
+                    session.session_id,
+                    worker_pool.external_address,
+                    StorageLevel.MEMORY,
+                )
+
 
 @pytest.mark.asyncio
 async def test_iterative_tiling(create_cluster):
@@ -299,6 +313,20 @@ async def test_iterative_tiling(create_cluster):
     assert df2.index_value.min_val >= 1
     assert df2.index_value.max_val <= 30
 
+    if (
+        not isinstance(session._isolated_session, _IsolatedWebSession)
+        and session.client
+    ):
+        worker_pools = session.client._cluster._worker_pools
+        await session.destroy()
+        for worker_pool in worker_pools:
+            if hasattr(worker_pool, "external_address"):
+                _assert_storage_cleaned(
+                    session.session_id,
+                    worker_pool.external_address,
+                    StorageLevel.MEMORY,
+                )
+
 
 @pytest.mark.asyncio
 async def test_execute_describe(create_cluster):
@@ -315,6 +343,20 @@ async def test_execute_describe(create_cluster):
     assert info.progress() == 1
     res = await session.fetch(r)
     pd.testing.assert_frame_equal(res, raw.describe())
+
+    if (
+        not isinstance(session._isolated_session, _IsolatedWebSession)
+        and session.client
+    ):
+        worker_pools = session.client._cluster._worker_pools
+        await session.destroy()
+        for worker_pool in worker_pools:
+            if hasattr(worker_pool, "external_address"):
+                _assert_storage_cleaned(
+                    session.session_id,
+                    worker_pool.external_address,
+                    StorageLevel.MEMORY,
+                )
 
 
 @pytest.mark.asyncio
@@ -439,6 +481,13 @@ async def test_web_session(create_cluster, config):
     AsyncSession.reset_default()
     await session.destroy()
     await _run_web_session_test(web_address)
+
+    worker_pools = client._cluster._worker_pools
+    for worker_pool in worker_pools:
+        if hasattr(worker_pool, "external_address"):
+            _assert_storage_cleaned(
+                session.session_id, worker_pool.external_address, StorageLevel.MEMORY
+            )
 
 
 @pytest.mark.parametrize("config", [{"backend": "mars", "incremental_index": True}])
@@ -590,6 +639,26 @@ def test_decref(setup_session):
     del df, df2
     ref_counts = session._get_ref_counts()
     assert len(ref_counts) == 0
+
+    with tempfile.TemporaryDirectory() as tempdir:
+        file_path = os.path.join(tempdir, "test.csv")
+        pdf = pd.DataFrame(
+            np.random.RandomState(0).rand(100, 10),
+            columns=[f"col{i}" for i in range(10)],
+        )
+        pdf.to_csv(file_path, index=False)
+
+        df = md.read_csv(file_path, chunk_bytes=os.stat(file_path).st_size / 5)
+        df2 = df.head(10)
+
+        result = df2.execute().fetch()
+        expected = pdf.head(10)
+        pd.testing.assert_frame_equal(result, expected)
+
+        del df, df2
+
+        ref_counts = session._get_ref_counts()
+        assert len(ref_counts) == 0
 
     worker_addr = session._session.client._cluster._worker_pools[0].external_address
     _assert_storage_cleaned(session.session_id, worker_addr, StorageLevel.MEMORY)
--- a/mars/deploy/oscar/tests/test_local.py
+++ b/mars/deploy/oscar/tests/test_local.py
@@ -139,7 +139,7 @@ def _assert_storage_cleaned(session_id: str, addr: str, level: StorageLevel):
 
 @pytest.mark.parametrize("backend", ["mars"])
 def test_new_session(backend):
-    from mars.services.task.execution.api import _name_to_config_cls
+    from ....services.task.execution.api import _name_to_config_cls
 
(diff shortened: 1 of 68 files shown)


mars/services/task/supervisor/manager.py
```diff
@@ -40,20 +40,20 @@ class TaskConfigurationActor(mo.Actor):
     def __init__(
         self,
         task_conf: Dict[str, Any],
-        task_executor_config: Dict[str, Any],
+        execution_config: Dict[str, Any],
```
qinxuye, line 43: Is the type right?

fyrestone: The type here is right. The `TaskExecutor` accepts the execution config in dict type and converts it to the correct `ExecutionConfig` object according to the backend. e.g. The `RayTaskExecutor` will be constructed with a `RayExecutionConfig`.

qinxuye (dismissed):
LGTM

hekaisheng (approved):
LGTM

chaokunyang:
How do I set default config which only ray backend need but mars doesn't with this API in rayconfig.yml

fyrestone:
> How do I set default config which only ray backend need but mars doesn't with this API in rayconfig.yml

The default config can be put in the `rayconfig.yml` like this,

```yml
task:
  execution_config:
    backend: mars
    ray:
        subtask_max_retries: 5
```
Then we can add some APIs in the `RayExecutionConfig` to read the `subtask_max_retries` value from `self._execution_config`:
```python
{
  "backend": "ray",
  "ray": {
    "subtask_max_retries": 5
  }
}
```

mars/deploy/oscar/tests/test_local.py
```diff
@@ -135,6 +138,48 @@ async def _assert(session_id: str, addr: str, level: StorageLevel):
...
+        return_deploy_band_resources = []
+
+        def _wrap_original_deploy_band_resources(*args, **kwargs):
+            nonlocal return_deploy_band_resources
+            return_deploy_band_resources = original_deploy_band_resources(
+                *args, **kwargs
+            )
+            return return_deploy_band_resources
+
+        config_init.side_effect = original_config_init
```
chaokunyang, line 163: Seems `config_init` is not mocked?

mars/services/task/execution/api.py
```diff
@@ -14,12 +14,124 @@
...
+            execution_config.get_execution_config(),
+        )
+        return self
+
+    @property
+    def backend(self) -> str:
+        """The backend from config."""
+        return self._execution_config["backend"]
+
+    def get_execution_config(self) -> Dict:
```
chaokunyang, line 74: Using a `property`?

mars/services/task/execution/api.py
```diff
@@ -14,12 +14,124 @@
...
+        """Get the execution config dict."""
+        return self._execution_config
+
+    @abstractmethod
+    def get_deploy_band_resources(self) -> List[Dict[str, Resource]]:
+        """Get the band resources for deployment."""
+
+    @classmethod
+    def from_config(cls, config: Dict, backend: str = None) -> "ExecutionConfig":
```
chaokunyang, line 83: This passes the whole mars config, do we really need this? Seems `from_execution_config` is enough.

mars/resource.py
```diff
@@ -391,3 +391,25 @@ def cuda_card_stats() -> List[_cuda_card_stat]:  # pragma: no cover
             )
         )
     return infos
+
+
+def build_band_resources(
```
chaokunyang, line 396:
```suggestion
def build_bands_resources(
```

mars/services/task/execution/core.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+from typing import List, Dict
+from ....resource import Resource, build_band_resources
+
+
+class _CommonPrivateConfigMixin:
+    """This class should ONLY provide the common private APIs for all backend."""
+
+    def _get_band_resources(self) -> List[Dict[str, Resource]]:
```
chaokunyang, line 23:
```suggestion
    def _get_bands_resources(self) -> List[Dict[str, Resource]]:
```

mars/services/task/execution/api.py
```diff
@@ -14,12 +14,124 @@
...
+    def backend(self) -> str:
+        """The backend from config."""
+        return self._execution_config["backend"]
+
+    def get_execution_config(self) -> Dict:
+        """Get the execution config dict."""
+        return self._execution_config
+
+    @abstractmethod
+    def get_deploy_band_resources(self) -> List[Dict[str, Resource]]:
```
chaokunyang, line 79:
```suggestion
    def get_deploy_bands_resources(self) -> List[Dict[str, Resource]]:
```

chaokunyang, line 79: What's the difference between `get_deploy_band_resources` and `get_band_resources`

mars/services/task/execution/ray/executor.py
```diff
@@ -236,9 +236,17 @@ async def __aexit__(self, exc_type, exc_val, exc_tb):
             await self._lifecycle_api.incref_tileables(tileable_keys)
 
     async def get_available_band_resources(self) -> Dict[BandType, Resource]:
-        async for bands in self._cluster_api.watch_all_bands():
-            if bands:
-                return bands
+        if self._available_band_resources is None:
```
chaokunyang, line 239:
```suggestion
        if self._available_bands_resources is None:
```

mars/services/task/execution/ray/executor.py
```diff
@@ -236,9 +236,17 @@ async def __aexit__(self, exc_type, exc_val, exc_tb):
...
+            band_resources = self._config.get_band_resources()
+            virtual_band_resources = {}
+            idx = 0
+            for band_resource in band_resources:
```
chaokunyang, line 243:
```suggestion
            for band_resource in bands_resources:
```

mars/services/task/execution/api.py
fyrestone: This is for merging the execution config part of the mars config: https://github.com/mars-project/mars/pull/2968/files#diff-1b98ea665f30f15485deb7982cdb0fc5e9127cb2225b9197811135ecbaefee73R156

With out this API, we have to access the execution config from dict by some keys, it is not maintainable.

mars/resource.py
fyrestone: It's a collection of (band, resource) pair, I think the `build_band_resources` is good enough. Currently, all the names are `band_resources` not `bands_resources`. @qinxuye @hekaisheng @wjsi What do you think?

mars/services/task/execution/api.py
fyrestone: The `get_deploy_band_resources` is for the deployment, the comments is here: https://github.com/mars-project/mars/pull/2968/files#diff-28df8dd40f8370801f54fe9d00b8f745446601d87f294d1be220d2afdc3c3acaR80

While, the `get_band_resources` just get the band resources from the config.

mars/resource.py
hekaisheng: I think `band_resources` is better for me.

chaokunyang (dismissed):
LGTM

mars/services/task/execution/api.py
fyrestone: I think that this API should not be used very often. So, it is a method not a property.

mars/resource.py
wjsi, line 396: `band` is a concept introduced in Cluster Service. Thus not appropriate to introduce it in `mars.resource`.

mars/services/task/execution/core.py
```diff
@@ -0,0 +1,31 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from typing import List, Dict
+from ....resource import Resource, build_band_resources
+
+
+class _CommonPrivateConfigMixin:
```
wjsi, line 20: A little strange for me as there is no reason to place a separate Mixin for all backends.

fyrestone: Fixed.

mars/resource.py
fyrestone: So, where is the best place for the `build_band_resources`?

fyrestone: Fixed. Thanks.

wjsi (approved):
LGTM

PR #2996, opened 2022-05-05
[Metrics] Optimize metric configs

zhongchun:
## What do these changes do?
The arguments  of `init_metrics(backend, port)` method are inappropriate. Because port is a configuration of prometheus. It is not necessary for `console` and `ray`. It's better to change it to `config` and in this way we could easily add new metric backend.

## Related issue number

Fixes #xxxx

## Check code requirements


qinxuye (approved):
LGTM

hekaisheng (approved):
LGTM

PR #2997, opened 2022-05-05
Expand the ability to resource evaluator

zhongchun:
## What do these changes do?
Resource evaluator is used to estimate and set resources required by subtasks. It can be an internal service or an external service. If it is an internal service, we can set default of adjustable resources for subtasks. If it is an external service, we should report the running result of the task to the external service, so that it can accurately predict the required resources of subtasks based on the historical running information, we call it HBO.
But it is not easy to implement a new resource evaluator and config it. This pr introduces an extension point of resource evaluator with which we could add a new evaluator as follows:

* Inherit `ResourceEvaluator` and implement `create`, `evaluate` and `report` methods. The `create` method is to create a new resource evaluator instance. The `evaluate` method is to estimate and set required resources for the subtasks of a task stage. And this method must be implemented. The `report` method is to report the running information and result of the task. And this method does not have to be implemented.

* Add default configs of the new evaluator needed in `base_config.xml` or its descendant files.

* Set the `resource_evaluator` to choose a resource evaluator in `base_config.xml` when running a mars job.

## Related issue number

Fixes #xxxx

## Check code requirements

### DIFF
--- a/mars/services/task/execution/mars/executor.py
+++ b/mars/services/task/execution/mars/executor.py
@@ -73,6 +73,7 @@ class MarsTaskExecutor(TaskExecutor):
         lifecycle_api: LifecycleAPI,
         scheduling_api: SchedulingAPI,
         meta_api: MetaAPI,
+        resource_evaluator,
     ):
         self._config = config
         self._task = task
@@ -94,6 +95,9 @@ class MarsTaskExecutor(TaskExecutor):
         self._subtask_decref_events = dict()
         self._meta_updated_tileables = set()
 
+        # Evaluate and initialize subtasks required resource.
+        self._resource_evaluator = resource_evaluator
+
     @classmethod
     async def create(
         cls,
@@ -111,6 +115,12 @@ class MarsTaskExecutor(TaskExecutor):
         cluster_api, lifecycle_api, scheduling_api, meta_api = await cls._get_apis(
             session_id, address
         )
+        resource_evaluator = await ResourceEvaluator.create(
+            config.get_execution_config(),
+            session_id=task.session_id,
+            task_id=task.task_id,
+            cluster_api=cluster_api,
+        )
         return cls(
             config,
             task,
@@ -119,6 +129,7 @@ class MarsTaskExecutor(TaskExecutor):
             lifecycle_api,
             scheduling_api,
             meta_api,
+            resource_evaluator,
         )
 
     @classmethod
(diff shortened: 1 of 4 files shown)


mars/services/task/execution/mars/executor.py
```diff
@@ -73,6 +73,7 @@ def __init__(
         lifecycle_api: LifecycleAPI,
         scheduling_api: SchedulingAPI,
         meta_api: MetaAPI,
+        resource_evaluator,
```
qinxuye, line 76: Can you add a type for `resource_evaluator`?

mars/services/task/execution/mars/resource.py
```diff
@@ -12,28 +12,89 @@
...
-                Resource(num_gpus=1) if is_gpu else Resource(num_cpus=1)
+    name = None
+
+    @classmethod
+    @abstractmethod
+    async def create(cls, config: Dict[str, Any], **kwargs) -> "ResourceEvaluator":
+        name = config.get("resource_evaluator", "default")
+        evaluator_config = config.get(name, {})
+        evaluator_cls = _name_to_resource_evaluator[name]
+        if evaluator_cls.create.__func__ is ResourceEvaluator.create.__func__:
```
qinxuye, line 73: Just make `ResourceEvaluator` as a abstract class? The abstractmethod has to be implemented then.

mars/services/task/execution/mars/executor.py
fyrestone, line 76: Can you add a type hint for `resource_evaluator`?

qinxuye (commented):
About resource evaluator, is there a plan about it? Apart from the plan, I left some comments.

mars/services/task/execution/mars/executor.py
zhongchun: Yes. Thanks for your reminder.

zhongchun: Thanks for your reminder.

mars/services/task/execution/mars/resource.py
zhongchun: Good suggestion. Thanks a lot.

zhongchun:
> 

It is necessary to build an external resource recommendation service that can take advantage of historical job information if we want to fully utilize resource evaluator. Maybe we could add an extension library of mars later.

zhongchun:
> About resource evaluator, is there a plan about it? Apart from the plan, I left some comments.

It is necessary to build an external resource recommendation service that can take advantage of historical job information if we want to fully utilize resource evaluator. Maybe we could add an extension library of mars later.

mars/services/task/execution/mars/resource.py
qinxuye: If `ResourceEvaluator` is a abstract class, is this check still necessary?

zhongchun: It's not necessary. I'll remove these code. Thanks for your reminder.

qinxuye (approved):
LGTM

Issue #2972, opened 2022-04-26
[Proposal] a lineage reconstruction based failover for mars


chaokunyang:
# Background
Large-scale distributed computing systems may fail due to various reasons, including network problems, machine failures, and process restarts. Network failures can cause nodes and workers to fail to send and receive data; machine failures and process restarts can cause data loss and tasks to be re-executed.

In Mars On Ray, failures fall into three main categories:
1. The ray objects (input parameters, etc.) that subtask depends on exist, and the execution of subtask itself fails (usually caused by OOM). In this case, subtask can be retried;
2. The ray objects (input parameters, etc.) that the subtask depends on are lost, but the lineage of those objects exist. In this case, the input parameters need to be recursively reconstructed through lineage reconstruction, and then the current subtask can be executed;
3. The ray objects (input parameters, etc.) that subtask depends on are lost, but the lineages of the objects are lost. In this case, the subtask cannot be recovered;

Here we propose how to implement Mars On Ray failover based on the distributed future provided by Ray:
1. Recover subtask based on task retry
2. Recover lost subtask results through lineage reconstruction
3. Cut off the lineage dependency chain through checkpoint and object management to avoid a large number of task reconstruction, and reduce the metadata storage overhead at the same time.

# Proposal
## Subtask retry
Most subtask execution failures can be directly recovered by task retry. Ray supports automatic rerunning of failed tasks. The default number of retries is 3. We can specify the number of retries through `max_retries` when submitting subtasks. Set to -1 to retry indefinitely; set to 0 to disable retry.
The key to task retry is that each subtask must be **idempotent and side-effect free**, otherwise there will be data and state inconsistencies. Therefore, for each subtask, if it is retryable, set the configured number of retries when submitting the remote task corresponding to the subtask; if it is not retryable, set `max_retries` to 0.

## Recover subtask through lineage reconstruction
When the input objects that the subtask depends on are lost (node ​​failure, object GC, etc.), the current subtask cannot be retried directly. All objects that the task depends on need to be recursively reconstructed before retrying the current subtask. Ray supports **automatic reconstruction of the lost objects through lineage reconstruction**, thereby recovering the failed subtask. When the task supports retry, the owner will cache the object lineage, that is, the task specification that needs to recreate the object. If all copies of the object are lost, the owner will resubmit the task that created the object, and the objects that the task depends on will be recursively reconstructed.
[...]

PR #3002, opened 2022-05-06
[Ray] Ray execution state

fyrestone:
## What do these changes do?

Mars execution context provides remote object to sync states between multiple running operands. Ray execution backends use a task state actor for this feature.

With this PR, ray backend supports incremental index.

- [x] Move the logic of initializing context _(the ThreadedServiceContext used in the supervisor)_ to the execution backend.
- [x] Add a task state actor to Ray execution backend.
- [x] Ray execution context supports remote object.
- [x] Not to fetch the chunk meta when tiling `HeadOptimizedDataSource`.

The drawback is that,

- Poor performance to sync states among operands, maybe we can avoid using remote object when executing operands in the future.
- Hard to fault tolerance because the remote object makes the operands stateful.

## Related issue number

## Check code requirements

### DIFF
--- a/mars/dataframe/datasource/core.py
+++ b/mars/dataframe/datasource/core.py
@@ -58,9 +58,7 @@ class HeadOptimizedDataSource(DataFrameOperand, DataFrameOperandMixin):
         # execute first chunk
         yield chunks[:1]
 
-        ctx = get_context()
-        chunk_shape = ctx.get_chunks_meta([chunks[0].key], fields=["shape"])[0]["shape"]
-
+        chunk_shape = chunks[0].shape
         if chunk_shape[0] == op.nrows:
             # the first chunk has enough data
             tileds[0]._nsplits = tuple((s,) for s in chunk_shape)
--- a/mars/services/task/execution/ray/context.py
+++ b/mars/services/task/execution/ray/context.py
@@ -14,6 +14,7 @@
 
 import functools
 import inspect
+from typing import Union
 
 from .....core.context import Context
 from .....utils import implements, lazy_import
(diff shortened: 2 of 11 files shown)


chaokunyang:
Will this ray actor influence the lineage reconstruction in #2972:
1. If this actor died, will lineage reconstruction still succeed?
2. If this actor restarted by ray, how the state auto recovered?
3. If the lineage reconstruction call this actor multiple times, how the idempotence be ensured?

fyrestone:
> Will this ray actor influence the lineage reconstruction in #2972:
> 
> 1. If this actor died, will lineage reconstruction still succeed?
> 2. If this actor restarted by ray, how the state auto recovered?
> 3. If the lineage reconstruction call this actor multiple times, how the idempotence be ensured?

Currently, some operands use a remote object to sync states. If the state actor is reconstructed, the simplest way to recover the compute is,

1. If call remote object can't find the remote object and the state actor is reconstructed, then raises a RecoveryFailed exception.
2. The fault recovery logic catch this exception to recompute the predecessors.

We should make the operands stateless to avoid above complex recovery.

This PR is to maximize compatibility with existing Mars execution logic, the fault recovery is not included in this PR.

qinxuye (approved):
LGTM

qinxuye:
@chaokunyang could you please review this PR?

mars/services/task/execution/ray/context.py
```diff
@@ -12,9 +12,108 @@
...
+
+    @implements(Context.destroy_remote_object)
+    def destroy_remote_object(self, name: str):
+        task_state_actor = self._get_task_state_actor()
+        task_state_actor.destroy_remote_object.remote(name)
+
+
+# TODO(fyrestone): Implement more APIs for Ray.
+class RayExecutionContext(_RayRemoteObjectContext, ThreadedServiceContext):
+    """The context for tiling."""
```
zhongchun, line 108: Is it possible not to inherit `ThreadedServiceContext` which has many properties and methods like `_subtask_api` that is not used in ray task executor?

mars/dataframe/datasource/core.py
```diff
@@ -58,9 +58,7 @@ def _tile_head(cls, op: "HeadOptimizedDataSource"):
         # execute first chunk
         yield chunks[:1]
 
-        ctx = get_context()
-        chunk_shape = ctx.get_chunks_meta([chunks[0].key], fields=["shape"])[0]["shape"]
-
+        chunk_shape = chunks[0].shape
```
zhongchun, line 61: I'm a little confused why it is possible to replace `ctx.get_chunks_meta` here.

zhongchun (commented):
LGTM. I left two comments where i am a little confused.

mars/services/task/execution/ray/context.py
fyrestone: There are many APIs in the `ThreadedServiceContext` are used by the task supervisor, e.g. `get_total_n_cpu`, `get_worker_addresses`, ... So, the `RayExecutionContext` inherits from `ThreadedServiceContext`.

Maybe we can split the `ThreadedServiceContext` into two classes in the future, one for tiling and the other for executing. But, the remote object management is shared by tiling and executing. In most cases, the `create_remote_object` is for tiling, while `get_remote_object` and `destroy_remote_object` are for executing.

mars/dataframe/datasource/core.py
fyrestone: Before getting the shape of chunk[0], the tiling of `HeadOptimizedDataSource` yield chunks[:1] so that the chunk[0] has been executed in a stage. The task processor updates the meta of stage output _(the chunk[0] in this case)_ at the end of stage, so we can get  the shape from the chunk instead of meta service.

zhongchun (approved):
LGTM

Issue #3039, opened 2022-05-17
Support get_chunk_meta in RayExecutionContext


chaokunyang:
Currently `RayExecutionContext.get_chunk_meta` is not supported, which will make any operands relied on this API failed on tiling, such as when call `DataFrame.groupby`:
```
df = md.DataFrame(mt.random.rand(300, 4, chunk_size=100), columns=list("abcd"))
df["a"], df["b"] = (df["a"] * 5).astype(int), (df["b"] * 2).astype(int)
df.groupby(["a", "b"]).apply(lambda pdf: pdf.sum()).execute()
```
Will got following error:
```
self = <mars.services.task.execution.ray.context.RayExecutionContext object at 0x7f7edab6a2e0>
data_keys = ['5ddaa3e5218721d55567a2fa991d31ce_0', '4647e59976f80420e69961a12ed7fe2d_0', '39697fd5b599391b86904e1d3cbd60b3_0'], fields = ['memory_size'], error = 'ignore'

    async def _get_chunks_meta(
        self, data_keys: List[str], fields: List[str] = None, error: str = "raise"
    ) -> List[Dict]:
        # get chunks meta
        get_metas = []
        for data_key in data_keys:
            meta = self._meta_api.get_chunk_meta.delay(
                data_key, fields=["bands"], error=error
            )
            get_metas.append(meta)
        metas = await self._meta_api.get_chunk_meta.batch(*get_metas)
        api_to_keys_calls = defaultdict(lambda: (list(), list()))
        for data_key, meta in zip(data_keys, metas):
>           addr = meta["bands"][0][0]
E           TypeError: 'NoneType' object is not subscriptable

mars/services/context.py:145: TypeError
```

We need to support get_chunk_meta for ray task backend.

PR #3040, opened 2022-05-17
[Ray] task based shuffle for ray

chaokunyang:
## What do these changes do?
This PR implements pull-based shuffle for ray task backend, which will be the basis for later push-based shuffle for ray. Main changes include:

- [x] Introduce `ShuffleFetchType` which contains `FETCH_BY_KEY`/`FETCH_BY_INDEX`. If `FETCH_BY_INDEX` is used, `source_keys/source_idxes/source_mappers` will be skipped. In this way the meta and scheduling cost for supervisor will be much smaller, about 1000~10000x smaller for big shuffle. This will make mars able to schedule large-scale shuffle, even PB-scale data shuffle.
- [x] add n_mappers/n_reducers to `FetchShuffle`, and skip `source_keys/source_idxes/source_mappers` when `shuffle_fetch_type` is `FETCH_BY_INDEX`. 
- [x] Operands compatibility: make mars operands compatible with `FETCH_BY_INDEX`. 
  - [x] Mappers may produce undefinite output nums: `TensorReshape`/`TensorBincount`
  - [x] Mappers may produce shuffle blocks greater than num_reducer: `PSRSAlign`.
- [x] Introduce a new subtask dependency resolve mechanism, which alse remove `data_keys` for subtask execution on ray
- [x] Add `ShuffleManager` for shuffle execution on ray
- [x] Skip recording shuffle operands meta in supervisor
- [x] Support all shuffle operands on ray

Before this PR, for a shuffle with 100000 partitions, every reducer subtask meta will took about 16M memory, which make mars to handle only less than 1000 partitions, i.e. less than 1 T data.

With this PR, every reducer subtask took only 2Kb memory, which make mars can handle 8000_000 partitions. If every partition has 1Gb data, this PR will enable mars handle 8PB data shuffle.

#2916
#3039
#2112

## Check code requirements


mars/deploy/oscar/local.py
```diff
@@ -152,7 +152,9 @@ def __init__(
         self._mem_bytes = mem_total() if mem_bytes == "auto" else mem_bytes
         self._cuda_devices = self._get_cuda_devices(cuda_devices, n_worker)
         self._subprocess_start_method = subprocess_start_method
-        self._config = load_config(config, default_config_file=DEFAULT_CONFIG_FILE)
+        self._config = load_config(
+            config, default_config_file=DEFAULT_CONFIG_FILE, backend=backend
+        )
```
fyrestone, line 157: We don't needs this because the execution config merging will update the backend.

mars/deploy/oscar/tests/test_ray_dag.py
```diff
@@ -122,3 +126,22 @@ def test_sync_execute(config):
 @pytest.mark.asyncio
 async def test_session_progress(ray_start_regular_shared2, create_cluster):
     test_local.test_session_progress(create_cluster)
+
+
+@require_ray
+@pytest.mark.asyncio
+async def test_shuffle(ray_start_regular_shared2, create_cluster):
+    df = md.DataFrame(mt.random.rand(300, 4, chunk_size=100), columns=list("abcd"))
```
fyrestone, line 134: Why not just run the `test_local.py::test_execute_describe`?

mars/deploy/utils.py
```diff
@@ -155,6 +157,8 @@ def load_config(config: Union[str, Dict], default_config_file: str):
     else:
         full_config = load_service_config_file(default_config_file)
         config = _merge_config(full_config, config)
+    if backend:
+        config["task"]["execution_config"]["backend"] = backend
```
fyrestone, line 161: Don't add the `backend` here, `backend` is for execution. You can add the `shuffle_type` to the execution config is it is related to the `backend`.

mars/services/task/execution/ray/executor.py
```diff
@@ -78,29 +81,72 @@ def execute_subtask(
...
+        reducer_operands = set(c.op for c in reducer_chunks)
+        assert len(reducer_operands) == 1, (
+            reducer_operands,
+            reducer_chunks,
+            subtask_digraph,
+        )
+        reducer_operand = reducer_chunks[0].op
+        reducer_index = reducer_operand.reducer_index
+        # mock input keys, keep this in sync with `MapReducerOperand#_iter_mapper_key_idx_pairs`
+        input_keys = [(i, reducer_index) for i in range(num_mappers)]
```
fyrestone, line 118: Any tests for the input keys that matched all the cases of shuffle operands?

mars/services/task/execution/ray/executor.py
```diff
@@ -78,29 +81,72 @@ def execute_subtask(
...
+    context = RayExecutionWorkerContext(
+        RayTaskState.gen_name(task_id), zip(input_keys, inputs)
+    )
+
     for chunk in subtask_chunk_graph.topological_iter():
         if chunk.key not in context:
-            execute(context, chunk.op)
+            try:
+                execute(context, chunk.op)
+            except Exception as e:  # pragma: no cover
```
fyrestone, line 129: Maybe we can add a decorator to the `execute_subtask` to log the exceptions.

mars/tensor/reshape/reshape.py
```diff
@@ -457,7 +458,17 @@ def _execute_map(cls, ctx, op):
                 t, target_chunk_idx[idx] = divmod(t, dim_chunk_count)
             target_chunk_idx.reverse()
 
-            ctx[chunk.key, tuple(target_chunk_idx)] = group_indices + (group_data,)
+            mapper_outputs[chunk.key, tuple(target_chunk_idx)] = group_indices + (
+                group_data,
+            )
+
+        # ensure all mapper data are inserted context in order and fill missing partition with None
```
fyrestone, line 465: This adds more costs to the Mars backend, can we move this to Ray executor?

mars/deploy/oscar/tests/test_ray_dag.py
chaokunyang: we added `df.groupby` in this tests case, `test_execute_describe` doesn't contains `groupby`.

mars/deploy/utils.py
chaokunyang: I think it's needed.
image
If we don't add this, and the execution backend is `ray`, `config["task"]["execution_config"]["backend"]` will still be `mars` which is not consistent.

mars/services/task/execution/ray/executor.py
chaokunyang: Yes, will add test cases to cover shuffle operands. we definitely need do this, otherwise the shuffle implementation will be easily broke

mars/tensor/reshape/reshape.py
chaokunyang: Actually the cost is minimal. Mars shuffle fetch will filter out all `None` values.

mars/services/task/execution/ray/executor.py
chaokunyang: I added a decorator at the first, but I removed it, cause I felt it kind of complex. 
This is my decorator definition:
```
def log_exception(func, *msg_args, ex_logger=None):
    import inspect
    if msg_args:
        msg_tpl, *msg_args = msg_args
    else:
        msg_tpl, msg_args = None, []
    if inspect.iscoroutinefunction(func):
        async def new_func(*args, **kwargs):
            try:
                return await func(*args, **kwargs)
            except Exception:
                (ex_logger or logger).exception(msg_tpl, *msg_args)
                raise
    else:
        def new_func(*args, **kwargs):
            try:
                return func(*args, **kwargs)
            except Exception:
                (ex_logger or logger).exception(msg_tpl, *msg_args)
                raise
    return new_func
```

Do we need add this?

mars/deploy/oscar/local.py
chaokunyang: It didn't update the config in `config["task"]["execution_config"]["backend"]`

mars/services/task/execution/ray/executor.py
chaokunyang: emm, `log_exception` seems reusable, will change to decorator

mars/services/task/execution/ray/executor.py
chaokunyang: There is no easy way for a unit test cases cover all shuffle operands. I added assert in `execute_subtask`. The runtime  cost is minimal, I think it's acceptable.

mars/deploy/oscar/tests/test_ray_dag.py
fyrestone: Then, just extends the `test_execute_describe` case or add a new case? We are trying to get both of the backends to run the same test cases.

mars/deploy/utils.py
fyrestone: Here updates the config, we can switch the backend by passing `backend="ray"` to the `new_session` and `new_test_session`.

image

mars/services/task/execution/ray/executor.py
fyrestone: Can you add some assertion to the `CheckedSubtaskProcessor` to make sure the map keys you generated is exactly the same with Mars original logic.

mars/deploy/utils.py
chaokunyang: Got it, I passed `shuffle_type` from `TaskProcessor` to `GraphAnalyzer` directly.

mars/deploy/oscar/tests/test_ray_dag.py
chaokunyang: Split it into `test_execute_describe` and `test_shuffle` test cases

mars/services/task/execution/ray/executor.py
chaokunyang: The keys will generated only when `PushShuffle` is used. Mars dones't build `PushShuffle` into subtask graph.

mars/tensor/reshape/reshape.py
chaokunyang: Yes, `None` values will be put into storage in Mars backend. But shuffle mapper meta are not recorded in mars, so the cost will be negligible？

zhongchun:
Is this implementation the solution proposed in MEP 0001?  Is `PushShuffle` the `FetchShuffleByIndex` mentioned in MEP 001?

chaokunyang:
> Is this implementation the solution proposed in MEP 0001? Is `PushShuffle` the `FetchShuffleByIndex` mentioned in MEP 001?

Yes, this is the implementation for  MEP 0001. I've removed `PushShuffle` and add the properties into `FetchShuffle`

mars/utils.py
```diff
@@ -544,34 +544,56 @@ def _is_fast_dtype(dtype):
...
+    assert isinstance(chunk_op, ShuffleProxy)
     params = chunk.params.copy()
+    n_mappers = len(chunk.inputs)
+    assert n_reducers > 0, n_reducers
+    # for shuffle nodes, we build FetchShuffle chunks
+    # to replace ShuffleProxy
+    source_keys, source_idxes, source_mappers = [], [], []
+    for pinp in chunk.inputs:
+        source_keys.append(pinp.key)
+        source_idxes.append(pinp.index)
```
fyrestone, line 563: Why not put these lines to `elif shuffle_fetch_type == ShuffleFetchType.FETCH_BY_KEY`? It seems that when `shuffle_fetch_type == ShuffleFetchType.FETCH_BY_INDEX`, the `source_keys, source_idxes, source_mappers` are None.

mars/utils.py
```diff
@@ -544,34 +544,56 @@ def _is_fast_dtype(dtype):
...
-            source_keys=source_keys,
-            source_idxes=source_idxes,
-            source_mappers=source_mappers,
-            gpu=chunk.op.gpu,
-        )
-    else:
-        # for non-shuffle nodes, we build Fetch chunks
-        # to replace original chunk
-        op = chunk_op.get_fetch_op_cls(chunk)(sparse=chunk.op.sparse, gpu=chunk.op.gpu)
+        return build_fetch_shuffle(chunk)
```
fyrestone, line 593: Is this branch `if isinstance(chunk_op, ShuffleProxy):` necessary? I noticed that the `ShuffleProxy` type is checked in the analyzer, if we return `build_fetch_shuffle(chunk)`, the `n_reducers` will be None, `assert n_reducers > 0, n_reducers` will fail.

mars/services/task/execution/ray/executor.py
```diff
@@ -118,29 +124,66 @@ def execute_subtask(
...
-    input_keys: List[str],
     *inputs,
 ):
-    logger.info("Begin to execute subtask: %s", subtask_id)
+    """The function used for execute subtask in ray task.
+    If subtask is shuffle mapper, no chunk meta will be returned, otherwise return chunk mata."""
     ensure_coverage()
     subtask_chunk_graph = deserialize(*subtask_chunk_graph)
-    # inputs = [i[1] for i in inputs]
+    subtask_digraph = subtask_chunk_graph.to_dot()
```
fyrestone, line 133: Can you dump the subtask graph dots if error occurred? We want the success execution branch as fast as possible.

mars/services/task/execution/ray/executor.py
```diff
@@ -519,26 +627,41 @@ async def cancel(self):
...
-                    key, fields=["object_refs"]
-                )
+        # for non-shuffle chunks, chunk key will be used for indexing object refs.
+        # for shuffle chunks, mapper subtasks will have only one mapper chunk, and all outputs for mapper
+        # subtask will be shuffle blocks, the downstream reducers will receive inputs in the mappers order.
+        start_chunks = list(subtask.chunk_graph.iter_indep())
+        for index, start_chunk in enumerate(start_chunks):
+            if isinstance(start_chunk.op, Fetch):
+                # don't skip `pure_depend_keys`, otherwise pure_depend subtask execution order can't be ensured,
+                # since those ray tasks don't has dependencies on other subtasks.
```
fyrestone, line 651: All the pure dependants are yield and executed when tiling (the iterative tiling), can you provide some cases that the execution is out of order?

mars/services/task/execution/ray/executor.py
```diff
@@ -118,29 +124,66 @@ def execute_subtask(
...
+            reducer_operands,
+            reducer_chunks,
+            subtask_digraph,
+        )
+        reducer_operand = reducer_chunks[0].op
+        reducer_index = reducer_operand.reducer_index
+        # mock input keys, keep this in sync with `MapReducerOperand#_iter_mapper_key_idx_pairs`
+        input_keys = [(i, reducer_index) for i in range(n_mappers)]
+    else:
+        input_keys = [c.key for c in start_chunks if isinstance(c.op, Fetch)]
```
fyrestone, line 161: The `iter_indep` order is guaranteed, please transfer the `input_keys` for non shuffle subtasks.

mars/utils.py
chaokunyang: Good catch, we should remove this and assert chunk.op is not ShuffleProxy

mars/services/task/execution/ray/executor.py
chaokunyang: I omitted `input_keys ` on purpose, so the meta for task spec will be small nad serialization cost is small too.

mars/utils.py
chaokunyang: Fixed

mars/services/task/execution/ray/executor.py
chaokunyang: The comments is stale, I'll remove it.

mars/services/task/execution/ray/executor.py
chaokunyang: Fixed

mars/services/task/execution/ray/executor.py
fyrestone: My point is the order of `iter_indep` is not guaranteed through serialize / deserialze. If the `input_keys` is omitted here, then we should add more checks to keep the order when serializing.

mars/services/task/execution/ray/executor.py
```diff
@@ -118,29 +124,64 @@ def execute_subtask(
...
-    # from data_key to results
+
     for chunk in subtask_chunk_graph.topological_iter():
         if chunk.key not in context:
-            execute(context, chunk.op)
+            wrapped_execute = log_exception_wrapper(
+                execute,
+                "Execute operand %s of graph %s failed.",
+                chunk.op,
+                subtask_chunk_graph.to_dot(),
```
fyrestone, line 170: `subtask_chunk_graph.to_dot()` every execute is very slow. We can pass a lambda instead of string formatter to the `log_exception_wrapper`.

fyrestone: Also, `log_exception_wrapper` every loop is very slow, too. Please wrap the entire loop or just wrap the `execute_subtask`.

mars/core/graph/core.pyx
```diff
@@ -339,28 +339,38 @@ cdef class DirectedGraph:
...
                         visited.add(op.key)
-                    sio.write(f'"{op_name}:{op.key[:trunc_key]}" -> "Chunk:{output_chunk.key[:5]}"\n')
+                    sio.write(f'"{op_name}:{op.key[:trunc_key]}" -> '
+                              f'"Chunk:{self._gen_chunk_key(output_chunk, trunc_key)}"\n')
 
         sio.write('}')
         return sio.getvalue()
 
+    @classmethod
+    def _gen_chunk_key(cls, chunk, trunc_key):
```
fyrestone, line 367: Is the `to_dot` changes are necessary for ray shuffle? If not, it's better to move them to another PR.

mars/tensor/random/core.py
```diff
@@ -313,8 +313,9 @@ def _on_serialize_random_state(rs):
 
 
 def _on_deserialize_random_state(tup):
-    rs = np.random.RandomState()
-    rs.set_state(tup)
+    rs = np.random.RandomState(0)
+    if tup is not None:
```
fyrestone, line 317: When the `tup` is None?

chaokunyang: when serializing `random_state`, if passed random state is  None, the `_on_serialize_random_state` just return None. In such situation,  the `tup` is `None`. I'm not sure why `random_state` is None, but I think it's a bug in previous and without this fix, `GroupBySample` shuffle on ray will fail. Mars tests won't fail just because it runs in test session which no serialziation happend. The exception is soemthing like follows and we can submmit another PR to fix it:
```
E         File "mars/serialization/core.pyx", line 924, in mars.serialization.core.deserialize
E           raise exc_value
E         File "mars/serialization/core.pyx", line 888, in mars.serialization.core.deserialize
E           deserialized = _deserial_single(
E         File "mars/serialization/core.pyx", line 802, in mars.serialization.core._deserial_single
E           res = serializer.deserial(serialized[_COMMON_HEADER_LEN:], context, subs)
E         File "mars/serialization/core.pyx", line 103, in mars.serialization.core.Serializer.deserial
E           cpdef deserial(self, tuple serialized, dict context, list subs):
E         File "/Users/chaokunyang/Desktop/chaokun/python/mars/mars/core/operand/base.py", line 338, in deserial
E           operand: Operand = super().deserial(serialized, context, subs)
E         File "/Users/chaokunyang/Desktop/chaokun/python/mars/mars/serialization/serializables/core.py", line 238, in deserial
E           self._set_field_value(obj, field, value)
E         File "/Users/chaokunyang/Desktop/chaokun/python/mars/mars/serialization/serializables/core.py", line 220, in _set_field_value
E           field.set(obj, field.on_deserialize(value))
E         File "/Users/chaokunyang/Desktop/chaokun/python/mars/mars/tensor/random/core.py", line 317, in _on_deserialize_random_state
E           rs.set_state(tup)
E         File "mtrand.pyx", line 357, in numpy.random.mtrand.RandomState.set_state
E       TypeError: state must be a dict or a tuple.
```

chaokunyang: Fixed it in #3149, I'll merge master after it's merged

mars/services/task/execution/ray/executor.py
chaokunyang: How about add `sort_by` parameter to `iter_indep` oor add a `iter_indep_sorted` API to `DirectedGraph`. I did a rough tests, sort chunks is much faster tghen serialization. And if we omit `input_keys`, the task spec is small, we can also cache more task spec for lineage reconstruction 
image

mars/services/task/execution/ray/executor.py
chaokunyang: reverted to print exception directly

mars/services/task/execution/ray/executor.py
chaokunyang: Changed to return None for pure_depdend_keys instead of coresponding ObjectRef

mars/services/ray_utils.py
```diff
@@ -0,0 +1,24 @@
...
+
+import functools
+
+from ..utils import lazy_import
+
+ray = lazy_import("ray")
+
+
+@functools.lru_cache(maxsize=None)  # Specify maxsize=None to make it faster
+def _ray_export_once(func_or_cls):
```
fyrestone, line 23: The original `_get_ray_executor` accepts no arguments, so it's OK to set the maxsize to None. The `_ray_export_once` accepts variety functions or classes, the `maxsize=None` will have potential memleak.

If the export once is not used very often, can we just leave it as it is and not introduce a new util?

mars/services/task/execution/ray/shuffle.py
```diff
@@ -0,0 +1,162 @@
...
+                }
+            )
+
+    @staticmethod
+    def _get_sorted_filled_reducers(
+        reducer_subtasks: Iterable[Subtask], n_reducers: int
+    ):
+        # For operands such as `PSRSAlign`, sometimes `reducer_subtasks` might be less than `n_reducers`.
+        # fill missing reducers with `None`.
+        filled_reducers = list(range(n_reducers))
```
fyrestone, line 77: If the `reducer_subtasks` is less than `n_reducers` as the comment, the missing value in the `filled_reducers` will be a number not None. The comment is not match with the code logic.

mars/services/task/execution/ray/shuffle.py
```diff
@@ -0,0 +1,162 @@
...
+            # reducers subtask should be sorted by reducer_index and MapReduceOperand.map should insert shuffle block
+            # in reducers order, otherwise shuffle blocks will be sent to wrong reducers.
+            sorted_filled_reducer_subtasks = self._get_sorted_filled_reducers(
+                reducer_subtasks, n_reducers
+            )
+            self.reducer_indices.update(
+                {
+                    subtask: (shuffle_index, reducer_ordinal)
+                    for reducer_ordinal, subtask in enumerate(
+                        sorted_filled_reducer_subtasks
```
fyrestone, line 66: If the item of `sorted_filled_reducer_subtasks` may be None, is it should be skipped?

mars/services/task/execution/ray/shuffle.py
```diff
@@ -0,0 +1,162 @@
...
+            A mapper or reducer subtask.
+        Returns
+        -------
+        n_reducers : int
+            The number of shuffle blocks that a mapper operand outputs.
+        """
+        mapper_index = self.mapper_indices.get(subtask)
+        if mapper_index:
+            shuffle_index = mapper_index[0]
+        else:
```
fyrestone, line 141: When goes to `else`?

mars/services/task/execution/ray/executor.py
chaokunyang: It depends on the keys size, the sort algorithm has `O(n)log(n)` time complexity. If the keys are small, sort is much faster than serialization.
image
And the input keys in subtask chunk graph is less than 5 normally so the cost is ignorable.

I agree with you that sort should be used sparely. If we can avoid sort, the code will be more clean. If we serializing dict by two list, the order will be fixed. But sort is more explict and customizing may should be put in another PR if we really need it?

mars/core/entity/tileables.py
```diff
@@ -81,6 +87,7 @@ def tile(
                 # without iterative tiling
                 tiled_result = tile_handler(op)
         finally:
+            cls._check_shuffle(tiled_result)
```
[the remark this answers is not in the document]
chaokunyang: Many chunks are not contained in tileable.chunks, such as `df.describe`.  We can't get all generated chunks here.
image

mars/core/operand/fetch.py
```diff
@@ -47,3 +54,11 @@ class FetchShuffle(Operand):
     source_keys = ListField("source_keys", FieldTypes.string)
     source_idxes = ListField("source_idxes", FieldTypes.tuple(FieldTypes.uint64))
     source_mappers = ListField("source_mappers", FieldTypes.uint16)
+    n_mappers = Int32Field("n_mappers")
+    n_reducers = Int32Field("n_reducers")
+    shuffle_fetch_type = AnyField("shuffle_fetch_type")
```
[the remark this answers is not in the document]
chaokunyang: Aggred, `ReferenceField` is better

mars/core/operand/shuffle.py
```diff
@@ -112,3 +128,11 @@ def iter_mapper_data(
             ctx, input_id, mapper_id, pop=pop, skip_none=skip_none
         ):
             yield data
+
+    def execute(self, ctx, op):
+        """The mapper stage must ensure all mapper blocks are inserted into ctx and no blocks
```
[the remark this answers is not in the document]
chaokunyang: Just for `fetch_by_index`

mars/services/task/execution/ray/executor.py
```diff
@@ -154,10 +219,41 @@ def execute_subtask(
...
+    if not is_mapper:
+        expect_output_keys, _ = _get_subtask_out_info(subtask_chunk_graph, False)
+        output_keys = output.keys()
+        assert expect_output_keys == output_keys, (expect_output_keys, output_keys)
+    logger.info("Finish executing subtask %s.", subtask_id)
     return output_values[0] if len(output_values) == 1 else output_values
 
 
+def _get_start_chunks(chunk_graph):
+    return sorted(list(chunk_graph.iter_indep()), key=lambda c: c.key)
```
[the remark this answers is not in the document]
chaokunyang: Good suggestion, this helps make code more cleaner.

mars/services/task/execution/ray/executor.py
```diff
@@ -118,29 +120,92 @@ def execute_subtask(
...
-            execute(context, chunk.op)
+            try:
+                execute(context, chunk.op)
+            except Exception:
+                logger.exception(
+                    "Execute operand %s of graph %s failed.",
+                    chunk.op,
+                    subtask_chunk_graph.to_dot(),
+                )
+                raise
```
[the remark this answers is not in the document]
chaokunyang: I prefer catch exception per loop.
image
is more cleaner for me compared to:
image

mars/services/task/execution/ray/shuffle.py
```diff
@@ -0,0 +1,162 @@
...
+        self.mapper_indices = {}
+        self.reducer_indices = {}
+        for shuffle_index, proxy_subtask in enumerate(self._proxy_subtasks):
+            # Note that the reducers can also be mappers such as `DuplicateOperand`.
+            mapper_subtasks = subtask_graph.predecessors(proxy_subtask)
+            reducer_subtasks = subtask_graph.successors(proxy_subtask)
+            n_mappers = len(mapper_subtasks)
+            n_reducers = _get_reducer_operand(
+                reducer_subtasks[0].chunk_graph
+            ).n_reducers
```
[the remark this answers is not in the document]
chaokunyang: Good suggestion, I'll submit another PR to add `n_reducers` and `n_mappers` to proxy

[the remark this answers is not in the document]
chaokunyang: added `n_reducers/n_mappers` in https://github.com/mars-project/mars/pull/3160

mars/services/ray_utils.py
chaokunyang: 1. As name `once` in `_ray_export_once` shows it export a `class/function` only once, so it won't be called frequently, and memleak won't happen.
2. Since memleak won't happen and the code ara repeted. why not just make it as an util?

mars/services/task/execution/ray/shuffle.py
chaokunyang: If the `subtask` isn't a mapper sutbask, it will go to `else` branch.

mars/core/entity/tileables.py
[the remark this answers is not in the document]
qinxuye: Just check the `mars/mars/services/task/supervisor/tests/task_preprocessor.py` to see if it's possible to integrate the check into it, because the check logic is sealed in the tests module instead of putting it directly into the normal code which looks really ugly.

mars/services/task/execution/ray/shuffle.py
chaokunyang: Good catch, Fixed it

mars/core/entity/tileables.py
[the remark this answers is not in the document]
chaokunyang: Actually I tried to add the checks to `CheckedTaskPreprocessor` at first, but the `CheckedTaskPreprocessor` only have `ChunkGraph` as inputs. Since some reducers are not contained in the chunk graph. Add checks here won't cover all cases. I agree with you add the checks in normal code is not elegant. But seems we can't find a place to get all tiled reducers. The checks here just to capture operands implementation error.

mars/services/ray_utils.py
fyrestone: If we make it as a util, then it may be used some where by somebody. We can't assume that the `_ray_export_once` only exports one or two `class / function`. My suggestion is mentioned as above,

- Not make it as a util (fastest).
- Make it as a util and remove `maxsize=None` to avoid potential memleak.

mars/services/task/execution/ray/shuffle.py
fyrestone: As above mentioned, the item of `sorted_filled_reducer_subtasks` may be None, so it's possible to update a `{None: (shuffle_index, reducer_ordinal)}` to the `self. reducer_indices`. Is that expected?

mars/services/ray_utils.py
chaokunyang: Reverted

mars/core/entity/tileables.py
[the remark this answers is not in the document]
chaokunyang: The checks here just to capture operands implementation error. The normal codepath doesn't need this.

mars/services/task/execution/ray/shuffle.py
fyrestone: This is a bit clearer:
```python
index = self.mapper_indices.get(subtask) or self.reducer_indices.get(subtask)
if index is None:
    raise Exception("The {subtask} should be a mapper or a reducer.")
else:
    shuffle_index, _ = index
```

chaokunyang:
Since this PR introduce `build_fetch_shuffle` which doesn't exist on master, some asv benchmarks will fail
image

mars/tests/test_utils.py
```diff
@@ -651,3 +651,58 @@ async def f2(x):
...
+            assert ex_log.count("raise_ex1") == 3, ex_log
+            assert ex_log.count("exception func1") == 1, ex_log
+            assert ex_log.count("raise_ex2") == 3, ex_log
+            assert ex_log.count("exception func2") == 1, ex_log
+    finally:
+        if os.path.exists(log_file_name):
+            os.remove(log_file_name)
+
+
+def _log_exception(log_file_name):
```
fyrestone, line 675: It seems that the `log_exception_wrapper` code is not used. Can you remove it and the related tests from this PR? Then you can add them back in some PR when they are needed.

mars/utils.py
```diff
@@ -544,34 +544,55 @@ def _is_fast_dtype(dtype):
...
+        source_mappers=source_mappers,
+        n_mappers=n_mappers,
+        n_reducers=n_reducers,
+        shuffle_fetch_type=shuffle_fetch_type,
+        gpu=chunk.op.gpu,
+    )
+    return op.new_chunk(
+        None,
+        is_broadcaster=chunk.is_broadcaster,
+        kws=[params],
```
fyrestone, line 581: Is the fetch op only has one output? I noticed that the `[params]` is passed to kws.

mars/tensor/reshape/reshape.py
```diff
@@ -462,7 +463,16 @@ def _execute_map(cls, ctx, op):
...
             target_chunk_idx.reverse()
 
-            ctx[chunk.key, tuple(target_chunk_idx)] = group_indices + (group_data,)
+            mapper_outputs[chunk.key, tuple(target_chunk_idx)] = group_indices + (
+                group_data,
+            )
+
+        # ensure all mapper data are inserted context and fill missing partition with None
+        for target_chunk_idx in itertools.product(
+            *[range(dim_chunk_cnt) for dim_chunk_cnt in dim_chunk_counts]
```
fyrestone, line 472:
```suggestion
            *(range(dim_chunk_cnt) for dim_chunk_cnt in dim_chunk_counts)
```

mars/services/task/execution/ray/shuffle.py
fyrestone: I noticed that the `get_n_reducers` is called only when the subtask is mapper, so is the `else` needed?

mars/core/entity/tileables.py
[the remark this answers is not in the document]
chaokunyang: Moved the checks to `CheckedTaskPreprocessor`

mars/services/task/execution/ray/shuffle.py
chaokunyang: They are already updated to `self.reducer_indices`.

mars/tests/test_utils.py
chaokunyang: reverted

mars/utils.py
chaokunyang: `FetchShuffle` only have one output which is reduce op. But is this related to this PR?

mars/services/task/execution/ray/shuffle.py
chaokunyang: `else` can be reserved for use in future

fyrestone (approved):
LGTM

qinxuye (approved):
LGTM

PR #3041, opened 2022-05-17
Shuffle both sides at the same time for `md.merge`

hekaisheng:
## What do these changes do?

For now, we shuffle left and right separately and make reducer data locate in different workers, this leads to additional transfer when execute merge operand. This PR only shuffles once for both left and right, so that reducer data locate in same worker and no transfer for following execution.

## Related issue number

Fixes #xxxx

## Check code requirements

### DIFF
--- a/mars/dataframe/merge/merge.py
+++ b/mars/dataframe/merge/merge.py
@@ -35,7 +35,7 @@ from ...serialization.serializables import (
     NamedTupleField,
 )
 from ...typing import TileableType
-from ...utils import has_unknown_shape
+from ...utils import has_unknown_shape, lazy_import
 from ..base.bloom_filter import filter_by_bloom_filter
 from ..core import DataFrame, Series, DataFrameChunk
 from ..operands import DataFrameOperand, DataFrameOperandMixin, DataFrameShuffleProxy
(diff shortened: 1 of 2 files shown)


hekaisheng:
As this PR changes the tile logic of merge, it will affect assign time which make asv benchmark fail in CI.

mars/dataframe/merge/merge.py
```diff
@@ -123,16 +116,17 @@ def execute_map(cls, ctx, op):
...
+            input_idx_to_df = dict(
+                op.iter_mapper_data_with_index(ctx, mapper_id=i, skip_none=True)
+            )
+            row_idxes = sorted({idx[0] for idx in input_idx_to_df})
+            res = []
+            for row_idx in row_idxes:
+                row_df = input_idx_to_df.get((row_idx, 0), None)
+                if row_df is not None:
+                    res.append(row_df)
+            ctx[chunk.key] = pd.concat(res, axis=0)
```
qinxuye, line 129: Is is possible to use `cudf` or `pd` to concat?

qinxuye (commented):
LGTM overall.

qinxuye (approved):
LGTM

PR #3052, opened 2022-05-19
[Ray] Implements get_chunks_meta for Ray execution context

fyrestone:
## What do these changes do?

- [x] Implement `get_chunks_meta` for Ray execution context.
- [x] Add `memory_size` meta for `auto_merge_chunks`.
- [x] Empty implements these APIs for Ray execution context:
  - new_custom_log_dir
  - register_custom_log_path
  - set_progress
  - set_running_operand_key
  - get_storage_info

## Related issue number

## Check code requirements

### DIFF
--- a/mars/services/task/execution/ray/executor.py
+++ b/mars/services/task/execution/ray/executor.py
@@ -224,7 +224,7 @@ class RayTaskExecutor(TaskExecutor):
         )
 
     @staticmethod
-    @functools.lru_cache(maxsize=1)
+    @functools.lru_cache(maxsize=None)
     def _get_ray_executor():
         # Export remote function once.
         return ray.remote(execute_subtask)
--- a/mars/services/task/execution/ray/executor.py
+++ b/mars/services/task/execution/ray/executor.py
@@ -15,6 +15,7 @@
 import asyncio
 import functools
 import logging
+from dataclasses import dataclass
 from typing import List, Dict, Any, Set
 from .....core import ChunkGraph, Chunk, TileContext
 from .....core.context import set_context
@@ -30,6 +31,7 @@ from .....resource import Resource
 from .....serialization import serialize, deserialize
 from .....typing import BandType
 from .....utils import (
+    calc_data_size,
     lazy_import,
     get_chunk_params,
     get_chunk_key_to_data_keys,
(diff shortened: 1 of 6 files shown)


mars/services/task/execution/ray/fetcher.py
```diff
@@ -23,17 +24,28 @@
...
     required_meta_keys = ("object_refs",)
 
     def __init__(self, **kwargs):
-        _make_query_function_remote()
-
         self._fetch_info_list = []
         self._no_conditions = True
 
+    @staticmethod
+    @functools.lru_cache(maxsize=None)
```
chaokunyang, line 44: `maxsize=None` is unnecessary

mars/services/task/execution/ray/executor.py
```diff
@@ -207,7 +224,7 @@ async def _get_apis(cls, session_id: str, address: str):
         )
 
     @staticmethod
-    @functools.lru_cache(maxsize=1)
+    @functools.lru_cache(maxsize=None)
```
chaokunyang, line 227: ditto

mars/services/task/execution/ray/fetcher.py
fyrestone: The default `max_size` of lru_cache is 128. if the `max_size > 0`, then the cache implementation is more complicated. Specify the `maxsize=None` to use a simpler cache implementation.

chaokunyang: `_remote_query_object_with_condition` has no params, so I guess there will be only one cache item. Skip specifying will be OK?

fyrestone: > `_remote_query_object_with_condition` has no params, so I guess there will be only one cache item. Skip specifying will be OK?

Since the `_remote_query_object_with_condition` has no params, the cache item should be always 1. Specify a None maxsize to make it faster because we don't need a maxsize.

fyrestone: Comments added. Thanks.

chaokunyang (approved):
LGTM

qinxuye (approved):
LGTM

PR #3053, opened 2022-05-19
Use OS-designated ports instead of random ports to create sub pools

wjsi:
## What do these changes do?

It is often hard to obtain free ports, esp. in containers with host ports. It is better to assign ports of sub-pools by OS itself.

## Related issue number

Fixes #xxxx

## Check code requirements


qinxuye (approved):
LGTM, about router.set_instance(None), another PR can be raised.

hekaisheng (approved):
LGTM

PR #3057, opened 2022-05-20
Fix chunk index error in auto_merge_chunks

### LINKED ISSUE #3056: [BUG] wrong index of the last chunk after auto merging
**Describe the bug**
When I run TCP-H 19th query under 10G scale factor，it throws an exception shown below.

image

or

image

It seems that some input chunk indices are wrong.

After debugging, I found that the last chunk index is wrong after `df.merge()` in TCP-H 19th query.

**To Reproduce**
Run TCP-H 19th query under 10G scale factor in a 5 node cluster.

yuyiming:
## What do these changes do?

if a single chunk remains to merge in auto_merge_chunks, current code tries not to create a new concat OP and just use the last chunk. But the index of last chunk should be changed after merging.

This PR fixes this issue.

## Related issue number

Fixes #3056

## Check code requirements

- [X] tests added / passed (if needed)
- [X] Ensure all linting tests pass, see [here](https://docs.pymars.org/en/latest/development/contributing.html#check-code-styles) for how to run them

### DIFF
--- a/mars/dataframe/tests/test_utils.py
+++ b/mars/dataframe/tests/test_utils.py
@@ -614,7 +614,10 @@ def test_auto_merge_chunks():
     assert len(df2.chunks) == 2
     assert isinstance(df2.chunks[0].op, DataFrameConcat)
     assert len(df2.chunks[0].op.inputs) == 3
-    assert df2.chunks[1] is df.chunks[-1]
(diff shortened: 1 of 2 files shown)


hekaisheng (approved):
LGTM

PR #3058, opened 2022-05-20
Add `_send_small_objects` for sender to optimize sending small objects

hekaisheng:
## Related issue number
This PR implements a method that send small objects directly to receiver without split them into small blocks.

Fixes #xxxx

## Check code requirements


qinxuye:
Seems vineyard cannot pass the ut for this PR, @sighingnow do you have to time to help to see what's going on?

sighingnow:
> Seems vineyard cannot pass the ut for this PR, @sighingnow do you have to time to help to see what's going on?

I would take a look later today.

sighingnow:
> > Seems vineyard cannot pass the ut for this PR, @sighingnow do you have to time to help to see what's going on?
> 
> I would take a look later today.

Sorry for the delay... Will be fixed today.

sighingnow:
I failed to reproduce the failure if only run the single test case. But I found the test case [`test_execute_describe`](https://github.com/mars-project/mars/blob/master/mars/deploy/oscar/tests/test_local.py#L336-L363) failed both for `default` storage and `vineyard` storage under this PR.

I have tried to debugging but failed to address the root cause. I think it is not a vineyard-only issue, @hekaisheng could you please help to try the `test_execute_describe` in your environment?

You could update the [`params`](https://github.com/mars-project/mars/blob/master/mars/deploy/oscar/tests/test_local.py#L104-L107) to test the `default` storage only. The failre could be reproduced by

```
pytest -s -vvv mars/deploy/oscar/tests/test_local.py -k test_execute_describe
```

Issue #3054, opened 2022-05-19
[RFC] Make mapper output blocks consist with num_reducers


chaokunyang:
# Background
When building chunk graph, Mars chunks won't be added to graph if some chunks not used by downstream, because mars build graph back to first. This will make shuffle optimization tricky. For example, Mars `describe` use `PSRS` algorithm, which has a shuffle stage named `PSRSAlign`. Sometimes some reducer outputs for this shuffle will not be used, and the resulting chunk graph will has reducer chunks less than mapper blocks.

In the following exmaple, we have three `PSRSAlign` mappers each produce three shuffle blocks, but we only got two reducers of three in the final chunk graph.

```
def test_describe_execution(setup):
    s_raw = pd.Series(np.random.rand(10))
    # test multi chunks
    series = from_pandas_series(s_raw, chunk_size=3)
    r = series.describe(percentiles=[])
    r.execute().fetch()
```
![image](https://user-images.githubusercontent.com/12445254/169302407-0a7acbf9-408b-43ca-a838-ef961931dfea.png)
![image](https://user-images.githubusercontent.com/12445254/169302218-c50e6a91-e604-45e2-95e0-11a4b47fc363.png)

PR #3078, opened 2022-05-24
[shuffle] Make shuffle mapper outputs deterministic

chaokunyang:
## What do these changes do?
This PR make shuffle mapper outputs and order deterministic, so that we can read shuffle data by index instead of data keys in future, which will reduce shuffle meta overhead.

## Related issue number
https://github.com/mars-project/meps/pull/2
https://github.com/mars-project/mars/issues/3054
https://github.com/mars-project/mars/pull/3040
https://github.com/mars-project/mars/issues/2916

## Check code requirements

### DIFF
--- a/mars/services/subtask/worker/tests/subtask_processor.py
+++ b/mars/services/subtask/worker/tests/subtask_processor.py
@@ -11,11 +11,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+from collections import defaultdict
 from typing import Any, Dict
 
 from .....core import OperandType
 from .....tests.core import _check_args, ObjectCheckMixin
+from ...utils import iter_output_data
 from ...worker.processor import SubtaskProcessor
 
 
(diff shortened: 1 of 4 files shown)


mars/services/subtask/worker/tests/subtask_processor.py
```diff
@@ -81,6 +82,27 @@ def _execute_operand(self, ctx: Dict[str, Any], op: OperandType):
...
+                self.subtask.chunk_graph, self._datastore
+            )
+            if is_shuffle
+        }
+        # assert output keys order consistent
+        if shuffle_output:
+            mapper_reducer_indices = defaultdict(list)
+            for chunk_key, reducer_index in shuffle_output.keys():
+                mapper_reducer_indices[chunk_key].append(reducer_index)
+            for reducer_indices in mapper_reducer_indices.values():
```
wjsi, line 99: As the assertion is a little expensive, is it possible to make it CI-only by judging from environ `CI=true`? (see usage of `_is_ci` in https://github.com/mars-project/mars/blob/master/mars/serialization/serializables/field.py).

wjsi (commented):
I don't think it is safe to rely on orders of context as there is no guarantation in our primitive designs. A smarter design might be changing keys into `(shuffle_proxy_key, mapper_id, reducer_id)` and spaces of mapper keys are also saved.

chaokunyang:
> 
@wjsi 
If there are lots of shuffle partitions, `(shuffle_proxy_key, mapper_id, reducer_id)` will have a big meta overhead. And it will be difficult to make mars support shuffle which has PB-scale data. If context insert order is not an appropriate assumption, how about order all shuffle outputs by reducer index in `mars.services.subtask.utils.iter_output_data`?

mars/services/subtask/worker/tests/subtask_processor.py
chaokunyang: The assertion is added into `CheckedSubtaskProcessor`, which won't be used in production. You mean we also add the check in `mars.services.subtask.worker.processor.SubtaskProcessor` and enable it when only running in CI?

chaokunyang: maybe this check is not needed, if we order all mapper outputs by reducer index in ` mars.services.subtask.utils.iter_output_data`

wjsi:
> > 
> 
> @wjsi If there are lots of shuffle partitions, `(shuffle_proxy_key, mapper_id, reducer_id)` will have a big meta overhead. And it will be difficult to make mars support shuffle which has PB-scale data. If context insert order is not an appropriate assumption, how about order all shuffle outputs by reducer index in `mars.services.subtask.utils.iter_output_data`?

That's acceptable.

PR #3080, opened 2022-05-24
[Ray] Fix ray worker failover

### LINKED ISSUE #3079: [BUG] Mars worker not recovered on ray master
**Describe the bug**
In ray master, If a actor created with `max_restarts=-1` is restarting, call actor method will raise exception instead of pending in caller, which make ray worker failover not work. 

**To Reproduce**
1. Your Python version: 3.7.9
3. Versions of crucial packages, such as ray, numpy, scipy and pandas: ray master
4. Full stack of the error.
image
```
import ray
import time

@ray.remote(max_restarts=-1)
class A:
  def __init__(self):
    if ray.get_runtime_context().was_current_actor_reconstructed:
      import os
      os._exit(-1)
    print(ray.get_runtime_context().was_current_actor_reconstructed)
    time.sleep(3)

  def f(self):
    return 1

  def f1():
    time.sleep(30)
a = A.remote()
print(ray.get(a.f.remote()))
r = a.f1.remote()
ray.kill(a, no_restart=False)
try:
  ray.get(r)
except Exception:
  pass
print(ray.get(a.f.remote()))
```
**Expected behavior**
This can be fixed by specifying max_retries=-1 when call actro methods

**Additional context**

chaokunyang:
## What do these changes do?
In ray master, If a actor created with `max_restarts=-1` is restarting, call actor method will raise exception instead of pending in caller. This PR fix it by specifing `max_retries=-1` when querying actor state.

## Related issue number

Fixes #3079 

## Check code requirements

### DIFF
--- a/mars/deploy/oscar/tests/test_ray.py
+++ b/mars/deploy/oscar/tests/test_ray.py
@@ -677,7 +677,7 @@ async def test_auto_scale_in(ray_large_cluster):
         assert await autoscaler_ref.get_dynamic_worker_nums() == 2
 
 
-@pytest.mark.timeout(timeout=1000)
+@pytest.mark.timeout(timeout=200)
 @pytest.mark.parametrize("ray_large_cluster", [{"num_nodes": 4}], indirect=True)
 @require_ray
 @pytest.mark.asyncio
--- a/mars/deploy/oscar/tests/test_ray.py
+++ b/mars/deploy/oscar/tests/test_ray.py
@@ -578,7 +578,7 @@ async def test_release_worker_during_reconstructing_worker(
     class FakeActor:
         state = FakeActorMethod()
 
-    def _get_actor(*args):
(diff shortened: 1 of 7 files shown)


qinxuye (approved):
LGTM

mars/deploy/oscar/tests/test_ray.py
```diff
@@ -578,7 +578,7 @@ async def remote(self):
     class FakeActor:
         state = FakeActorMethod()
 
-    def _get_actor(*args):
+    def _get_actor(*args, **kwargs):
```
zhongchun, line 581: A little confused. It doesn't seem necessary to do this change.

chaokunyang: In some version of ray, ray will pass `namespace` named parameter to this method. Without it, the method call will fail

zhongchun (approved):
LGTM.

PR #3081, opened 2022-05-24
[Ray] Create RayTaskState actor as needed by default

fyrestone:
## What do these changes do?

- Create RayTaskState actor as needed by default _(the `create_task_state_actor_as_needed` is True)_.
```yml
task:
  execution_config:
    backend: ray
      ray:
        create_task_state_actor_as_needed: True
```
- Implements `get_worker_addresses` API for Ray execution context. It returns virtual worker addresses ~~for auto choosing groupby method~~ _(this PR https://github.com/mars-project/mars/pull/3084 has removed get_worker_addresses() usage from groupby operand)_.
- Remove `test_task_manager_on_ray_dag.py`, mark `@pytest.mark.ray_dag` to the cases in `test_task_manager.py` instead.
- Pin `protobuf < 4` to avoid Ray import error.

## Related issue number

## Check code requirements


mars/services/task/execution/ray/executor.py
```diff
@@ -426,7 +448,9 @@ async def get_available_band_resources(self) -> Dict[BandType, Resource]:
             idx = 0
             for band_resource in band_resources:
                 for band, resource in band_resource.items():
-                    virtual_band_resources[(f"ray_virtual://{idx}", band)] = resource
+                    virtual_band_resources[
+                        (f"ray_virtual_address_{idx}:0", band)
```
zhongchun, line 452: The band name of mars on ray is like this: `ray://mars_cluster_1653388173/0/0`, should we keep it be consistent with that?

fyrestone: The virtual worker address is for compatible with Mars worker ip address.

zhongchun (approved):
LGTM.

wjsi (approved):
LGTM

PR #3085, opened 2022-05-25
Call immutable web API only once when previous call blocks

wjsi:
## What do these changes do?

Call immutable web API only once when previous call does not return. This reduces potential call flood for services.

## Related issue number

Fixes #xxxx

## Check code requirements


qinxuye (approved):
LGTM

hekaisheng (approved):
LGTM

PR #3086, opened 2022-05-25
Simplify router clean up when pools or clusters ends

wjsi:
## What do these changes do?

Simplify router clean up when pools or clusters ends.

## Related issue number

Fixes #xxxx

## Check code requirements

- [ ] tests added / passed (if needed)
- [ ] Ensure all linting tests pass, see [here](https://docs.pymars.org/en/latest/development/contributing.html#check-code-styles) for how to run them

PR #3087, opened 2022-05-25
[BACKPORT] Use OS-designated ports instead of random ports to create sub pools (#3053)

wjsi:
Backport of #3053 and #3086


qinxuye (approved):
LGTM

PR #3092, opened 2022-05-27
Support SciPy special functions with multiple output in the form of np.array

RandomY-2:
## What do these changes do?

Currently, if we want to integrate special functions such as fresnel, we get an AttributeError saying tuple objects have no attribute dtype. This is because, for functions with multiple outputs, SciPy returns a tuple instead of an array, but during later processings, mars tries to get certain numpy attributes such as dtype from the returned value. One approach to solve this is to convert tuple objects to np.array when handling the output data type. Then, we can support functions with multiple outputs such as ```special.fresnel```.

One difference now though is that users will receive an array as the output instead of a tuple, which I don't know if is fine. 

## Check code requirements

### DIFF
--- a/docs/source/reference/tensor/special.rst
+++ b/docs/source/reference/tensor/special.rst
@@ -35,7 +35,7 @@ Bessel functions
    mars.tensor.special.hankel2e
 
 
-Error function
+Error function and Fresnel Integrals
 --------------
 
 .. autosummary::
@@ -48,6 +48,7 @@ Error function
    mars.tensor.special.erfi
    mars.tensor.special.erfinv
    mars.tensor.special.erfcinv
+   mars.tensor.special.fresnel
 
 
 Ellipsoidal harmonics
--- a/mars/lib/sparse/__init__.py
+++ b/mars/lib/sparse/__init__.py
@@ -271,6 +271,8 @@ erfi = partial(_call_unary, "erfi")
 erfinv = partial(_call_unary, "erfinv")
 erfcinv = partial(_call_unary, "erfcinv")
 
+fresnel = partial(_call_unary, "fresnel")
+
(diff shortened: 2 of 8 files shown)


wjsi:
For n-output functions, we expect the Mars implementation to output n Mars objects instead of one. Implementations of current specials only considered single outputs. I would expect the implementation like

```python
@_register_special_op
@arithmetic_operand(sparse_mode="unary", n_outputs=2)
class TensorFresnel(TensorSpecialUnaryOp):
    _func_name = "fresnel"


@implement_scipy(spspecial.fresnel)
@infer_dtype(spspecial.fresnel)
def fresnel(x, out=None, where=None, **kwargs):
    op = TensorFresnel(**kwargs)
    return op(x, out=out, where=where)
```

where `n_outputs` argument and multiple-output version of `infer_dtype` are to be implemented. This might be a little challenging and you can have a try.

RandomY-2:
> For n-output functions, we expect the Mars implementation to output n individual Mars objects instead of one. Implementations of current specials only considered single outputs. I would expect the implementation like
> 
> ```python
> @_register_special_op
> @arithmetic_operand(sparse_mode="unary", n_outputs=2)
> class TensorFresnel(TensorSpecialUnaryOp):
>     _func_name = "fresnel"
> 
> 
> @implement_scipy(spspecial.fresnel)
> @infer_dtype(spspecial.fresnel)
> def fresnel(x, out=None, where=None, **kwargs):
>     op = TensorFresnel(**kwargs)
>     return op(x, out=out, where=where)
> ```
> 
> where `n_outputs` argument and multiple-output version of `infer_dtype` are to be implemented. This might be a little challenging and you can have a try.

Thanks for the suggestion! I can give it a try.

PR #3099, opened 2022-05-31
Basic support for tensors with tuple outputs

RandomY-2:
## What do these changes do?

Currently, the execute function under tensor.arithmetic only checks for returns that have only one output array. When fetching from StorageAPI, the transformation is also to the entire output. This PR changes these two parts to be applied to each individual output array if there are multiple of them returned in the form of a tuple. So now we can return a tuple for functions such as special.fresnel and can interact with the returned value using tuple syntax. For example:

```
from mars.tensor import special
import mars.tensor as mt

x = mt.linspace(-3, 3)
S, C = special.fresnel(x).execute().fetch()
```
will assign the two arrays to S and C respectively. 

However, this tuple support is still basic. For example, we still cannot do tensor-based computation between the executed result and another tensor, which gives AttributeError because the underlying data structure is a tuple instead of the original array.

## Check code requirements


wjsi:
Actually we need

```python
from mars.tensor import special
import mars.tensor as mt

x = mt.linspace(-3, 3)
S, C = special.fresnel(x)
```

while `S` and `C` are both `mt.Tensor` and can be executed separately. Returning tuples for tensors may only be used as internal implementations, not for user APIs.

RandomY-2:
> Actually we need
> 
> ```python
> from mars.tensor import special
> import mars.tensor as mt
> 
> x = mt.linspace(-3, 3)
> S, C = special.fresnel(x)
> ```
> 
> while `S` and `C` are both `mt.Tensor` and can be executed separately. Returning tuples for tensors may only be used as internal implementations, not for user APIs.

I see. And do we want fresnel itself to be executable? In which case it would need to be able to return both a tuple of tensors and a tensor itself, which output will be the tuple with S and C computed and we can add logics to process the tuple. In that case I am thinking about maybe adding a flag to fresnel to tell it what to output, so something like:

```python
x = mt.linspace(-3, 3)
S, C = special.fresnel(x, return_tuple=True) # fresnel returns a tuple with two executable tensors

res = special.fresnel(x) # if return_tuple=False or no return_tuple, fresnel returns a tensor
res.execute()            # which can be executed and the fetched output is a tuple with two arrays, and we
                         # add logics to handle tensor-based computation between the underlying tuple and other arrays
```

wjsi:
> > Actually we need
> > 
> > ```python
> > from mars.tensor import special
> > import mars.tensor as mt
> > 
> > x = mt.linspace(-3, 3)
> > S, C = special.fresnel(x)
> > ```
> > 
> > while `S` and `C` are both `mt.Tensor` and can be executed separately. Returning tuples for tensors may only be used as internal implementations, not for user APIs.
> 
> I see. And do we want fresnel itself to be executable? In which case it would need to be able to return both a tuple of tensors and a tensor itself. In that case, I am thinking about maybe adding a flag to fresnel to tell it what to output, so something like:
> 
> ```python
> x = mt.linspace(-3, 3)
> S, C = special.fresnel(x, return_tuple=True) # fresnel returns a tuple with two executable tensors
> 
> res = special.fresnel(x) # if return_tuple=False or no return_tuple, fresnel returns a tensor
> res.execute()            # which can be executed and the fetched output is a tuple with two arrays, and we
>                          # add logics to handle tensor-based computation between the underlying tuple and other arrays
> ```
> 
> 
> But this would make the API different from scipy API, so maybe I should focus on returning a tuple of two executable tensors first.

Good question. We have an `ExecutableTuple` class which supports execute method and tuple extraction at the same time, thus `return_tuple` is not needed. That is, we can support `special.fresnel(x)` and `S, C = special.fresnel(x)` at the same time with an `ExecutableTuple` without any extra arguments.

RandomY-2:
> > > Actually we need
> > > ```python
> > > from mars.tensor import special
> > > import mars.tensor as mt
> > > 
> > > x = mt.linspace(-3, 3)
> > > S, C = special.fresnel(x)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > while `S` and `C` are both `mt.Tensor` and can be executed separately. Returning tuples for tensors may only be used as internal implementations, not for user APIs.
> > 
> > 
> > I see. And do we want fresnel itself to be executable? In which case it would need to be able to return both a tuple of tensors and a tensor itself. In that case, I am thinking about maybe adding a flag to fresnel to tell it what to output, so something like:
> > ```python
> > x = mt.linspace(-3, 3)
> > S, C = special.fresnel(x, return_tuple=True) # fresnel returns a tuple with two executable tensors
> > 
> > res = special.fresnel(x) # if return_tuple=False or no return_tuple, fresnel returns a tensor
> > res.execute()            # which can be executed and the fetched output is a tuple with two arrays, and we
> >                          # add logics to handle tensor-based computation between the underlying tuple and other arrays
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > But this would make the API different from scipy API, so maybe I should focus on returning a tuple of two executable tensors first.
> 
> Good question. We have an `ExecutableTuple` class which supports execute method and tuple extraction at the same time, thus `return_tuple` is not needed. That is, we can support `special.fresnel(x).execute()` and `S, C = special.fresnel(x)` at the same time with an `ExecutableTuple`, and the return type does not have to be a tuple all the time.

Thanks! I will take a look.

Issue #751, opened 2019-09-20
Support elliptic functions and integrals


hekaisheng:
Implement elliptic functions and integrals in special functions.

list in https://docs.scipy.org/doc/scipy/reference/special.html#elliptic-functions-and-integrals.

meenakshi0910:
hi @hekaisheng, Can I work on this?

hekaisheng:
> hi @hekaisheng, Can I work on this?

Sure, PR is welcome.

xsraut:
Hello, I was trying to fix this, can you please guide me on integration part, what (libraries) should I use to integrate?
Thanks.

PR #3111, opened 2022-06-02
Integrate part of scipy elliptic functions and integrals

RandomY-2:
## What do these changes do?

Integrate scipy elliptic functions and integrals except ellipj which returns multiple outputs. Some functions are within a try except block because they were not implemented in scipy v1.7, so spspecial.func may throw an AttributeError if the scipy version is low.

## Related issue number

#751 (except ellipj)

## Check code requirements

### DIFF
--- a/mars/tensor/special/tests/test_special.py
+++ b/mars/tensor/special/tests/test_special.py
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 import numpy as np
+import scipy
+import pytest
 from scipy.special import (
     gammaln as scipy_gammaln,
     erf as scipy_erf,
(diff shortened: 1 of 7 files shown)


mars/tensor/special/tests/test_special.py
```diff
@@ -291,3 +325,232 @@ def test_beta_inc():
...
+
+    assert r.nsplits == tiled_a.nsplits
+    for chunk in r.chunks:
+        assert isinstance(chunk.op, TensorEllipeinc)
+        assert chunk.index == chunk.inputs[0].index
+        assert chunk.shape == chunk.inputs[0].shape
+
+
+@pytest.mark.skipif(
+    scipy.__version__ < "1.8.0", reason="function not implemented in scipy."
```
wjsi, line 430: Comparing versions by strings is not safe. Try looking up `parse_version` in the code repo and see how we handle dependency versions like pandas.

mars/tensor/special/tests/test_special.py
```diff
@@ -12,7 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from ....lib.version import parse as parse_version
```
wjsi, line 15: We used to put imports in builtin-thirdparty-Mars order. Please take a look at other py files and change the order.

wjsi (approved):
LGTM. Thanks for the contribution!

qinxuye (approved):
LGTM

PR #3118, opened 2022-06-06
[Ray] Fix Ray context GC

fyrestone:
## What do these changes do?

This PR fixes the context GC bug of Ray. When the `MARS_CI_BACKEND=ray` env is set, the GC interval will be set to 0 to detect the GC bug as early as possible.

## Related issue number

Fixes https://github.com/mars-project/mars/issues/3119

## Check code requirements


qinxuye (approved):
LGTM

zhongchun (approved):
LGTM.

PR #3139, opened 2022-06-12
Support scipy special function with tuple output

RandomY-2:
## What do these changes do?

We can use ExecutableTuple and TensorTupleElementOp to support mars execution of scipy special functions with tuple returns. Specifically, the following syntaxes can be supported:

```python
from mars.tensor import special
import mars.tensor as mt
x = mt.linspace(-3, 3)

S, C = special.fresnel(x)
S.execute()
C.execute()
special.fresnel(x).execute()
(S + C).execute()
```

Also since we invoke scipy.special.fresnel during tensor executions, a cache is added in core.py so if we execute multiple elements of the same function during one execute() call, we don't need to do the scipy computation for each element. Later executes can just use the cached scipy outputs.

## Check code requirements

### DIFF
--- a/mars/tensor/special/core.py
+++ b/mars/tensor/special/core.py
@@ -27,11 +27,19 @@ from ..array_utils import (
 
 
 _func_name_to_special_cls = {}
+_tuple_func_to_res = {}
 
 
 def _register_special_op(cls):
     if cls._func_name is not None:
-        _func_name_to_special_cls[cls._func_name] = cls
+        if getattr(cls, "_output_index", None) is None:
+            _func_name_to_special_cls[cls._func_name] = cls
+        else:
+            if cls._func_name not in _func_name_to_special_cls:
+                _func_name_to_special_cls[cls._func_name] = [
+                    None for _ in range(cls._func_outputs)
+                ]
+            _func_name_to_special_cls[cls._func_name][cls._output_index] = cls
     return cls
 
 
@@ -41,7 +49,13 @@ class TensorSpecialOperandMixin:
 
     def __new__(cls, *args, **kwargs):
         if cls._func_name is not None:
-            return object.__new__(_func_name_to_special_cls[cls._func_name])
+            if getattr(cls, "_output_index", None) is None:
+                return object.__new__(_func_name_to_special_cls[cls._func_name])
+            else:
+                return object.__new__(
+                    _func_name_to_special_cls[cls._func_name][cls._output_index]
+                )
+
         return super().__new__(cls, *args, **kwargs)
 
     @classmethod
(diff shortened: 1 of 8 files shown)


wjsi:
Given your implementation, it is not hard to use one single operand for two outputs. You may take a look at `TensorSVD` for an example, where we use `output_limit` property to define the number of outputs:

https://github.com/mars-project/mars/blob/424cfb9d422a0918a089821f2710340ff94e101a/mars/tensor/linalg/svd.py#L44-L46

use `new_tensors(..., kws[kw1, kw2])` to create multiple output objects:

https://github.com/mars-project/mars/blob/424cfb9d422a0918a089821f2710340ff94e101a/mars/tensor/linalg/svd.py#L69-L77

 and use `ctx[out.key] = xxx` to store execute results.

https://github.com/mars-project/mars/blob/424cfb9d422a0918a089821f2710340ff94e101a/mars/tensor/linalg/svd.py#L150-L155

RandomY-2:
> Given your implementation, it is not hard to use one single operand for two outputs. You may take a look at `TensorSVD` for an example, where we use `output_limit` property to define the number of outputs:
> 
> https://github.com/mars-project/mars/blob/424cfb9d422a0918a089821f2710340ff94e101a/mars/tensor/linalg/svd.py#L44-L46
> 
> use `new_tensors(..., kws[kw1, kw2])` to create multiple output objects:
> 
> https://github.com/mars-project/mars/blob/424cfb9d422a0918a089821f2710340ff94e101a/mars/tensor/linalg/svd.py#L69-L78
> 
> and use `ctx[out.key] = xxx` to store execute results.
> 
> https://github.com/mars-project/mars/blob/424cfb9d422a0918a089821f2710340ff94e101a/mars/tensor/linalg/svd.py#L150-L155

Thank you for your feedback! Will take a look.

mars/tensor/special/core.py
```diff
@@ -112,3 +115,54 @@ def execute(cls, ctx, op):
...
+
+        with device(device_id):
+            with np.errstate(**op.err):
+                if op.is_gpu():
+                    ret = cls._execute_gpu(op, xp, inputs[0])
+                else:
+                    ret = cls._execute_cpu(op, xp, inputs[0])
+
+                for i in range(len(op.outputs)):
+                    ctx[op.outputs[i].key] = ret[i]
```
wjsi, line 168: Loop over `op.outputs` itself, not indices.

```suggestion
                for output in op.outputs:
                    ctx[output] = ret[i]
```

mars/tensor/special/core.py
```diff
@@ -112,3 +115,54 @@ def execute(cls, ctx, op):
...
+                }
+                for i in range(len(res))
+            ],
+        )
+
+        if out is None:
+            return ExecutableTuple(res_tensors)
+
+        for i in range(len(res_tensors)):
+            out[i].data = res_tensors[i].data
```
wjsi, line 151: Ditto.

mars/tensor/special/core.py
```diff
@@ -112,3 +115,54 @@ def execute(cls, ctx, op):
...
+        res = func(np.ones(x.shape, dtype=x.dtype))
+        res_tensors = self.new_tensors(
+            [x],
+            kws=[
+                {
+                    "side": f"{self._func_name}[{i}]",
+                    "dtype": res[i].dtype,
+                    "shape": res[i].shape,
+                }
+                for i in range(len(res))
```
wjsi, line 143: Ditto.

mars/tensor/special/tests/test_special.py
```diff
@@ -276,6 +279,44 @@ def test_erfcinv():
...
+    raw = np.random.rand(10, 8, 5)
+    t = tensor(raw, chunk_size=3)
+
+    r = fresnel(t)
+    expect = scipy_fresnel(raw)
+
+    assert isinstance(r, ExecutableTuple)
+    assert len(r) == 2
+
+    for i in range(len(r)):
```
wjsi, line 292: Ditto.

mars/tensor/special/tests/test_special.py
```diff
@@ -276,6 +279,44 @@ def test_erfcinv():
...
+    out = ExecutableTuple([t, t])
+    r_out = fresnel(t, out=out)
+
+    assert isinstance(out, ExecutableTuple)
+    assert isinstance(r_out, ExecutableTuple)
+
+    assert len(out) == 2
+    assert len(r_out) == 2
+
+    for i in range(len(r_out)):
```
wjsi, line 310: Ditto.

mars/tensor/special/tests/test_special_execution.py
```diff
@@ -299,3 +299,25 @@ def test_quintuple_execution(setup, func):
...
+
+    raw = np.random.rand(10, 8, 6)
+    a = tensor(raw, chunk_size=3)
+
+    r = mt_func(a)
+
+    result = r.execute().fetch()
+    expected = sp_func(raw)
+
+    for i in range(len(result)):
```
wjsi, line 322: Ditto.

mars/tensor/special/err_fresnel.py
```diff
@@ -55,6 +59,12 @@ class TensorErfcinv(TensorSpecialUnaryOp):
     _func_name = "erfcinv"
 
 
+@_register_special_op
+class TensorFresnel(TensorTupleOp):
+    _func_name = "fresnel"
+    _func_outputs = 2
```
wjsi, line 65: Rename to `_n_outputs` might be better.

wjsi (commented):
Code structure LGTM. Do not use `for i in range(len(list_obj))`. Using `for i, obj in enumerate(obj_list)` when indices are needed or `for obj in obj_list` is more pythonic.

wjsi (approved):
LGTM

hekaisheng (approved):
LGTM

PR #3144, opened 2022-06-14
Cache the created worker meta api for task stage processor

### LINKED ISSUE #3143: [BUG] There maybe duplicate `WorkerMetaAPI` creation in `TaskStageProcessor`
**Describe the bug**

We can see there is a lot of duplicate `WorkerMetaAPI` creation in the method `_update_result_meta` of `mars/services/task/execution/mars/stage.py` if there is too many chunk result and their bands are same:
```python
for tileable in tile_context.values():
    chunks = [c.data for c in tileable.chunks]
    for c, params_fields in zip(chunks, self._get_params_fields(tileable)):
        address = chunk_to_result[c].meta["bands"][0][0]
        meta_api = await WorkerMetaAPI.create(session_id, address)
        call = meta_api.get_chunk_meta.delay(c.key, fields=params_fields)
        worker_meta_api_to_chunk_delays[meta_api][c] = calll
```
There are 3639 chunk results and 200 workers in my job and it costs 1414s to create worker meta apis.
Maybe we can cache the worker meta apis to avoid duplicate creation.

**To Reproduce**

1. Your Python version: python-3.7.9

zhongchun:
## What do these changes do?
There is a lot of duplicate `WorkerMetaAPI` creation in the method `_update_result_meta` of `mars/services/task/execution/mars/stage.py` if there is too many chunk result and their bands are same:
```python
for tileable in tile_context.values():
    chunks = [c.data for c in tileable.chunks]
    for c, params_fields in zip(chunks, self._get_params_fields(tileable)):
        address = chunk_to_result[c].meta["bands"][0][0]
        meta_api = await WorkerMetaAPI.create(session_id, address)
        call = meta_api.get_chunk_meta.delay(c.key, fields=params_fields)
        worker_meta_api_to_chunk_delays[meta_api][c] = call
```
In my job there are 3639 chunk results and 200 workers and it costs 1414s to create worker meta apis. After caching the created meta apis, it only takes 156s by accelerating 9x speed.

## Related issue number

Fixes #3143 

## Check code requirements

### DIFF
--- a/mars/services/task/execution/mars/stage.py
+++ b/mars/services/task/execution/mars/stage.py
@@ -33,6 +33,16 @@ from ..api import ExecutionChunkResult
 
 logger = logging.getLogger(__name__)
 
+_address_to_meta_api = defaultdict(dict)
+
+
+async def _get_meta_api(session_id: str, address: str):
+    meta_api = _address_to_meta_api[session_id].get(address)
+    if meta_api is None:
+        meta_api = await WorkerMetaAPI.create(session_id, address)
+        _address_to_meta_api[session_id][address] = meta_api
+    return meta_api
+
 
 class TaskStageProcessor:
     def __init__(
@@ -275,7 +285,7 @@ class TaskStageProcessor:
         worker_meta_api_to_chunk_delays = defaultdict(dict)
         for c in update_meta_chunks:
             address = chunk_to_result[c].meta["bands"][0][0]
-            meta_api = await WorkerMetaAPI.create(session_id, address)
+            meta_api = await _get_meta_api(session_id, address)
             call = meta_api.get_chunk_meta.delay(
                 c.key, fields=list(get_chunk_params(c).keys())
             )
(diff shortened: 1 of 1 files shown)


wjsi:
We've already annotated `WorkerMetaAPI.create` with `alru_cache`. Thus it is better to inspect why this cache does not work.

zhongchun:
> We've already annotated `WorkerMetaAPI.create` with `alru_cache`. Thus it is better to inspect why this cache does not work.

OK. I'll check it.

zhongchun:
> We've already annotated `WorkerMetaAPI.create` with `alru_cache`. Thus it is better to inspect why this cache does not work.

The reason is the default `maxsize` of `alru_cache` is 128. There will be lots of duplicate creations if number of workers is greater than 128. So in my opinion we should increase the default value of `alru_cache` or just give a larger value for `WorkerMetaAPI.create`.

PR #3146, opened 2022-06-15
Increase the default value of alru cache max size

### LINKED ISSUE #3143: [BUG] There maybe duplicate `WorkerMetaAPI` creation in `TaskStageProcessor`

zhongchun:
## What do these changes do?
The default value of alru cache is 128 which is small. Like `WorkerMetaAPI.create`, there will be lots of duplicate creations if number of workers is greater than 128. 

In my job there are 3639 chunk results and 200 workers and it costs 1414s to create worker meta apis. After caching the created meta apis, it only takes 156s by accelerating 9x speed.

We should increase the default value of alru_cache like 1024.

## Related issue number

Fixes #3143 

## Check code requirements

### DIFF
--- a/mars/lib/aio/lru.py
+++ b/mars/lib/aio/lru.py
@@ -152,7 +152,7 @@ def _cache_miss(wrapped, key):
 
 def alru_cache(
     fn=None,
-    maxsize=128,
+    maxsize=1024,
     typed=False,
     *,
     cache_exceptions=True,
--- a/mars/lib/aio/lru.py
+++ b/mars/lib/aio/lru.py
@@ -152,7 +152,7 @@ def _cache_miss(wrapped, key):
 
 def alru_cache(
     fn=None,
-    maxsize=1024,
+    maxsize=128,
     typed=False,
     *,
     cache_exceptions=True,
--- a/mars/services/meta/api/oscar.py
+++ b/mars/services/meta/api/oscar.py
@@ -237,7 +237,7 @@ class BaseMetaAPI(AbstractMetaAPI):
 
 class MetaAPI(BaseMetaAPI):
     @classmethod
-    @alru_cache(cache_exceptions=False)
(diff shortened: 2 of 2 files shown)


mars/lib/aio/lru.py
```diff
@@ -152,7 +152,7 @@ def _cache_miss(wrapped, key):
 
 def alru_cache(
     fn=None,
-    maxsize=128,
+    maxsize=1024,
```
fyrestone, line 155: Why not set the maxsize for `WorkerMetaAPI`? The default value 128 is the same with `functools.lru_cache`: https://docs.python.org/3/library/functools.html#functools.lru_cache

zhongchun: Agree with you. We can only set cache max size for `WorkerMetaAPI` if 128 is a common value.

wjsi (approved):
LGTM

hekaisheng (commented):
LGTM

hekaisheng (approved):
LGTM

PR #3157, opened 2022-06-20
Enable ```TensorTupleOp``` to support functions with scalar inputs

RandomY-2:
## What do these changes do?

Functions such as ```special.fresnel_zeros``` take in scalar inputs, but currently, TensorTupleOp will convert the input to be a tensor and feed to the original scipy special function. This change allows scalar inputs to be stored and used when needed. 

## Check code requirements

### DIFF
--- a/mars/tensor/special/core.py
+++ b/mars/tensor/special/core.py
@@ -17,7 +17,7 @@ import scipy.special as spspecial
 from ...core import ExecutableTuple
 from ... import opcodes
 from ..datasource import tensor as astensor
-from ..arithmetic.core import TensorUnaryOp, TensorBinOp, TensorMultiOp
+from ..arithmetic.core import TensorElementWise, TensorUnaryOp, TensorBinOp, TensorMultiOp
 from ..array_utils import (
     np,
     cp,
@@ -166,7 +166,7 @@ class TensorTupleOp(TensorSpecialUnaryOp):
         in_tensor = op.input
 
         if in_tensor.ndim != 0:
-            return (yield from super().tile(op))
+            return (yield from TensorElementWise.tile(op))
         else:
             in_chunk = in_tensor.chunks[0]
             chunk_op = op.copy().reset_key()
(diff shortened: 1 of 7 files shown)


wjsi:
Please fix failed tests in `CI Tensor` in azure pipelines.

mars/tensor/special/core.py
```diff
@@ -156,6 +166,60 @@ def __call__(self, x, out=None):
             out_tensor.data = res_tensor.data
         return out
 
+    @classmethod
+    def tile(cls, op):
+        in_tensor = op.input
+
+        if in_tensor.ndim != 0:
+            return (yield from TensorElementWise.tile(op))
```
wjsi, line 174:
```suggestion
            return (yield from super().tile(op))
```

wjsi (commented):
Generally LGTM, I left two comments.

wjsi:
Wait...As `fresnel_zeros` only accepts scalar inputs, it is not a typical `scipy.special` function. Thus it should be treated as something like `mt.ones` which accept scalar dimensions and outputs tensors with multiple chunks other than simple arithmetic operands. Maybe rewriting the whole thing is needed...

RandomY-2:
> Wait...As `fresnel_zeros` only accepts scalar inputs, it is not a typical `scipy.special` function. Thus it should be treated as something like `mt.ones` which accept scalar dimensions and outputs tensors with multiple chunks other than simple arithmetic operands. Maybe rewriting the whole thing is needed...

Do you mean we should create a separate tensor just for `fresnel_zeros` like how we did for `mt.ones`? Instead of putting this logic in `TensorTupleOp` as a special case.

wjsi:
> > Wait...As `fresnel_zeros` only accepts scalar inputs, it is not a typical `scipy.special` function. Thus it should be treated as something like `mt.ones` which accept scalar dimensions and outputs tensors with multiple chunks other than simple arithmetic operands. Maybe rewriting the whole thing is needed...
> 
> Do you mean we should create a separate tensor just for `fresnel_zeros` like how we did for `mt.ones`? Instead of putting this logic in `TensorTupleOp` as a special case.

Right. The operand should produce multiple chunks given `options.chunk_store_limit` or some `chunk_size` value provided by the caller.

RandomY-2:
cannot really figure out how to create the chunks. will close this pr first.

PR #3165, opened 2022-06-23
[ray] Support scheduling ray tasks in Ray oscar deploy backend

### LINKED ISSUE #3164: [Ray] Support supervisor on oscar for ray task mode
Currently ray dag mode run supervisor in client, which have some issues:
- Client and Ray cluster are usually not in the same cluster or data center.  There may be instability and delay in communication, which will lead to serious latency in Ray Task scheduling, resulting in insufficient pipelineization and low task throughput and resource utilization.
-  The Mars Dashboard address cannot be accessed through a browser. Since Supervisor is created in Notebook, Mars Dashboard cannot be accessed through a proxy in ray cluster.
- The small size of Notebook container may not meet the resource requirements for supervisor. Generally, the Notebook container specification is 2C4G or 4C8G. When the data scale is large, Supervisor may OOM.
- Can not support the large-scale Failover which needs distributed Supervisor. Large-scale Failover need to save the lineage of a large number of subtasks, so it may need to make the Supervisor running in multiple Ray Actors to ensure that the Supervisor can store a large number of fine-grained lineages. If the Supervisor is running in the client, it is not an independent instance, and it is difficult to extend and make it distributed.
- Ray driver resource usage are not managed by ray cluster, which also increase the possibilities of OOM

We should support scheduling ray tasks in ray actors.

chaokunyang:
## What do these changes do?
This PR add support for scheduling ray tasks in Ray oscar deploy backend

## Related issue number

Closes #3164 

## Check code requirements

### DIFF
--- a/mars/deploy/oscar/ray.py
+++ b/mars/deploy/oscar/ray.py
@@ -451,7 +451,7 @@ class RayCluster:
                 self._worker_cpu,
                 self._worker_mem,
             )
-        else:
+        elif self.backend == "ray":
             execution_config.merge_from(
                 ExecutionConfig.from_params(
                     backend=self.backend,
@@ -464,6 +464,8 @@ class RayCluster:
             await self.start_oscar(
                 self._n_supervisor_process, self._supervisor_mem, 0, 0, 0
             )
+        else:
+            raise ValueError(f"Unsupported backend type: {self.backend}.")
 
     async def start_oscar(
         self, n_supervisor_process, supervisor_mem, worker_num, worker_cpu, worker_mem
--- a/mars/deploy/oscar/ray.py
+++ b/mars/deploy/oscar/ray.py
@@ -37,6 +37,7 @@ from ...services.cluster.backends.base import (
     AbstractClusterBackend,
 )
 from ...services import NodeRole
+from ...services.task.execution.api import ExecutionConfig
 from ...utils import lazy_import, retry_callable
 from ..utils import (
     load_config,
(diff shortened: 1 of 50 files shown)


fyrestone:
Why not create an actor of Mars `LocalCluster` of `mars/deploy/oscar/local.py`? This is much simpler and most of the logic is shared with Mars which is well tested in CI.

chaokunyang:
> Why not create an actor of Mars `LocalCluster` of `mars/deploy/oscar/local.py`? This is much simpler and most of the logic is shared with Mars which is well tested in CI.

Creating supervisor in Ray oscar can reuse all existing integration with ray. If we create an actor of Mars `LocalCluster` of `mars/deploy/oscar/local.py`, the serialization and communication between mars client and supervisor will be an issue:
- We need to implement a proxy for LocalCluster based on ray call
- ObjectRef can't be passed to supervisor from mars client directly

fyrestone (dismissed):
LGTM

mars/deploy/oscar/ray.py
```diff
@@ -434,6 +439,35 @@ async def start(self):
...
+        if self.backend == "mars":
+            await self.start_oscar(
+                self._n_supervisor_process,
+                self._supervisor_mem,
+                self._worker_num,
+                self._worker_cpu,
+                self._worker_mem,
+            )
+        else:
+            execution_config.merge_from(
```
zhongchun, line 455: Maybe it's better to specify the backend explicitly like `ray` or something others instead of `else` statement.
And we can raise an exception if user passes an unsupported backend.

chaokunyang: Good catch, we should validate the passed backend type, just fixed it.

zhongchun (approved):
LGTM.

fyrestone (approved):
LGTM

PR #3177, opened 2022-07-01
Improve numexpr fusion

fyrestone:
### Fuse a sub graph instead of a line
graph(@: node, S: Slice Chunk, #: fused_node): fuse stopped at S, because numexpr don't support Slice op

- before

```python
@                   @              @             @
  \               /                  \         /
    @ --> @ --> S      ========>       # --> S
  /               \                  /         \
@                   @              @             @
```

- after

```python
@                   @                          @
  \               /                          /
    @ --> @ --> S      ========>     # --> S
  /               \                          \
@                   @                          @
```

### New support operands

Update Mars fuse logic to support new operands:
- TensorFloor
- TensorCeil
- TensorAnd
- TensorOr
- TensorNot

### Better cache hit

- before

`((V_7db74de3f328b72d4e13457955bf5289_0+V_5b589bdfffe813d3d0414d0a7fe76c1e_0)+10)`

- after

`((V_0+V_1)+10)`

## Performance

The fused expr has the same performance boost according to the numexpr benchmark. My test environment,
- Quad-Core Intel Core i7 @ 2.2GHz
- Python 3.8.7
- numexpr 2.8.1 without MKL support
- numpy 1.22.4

The test code
```python
import time
import mars
import mars.tensor as mt

mars.new_session()


start_time = time.time()
for _ in range(100):
    a = mt.arange(1e6)
    b = mt.arange(1e6) * 0.1
    c = mt.sin(a) + mt.arcsinh(a / b)
    c.execute()
    c = a * b - 4.1 * a > 2.5 * b
    c.execute()
print(time.time() - start_time)
```
 Result
- The average cost of subtask `mt.sin(a) + mt.arcsinh(a / b)` is reduced from 0.07 to 0.02 (faster ~3x)
- The average cost of subtask `a * b - 4.1 * a > 2.5 * b` is reduced from 0.019 to 0.004 (faster ~5x)
- The cost of 100 loops of the test code  is reduced from 16.9 to 11 (faster ~30%)

For the complex expression, the performance of single subtask execution (including fusing cost) is faster about **3x~5x**, the end to end is faster about **25%~35%**.

In the future, we can try to move the numexpr fusion from physical to logical to get better performance.

## Related issue number

Fixes #xxxx

## Check code requirements

### DIFF
--- a/mars/optimization/physical/numexpr.py
+++ b/mars/optimization/physical/numexpr.py
@@ -15,9 +15,9 @@
 import dataclasses
 import functools
 import logging
-import numpy as np
-
 from typing import List, Set
+
+import numpy as np
 
 from ...core import ChunkType, ChunkGraph
 from ...tensor import arithmetic
--- a/mars/optimization/physical/numexpr.py
+++ b/mars/optimization/physical/numexpr.py
@@ -17,7 +17,7 @@ import functools
 import logging
 import numpy as np
 
-from typing import List
+from typing import List, Set
 
 from ...core import ChunkType, ChunkGraph
 from ...tensor import arithmetic
@@ -116,6 +116,7 @@ def _can_fuse(node: ChunkType):
 def _collect_fuse(
     graph: ChunkGraph,
     node: ChunkType,
+    graph_results: Set[ChunkType],
     cached_can_fuse,
 ):
     fuse_graph = ChunkGraph()
@@ -143,7 +144,7 @@ def _collect_fuse(
         # Skip the successors of tail reduction node.
         if node is tail_reduction_node:
             continue
-        is_tail = graph.count_successors(node) == 0
+        is_tail = graph.count_successors(node) == 0 or node in graph_results
         for n in graph.iter_successors(node):
             can_fuse = cached_can_fuse(n)
             if can_fuse is False:
@@ -192,7 +193,7 @@ class NumexprRuntimeOptimizer(RuntimeOptimizer):
                 continue
             can_fuse = cached_can_fuse(node)
             if can_fuse is True:
-                fuse = _collect_fuse(graph, node, cached_can_fuse)
+                fuse = _collect_fuse(graph, node, graph_results, cached_can_fuse)
                 if len(fuse.graph) > 1:
                     explored.update(fuse.graph)
                     if len(fuse.tails) == 1:
--- a/mars/optimization/physical/numexpr.py
+++ b/mars/optimization/physical/numexpr.py
@@ -12,7 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from ...core import ChunkType
+import dataclasses
+import functools
+import logging
+import numpy as np
+
+from typing import List
+
+from ...core import ChunkType, ChunkGraph
 from ...tensor import arithmetic
 from ...tensor import reduction
 from ...tensor.fuse import TensorNeFuseChunk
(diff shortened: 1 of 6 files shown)


dlee992:
Just curious: I saw related code in `optimization/logical/tileable/arithmetic_query.py`, which in the logical graph stage. But ur PR is valid in the physical graph stage. Why is it different? I feel these two parts can both optimize `tensor/series/dataframe expression` into numexpr expression.

fyrestone:
> Just curious: I saw related code in `optimization/logical/tileable/arithmetic_query.py`, which in the logical graph stage. But ur PR is valid in the physical graph stage. Why is it different? I feel these two parts can both optimize `tensor/series/dataframe expression` into numexpr expression.

They are different,

- arithmetic_query optimize the logical graph, convert graph to expression, then use one `pandas.eval` to compute the graph.
- the numexpr optimize the physical graph, convert graph to expression, then use one `numexpr.evaluate` to compute the graph.

The numexpr fuse can be moved to the logical optimization just as the arithmetic_query. Maybe some logic can be shared between them.

Currently, the `pandas.eval` can not replace the `numexpr.evaluate`. In my test,

``` python
# Python 3.8.7 numexpr 2.8.1 pandas 1.4.2

import numpy as np
import pandas as pd
COUNT = 1000000
df = pd.DataFrame({
    'y': np.random.normal(0, 1, COUNT),
    'z': np.random.gamma(50, 1, COUNT),
})

# 18.7 ms ± 2.49 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
timeit df.y[(10 < df.z) & (df.z < 50)].mean()
# 8.23 ms ± 203 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
timeit df.y.values[(10 < df.z.values) & (df.z.values < 50)].mean()
# 35.4 ms ± 5.84 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
timeit df.eval('y[(10 < z) & (z < 50)].mean()', engine='numexpr')
```

Please refer to the issue: https://stackoverflow.com/questions/52588653/why-is-pandas-eval-with-numexpr-so-slow

dlee992:
Awesome, thanks! I kind of got it.

mars/optimization/physical/numexpr.py
```diff
@@ -100,35 +181,74 @@ def is_available(cls) -> bool:
     def optimize(self):
         fuses = []
         explored = set()
+        cached_can_fuse = functools.lru_cache(maxsize=None)(_can_fuse)
```
qinxuye, line 184: Why not just add annotation `lru_cache` to `_can_fuse` directly?

mars/tensor/fuse/tests/test_numexpr_execution.py
```diff
@@ -17,7 +17,9 @@
 import numpy as np
 
 from ....utils import ignore_warning
-from ...datasource import tensor
+from ...datasource import tensor, arange
+from ...reduction import sum as mt_sum
+from ...arithmetic import abs as mt_abs
```
qinxuye, line 22: Import arithmetic before datasource.

qinxuye (commented):
Nice work, there is an empty file `b2.py`, could you remove it?

mars/optimization/physical/numexpr.py
fyrestone: To avoid caching chunks in global and is slightly faster.

- `lru_cache` in optimize: unlimited cache size without lru.
- `lru_cache` to `_can_fuse`: fixed cache size with lru.

The chunks may be unused after the subtask execution, we don't need to cache it in global.

mars/tensor/fuse/tests/test_numexpr_execution.py
fyrestone: Thanks. I will fix it.

qinxuye (approved):
LGTM

hekaisheng (approved):
LGTM

PR #3205, opened 2022-08-04
Apply Operand Closure clean up

### LINKED ISSUE #3193: [Operand] Apply operand closure cleanup
For function calls such as DataFrame.apply, DataFrameGroupby.apply, Series.apply, etc., the user will pass in a custom function, i.e. UDF. Mars will serialize the custom function multiple times during graph building, scheduling and execution. If the custom function captures a large amount of data, such as pandas DataFrame internally, it will cause **bottlenecks in scheduling and execution**:
- In the tile phase, the udf will be serialized when the key is generated for each operator
- In the tile phase, the udf is serialized when the key is generated for each chunk
- In the stage of generating SubtaskGraph, the udf will be serialized when the logic key is generated for each chunk
- When submitting each Subtask to the Worker main pool, the udf will be serialized for each Subtask
- When receiving Subtask in Worker main pool, it will deserialize the UDF
- When the Worker main pool submits each Subtask to the Worker sub pool, the udf will be serialized for each Subtask
- When the Worker sub pool receives the Subtask, the UDF will be deserialized

Although the serialization of udf functions can be cached by custom Mars serialization (currently cached when calculating chunk key/logic key in our inner codebase), if there is a large amount of data in udf, the result of serialization of a single Subtask may have Hundreds of megabytes, repeatedly sending these data to the Worker main pool and Sub pool during Subtask scheduling will still cause **Supervisor bottlenecks and scheduling delays**.
On the other hand, if a single Subtask object is too large, **it will also lead to a large amount of bloodline storage overhead, resulting in unrecoverable bloodline loss after the task fails**.
Based on this, consider implementing a function closure cleaner at the Mars operator level, clean up the function closure in the tile phase, serialize the closure result into Storage, and then the operator only holds the reference of the closure stored in the storage. Before the operator is executed, the closure object is obtained from Storage, the udf is restored, and then the specific calculation is performed.
 
# Proposal
## Function closure cleanup
The function closure object is in the __closure__ attribute of the function. You can get the object first, then put it into the storage, and then get the specific closure object from the storage to restore it when executing. The pseudo code is as follows:
```python
@classmethod
def tile(cls, op):
    closure = getattr(func, "__closure__")
    closure_ref = storage_api.put(closure)
    self.closure_ref = closure_ref
    if closure is not None:
        closure.__closure__ = None
        
@classmethod
def execute(cls, ctx, op):
    if self.closure_ref is not None:
        closure =  storage_api.get(self.closure_ref)
        func.__closure__ = closure
     # 执行具体的计算
```

Note: If the function closure is relatively small, it can be directly inlined in the code.
## Callable cleanup
The UDF passed in by the user may not be a function, but a subclass that implements the __callable__ method. In this case, the captured object cannot be obtained through the __closure__ attribute. In this case special handling of __callable__ object clipping in the case of this class is required. If the object implements python's __reduce__ or __getstate__ method, further processing is required

vcfgv:
## What do these changes do?

1. Compute logic key and for specific operands before tiling.
2. While tiling, put entire UDF to storage when sum of sizes of cell_contents exceed a certain threshold, which is customizable for users to adjust or disable this feature.
3. While executing, get UDF from storage and restore to op.func.
4. Add storage service to supervisor to avoid store func on workers who may risk being released.

Under ray task mode, func key, i.e. ray.ObjectRef, shall be serialized by Ray rather than Mars. As a result, relevant tests are temporarily disabled for Github checks but passed locally with some modifications.

Lifecycle management of UDF storage object is planned in another PR.

## Related issue number

#3193 

## Check code requirements


mars/services/context.py
```diff
@@ -193,6 +193,20 @@ async def _fetch_chunks(self, data_keys: List[str]):
...
         await storage_api.fetch.batch(*fetches)
 
+    def supervisor_storage_put(self, key: str, object=None):
+        return self._call(self._supervisor_storage_put(key, object))
+
+    async def _supervisor_storage_put(self, key: str, object=None):
+        storage_api = await StorageAPI.create(self.session_id, self.supervisor_address)
+        return await storage_api.put(key, object)
+
+    def supervisor_storage_get(self, key: str):
```
fyrestone, line 203: The `supervisor_storage_put` and `supervisor_storage_get` is a bit confusing. In this PR, the closure is put in supervisor but get from worker.

mars/dataframe/base/apply.py
```diff
@@ -280,11 +310,53 @@ def _tile_series(cls, op):
...
     @classmethod
     def tile(cls, op):
+        cls._clean_up_closure(op)
         if op.inputs[0].ndim == 2:
             return (yield from cls._tile_df(op))
         else:
             return cls._tile_series(op)
 
+    @classmethod
+    def _clean_up_closure(cls, op):
```
fyrestone, line 320: It's better to put the entire UDF.

mars/services/context.py
fyrestone: The service context is already messy, but the simplest way to implement this feature is to extends the context. We can wrap the storage API like this, then in the `RayExecutionContext`, they can be implemented by `ray.put` and `ray.get`.

Ray object can be evicted if no ref count exists in the cluster. But, for Mars, the UDF object should be managed by lifecycle.

```python
def storage_put(self, object=None):
    """Put the object to the storage and returns a Key."""

def storage_get(self, key) -> object:
    """Get the object from the storage"""
```

fyrestone:
Some suggestions are here: https://github.com/mars-project/mars/issues/3193#issuecomment-1204708564

mars/dataframe/base/apply.py
```diff
@@ -126,6 +143,30 @@ def result_type(self):
...
+    def func_key(self):
+        return self._func_key
+
+    @func_key.setter
+    def func_key(self, func_key):
+        self._func_key = func_key
+
+    @property
+    def func_clean_up(self):
+        return self._func_clean_up
```
chaokunyang, line 164: Seems `self._func_clean_up` not used, why not invoke `func_clean_up` directly?

mars/dataframe/utils.py
```diff
@@ -1424,3 +1431,44 @@ def _concat_chunks(merge_chunks: List[ChunkType], output_index: int):
...
+
+def clean_up_func(op):
+    closure_clean_up_bytes_threshold = int(
+        os.getenv("MARS_CLOSURE_CLEAN_UP_BYTES_THRESHOLD", 10**4)
+    )
+    if closure_clean_up_bytes_threshold == -1:  # pragma: no cover
+        return
+    # note: Vineyard internally uses `pickle` which fails to pickle
+    # cell objects and corresponding functions.
+    if vineyard is not None:
```
chaokunyang, line 1444: If `vineyard` is installed but other storage are used, this will skip cleanup too?

mars/dataframe/base/apply.py
```diff
@@ -126,6 +143,30 @@ def result_type(self):
...
+
+    @func_clean_up.setter
+    def func_clean_up(self, func_clean_up):
+        self._func_clean_up = func_clean_up
```
chaokunyang, line 168: a little confusing with `clean_up_func`, better add some docs

mars/dataframe/base/apply.py
```diff
@@ -126,6 +143,30 @@ def result_type(self):
...
+    @func_key.setter
+    def func_key(self, func_key):
+        self._func_key = func_key
+
+    @property
+    def func_clean_up(self):
+        return self._func_clean_up
+
+    @func_clean_up.setter
+    def func_clean_up(self, func_clean_up):
```
chaokunyang, line 167:
```suggestion
    def func_clean_up(self, func_clean_up: bool):
```

chaokunyang:
Can we add some tests to check that closure cleanup are actually executed?

mars/dataframe/tests/test_utils.py
```diff
@@ -644,3 +651,65 @@ def get_chunks_meta(self, data_keys: List[str], **_) -> List[Dict]:
     assert isinstance(s2.chunks[1].op, DataFrameConcat)
     assert s2.chunks[1].name == "a"
     assert len(s2.chunks[1].op.inputs) == 2
+
+
+async def set_up_context(session_id, address):
```
chaokunyang, line 656: Make this as a fixture and cleanup when test finished

mars/dataframe/tests/test_utils.py
```diff
@@ -644,3 +651,65 @@ def get_chunks_meta(self, data_keys: List[str], **_) -> List[Dict]:
...
+    context = ThreadedServiceContext(session_id, address, address, address, loop=loop)
+    await context.init()
+    set_context(context)
+
+
+def test_clean_up_and_restore_func(setup):
+    asyncio.run(clean_up_and_restore_func(setup=setup))
+
+
+async def clean_up_and_restore_func(setup):
```
chaokunyang, line 667: Can ew use pytest.asyncio ?

chaokunyang (dismissed):
LGTM

mars/dataframe/utils.py
```diff
@@ -1424,3 +1431,55 @@ def _concat_chunks(merge_chunks: List[ChunkType], output_index: int):
...
+            counted_bytes += sys.getsizeof(cell.cell_contents)
+            if counted_bytes >= closure_clean_up_bytes_threshold:
+                op.need_clean_up_func = True
+                break
+    # note: op.func_key is set to not None only when op.need_clean_up_func is True.
+    if op.need_clean_up_func:
+        # note: when used in ray task mode, data key is ray.ObjectRef which is returned
+        # after func being put into storage.
+        if isinstance(ctx, RayExecutionContext):  # pragma: no cover
+            op.func_key = ctx.storage_put(op.func)
```
fyrestone, line 1469: Is this line covered by test? The call does not match the signature.

mars/dataframe/groupby/tests/test_groupby_execution.py
```diff
@@ -932,6 +932,43 @@ def apply_series(s, truncate=True):
...
+    # Series
+    series1 = pd.Series([3, 4, 5, 3, 5, 4, 1, 2, 3])
+    ms1 = md.Series(series1, chunk_size=3)
+
+    applied = ms1.groupby(lambda x: x % 3).apply(apply_closure_series)
+    pd.testing.assert_series_equal(
+        applied.execute().fetch().sort_index(),
+        series1.groupby(lambda x: x % 3).apply(apply_closure_series).sort_index(),
+    )
+
```
zhongchun, line 970: We should also check the clean up closure is working or not.

mars/dataframe/utils.py
```diff
@@ -1424,3 +1431,55 @@ def _concat_chunks(merge_chunks: List[ChunkType], output_index: int):
...
+            logger.warning(
+                "Func cleanup is currently disabled when vineyard is used as storage backend."
+            )
+            return
+
+    func = op.func
+    if hasattr(func, "__closure__") and func.__closure__ is not None:
+        counted_bytes = 0
+        for cell in func.__closure__:
+            # note: another applicable way of measurements is df.memory_usage(index=True, deep=False).sum()
```
zhongchun, line 1459: Maybe we should use `# Note: `

mars/dataframe/utils.py
```diff
@@ -1424,3 +1431,55 @@ def _concat_chunks(merge_chunks: List[ChunkType], output_index: int):
...
+            counted_bytes += sys.getsizeof(cell.cell_contents)
+            if counted_bytes >= closure_clean_up_bytes_threshold:
+                op.need_clean_up_func = True
+                break
+    # note: op.func_key is set to not None only when op.need_clean_up_func is True.
```
zhongchun, line 1464: Ditto.

mars/dataframe/utils.py
```diff
@@ -1424,3 +1431,55 @@ def _concat_chunks(merge_chunks: List[ChunkType], output_index: int):
...
+    if op.need_clean_up_func:
+        # note: when used in ray task mode, data key is ray.ObjectRef which is returned
```
zhongchun, line 1466: Ditto.

mars/dataframe/utils.py
```diff
@@ -1424,3 +1431,55 @@ def _concat_chunks(merge_chunks: List[ChunkType], output_index: int):
...
+        # after func being put into storage.
+        if isinstance(ctx, RayExecutionContext):  # pragma: no cover
+            op.func_key = ctx.storage_put(op.func)
+        # note: when used in normal mode, data key is logic key while the returned data_info
```
zhongchun, line 1470: Ditto.

mars/dataframe/utils.py
```diff
@@ -1424,3 +1431,55 @@ def _concat_chunks(merge_chunks: List[ChunkType], output_index: int):
...
+            ctx.storage_put(op.func_key, op.func)
+        else:  # pragma: no cover
+            raise Exception("unknown context type: %s", type(ctx))
+        op.func = None
+
+
+def restore_func(ctx: Context, op):
+    if op.need_clean_up_func and ctx is not None:
+        assert op.func_key is not None
+        assert op.func is None
```
zhongchun, line 1484: We'd better add some assertion message.

mars/services/task/execution/ray/context.py
```diff
@@ -155,6 +155,9 @@ def get_worker_addresses(self) -> List[str]:
         # Returns virtual worker addresses.
         return self._worker_addresses
 
+    def storage_put(self, key: str = None, obj=None):  # pragma: no cover
```
zhongchun, line 158: Maybe we could add some tests to cover this method and `storage_get` is the same.

mars/dataframe/base/tests/test_base_execution.py
```diff
@@ -382,6 +382,23 @@ def test_data_frame_apply_execute(setup):
...
+    x = pd.Series([i for i in range(10**4)])
+    y = pd.Series([i for i in range(10**4)])
+
+    def closure(z):
+        return pd.concat([x, y], ignore_index=True)
+
+    r = df.apply(closure, axis=1)
+    result = r.execute().fetch()
+    expected = df_raw.apply(closure, axis=1)
+    pd.testing.assert_frame_equal(result, expected)
```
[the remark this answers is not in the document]
vcfgv: Codes to check whether the func clean up is executed as expected are located in test_checked_session.py with the test named `test_clean_up_and_restore_func`.

mars/dataframe/utils.py
```diff
@@ -1424,3 +1431,65 @@ def _concat_chunks(merge_chunks: List[ChunkType], output_index: int):
...
+        # if isinstance(ctx, RayExecutionContext):
+        #     op.func_key = ctx.storage_put(None, op.func)
+        # Note: when used in normal mode, data key is logic key while the returned data_info
+        # is not used as key.
+        if isinstance(ctx, ThreadedServiceContext):
+            assert (
+                op.logic_key is not None
+            ), "Logic key wasn't calculated before cleaning up func."
+            op.func_key = op.logic_key
+            ctx.storage_put(op.func_key, op.func)
```
fyrestone, line 1481: Can you return a storage object key from the `storage_put`? Then, we don't need to check instances here.

zhongchun (approved):
LGTM.

fyrestone (approved):
LGTM

PR #3208, opened 2022-08-08
Refine apply key generation

chaokunyang:
## What do these changes do?
This PR add func key cache to avoid duplicate func serialization when generate apply key which will be costly when func capture some big objects 

## Related issue number

Fixes #xxxx

## Check code requirements

### DIFF
--- a/mars/utils.py
+++ b/mars/utils.py
@@ -1623,12 +1623,15 @@ _func_token_cache = weakref.WeakKeyDictionary()
 
 
 def get_func_token(func):
-    token = _func_token_cache.get(func)
-    if token is None:
-        fields = get_func_token_values(func)
-        token = tokenize(*fields)
-        _func_token_cache[func] = token
-    return token
+    try:
+        token = _func_token_cache.get(func)
+        if token is None:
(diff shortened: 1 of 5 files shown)


zhongchun (dismissed):
LGTM.

mars/utils.py
```diff
@@ -1617,6 +1619,21 @@ def __str__(self):
...
+        if token is None:
+            fields = get_func_token_values(func)
+            token = tokenize(*fields)
+            _func_token_cache[func] = token
+        return token
+    except TypeError:  # cannot create weak reference to func like 'numpy.ufunc'
+        return tokenize(*get_func_token_values(func))
+
+
 def get_func_token_values(func):
```
fyrestone, line 1637: The `get_func_token_values` is not used by other modules except the `mars/utils.py`, it's better to rename it to `_get_func_token_values` to make it private.

chaokunyang: Updated

fyrestone (approved):
LGTM

zhongchun (approved):
LGTM.

Issue #3192, opened 2022-07-15
[BUG] supervisor memory leak


fyrestone:
**Describe the bug**

![image](https://user-images.githubusercontent.com/6308809/179145659-00ac7756-e65f-4938-83a3-82e62dc21b14.png)

The Mars job executed hundreds of tasks, some task has a large graph:

![image](https://user-images.githubusercontent.com/6308809/179145897-23bb9e41-1a13-4eb5-ae25-e14ed5da3836.png)

Set environ var `PYTHONTRACEMALLOC=1`, then run the following code:

```python
import tracemalloc
import gc
import mars
import mars.tensor as mt
import mars.dataframe as md


def run_round(extra_config=None):
    df = md.DataFrame(
        mt.random.rand(1200, 70, chunk_size=70),
        columns=[f"col{i}" for i in range(70)])

    for i in range(70):
        df[f"col{i + 70}"] = df[f"col{i}"].fillna(0)
        df[f"col{i + 140}"] = df[f"col{i}"].fillna(0)
    for i in range(70):
        df[f"col{i}"] = df[f"col{i}"] / 100
    df = df.fillna(0)
    df.map_chunk(lambda x: x).execute(extra_config=extra_config)


def main():
    mars.new_session()
    count = 0
    run_round(extra_config={"enable_profiling": True})
    gc.collect()
    s1 = tracemalloc.take_snapshot()

    while True:
        print(f"==============>run {count}")
        count += 1
        run_round()
        if count == 3:
            break

    gc.collect()
    s2 = tracemalloc.take_snapshot()
    diff = s2.compare_to(s1, "lineno")
    print("[ Top 10 differences ]")
    for stat in diff[:10]:
        print(stat)


if __name__ == "__main__":
    main()
```

output

``` python
[ Top 10 differences ]
/home/admin/mars/mars/core/base.py:96: size=10.5 MiB (+8098 KiB), count=59034 (+44276), average=187 B
/home/admin/mars/mars/core/operand/core.py:100: size=10200 KiB (+7647 KiB), count=61221 (+45900), average=171 B
/home/admin/mars/mars/core/graph/builder/base.py:72: size=9861 KiB (+7400 KiB), count=85245 (+63956), average=118 B
/home/admin/mars/mars/services/task/analyzer/analyzer.py:250: size=9629 KiB (+7222 KiB), count=91908 (+68931), average=107 B
[...]

PR #3230, opened 2022-08-19
Lifecycle gc task service

fyrestone:
## What do these changes do?

All the TaskProcessorActors, TaskProcessors and related contexts, graphs are not GC.  This PR is to GC the task services by lifecycle.

- [x] Destroy the `TaskProcessorActor` only if all the result tileables have no refs and the task is finish.
  - New config `task.default_config.reserved_finish_tasks`, default value is 10.
- [x] Remove task name feature to simplify the task manager. 

## Related issue number

https://github.com/mars-project/mars/issues/3192
https://github.com/mars-project/mars/issues/3236

## Check code requirements

### DIFF
--- a/mars/services/lifecycle/supervisor/tracker.py
+++ b/mars/services/lifecycle/supervisor/tracker.py
@@ -228,6 +228,17 @@ class LifecycleTrackerActor(mo.Actor):
         )
         return decref_chunk_keys
 
+    def _get_remove_tileable_keys(self, tileable_keys: List[str]) -> List[str]:
+        to_remove_tileable_keys = []
+        for tileable_key in tileable_keys:
+            if self._tileable_ref_counts[tileable_key] <= 0:
+                if all(
+                    self._chunk_ref_counts[chunk_key] <= 0
+                    for chunk_key in self._tileable_key_to_chunk_keys[tileable_key]
+                ):
+                    to_remove_tileable_keys.append(tileable_key)
+        return to_remove_tileable_keys
+
     async def decref_tileables(
         self, tileable_keys: List[str], counts: List[int] = None
     ):
@@ -241,7 +252,7 @@ class LifecycleTrackerActor(mo.Actor):
             counts=list(decref_chunk_key_to_counts.values()),
         )
         to_remove_tileable_keys = await asyncio.to_thread(
-            list, (key for key in tileable_keys if self._tileable_ref_counts[key] <= 0)
+            self._get_remove_tileable_keys, tileable_keys
         )
         coros = []
         if to_remove_chunk_keys:
--- a/mars/services/lifecycle/supervisor/tracker.py
+++ b/mars/services/lifecycle/supervisor/tracker.py
@@ -20,6 +20,7 @@ from collections import defaultdict
 from typing import Dict, List, Optional
 
 from .... import oscar as mo
+from ....lib.aio import alru_cache
 from ...meta.api import MetaAPI
 from ...storage.api import StorageAPI
 from ..errors import TileableNotTracked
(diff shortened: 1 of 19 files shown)


mars/utils.py
```diff
@@ -91,6 +91,7 @@
 reset_id_random_seed = reset_id_random_seed
 new_random_id = new_random_id
 _create_task = asyncio.create_task
+_is_ci = (os.environ.get("CI") or "0").lower() in ("1", "true")
```
zhongchun, line 94: There is `_is_ci` in `mars/serialization/serializables/core.py` and `mars/serialization/serializables/field.py`. Maybe we can use this to instead them.

fyrestone: Thanks. I will replace the `_is_ci` in  serializables by this one.

zhongchun (approved):
LGTM.

mars/utils.py
```diff
@@ -1766,3 +1767,20 @@ def retry_call(*args, **kwargs):
...
+        try:
+            fut.result()
+        except asyncio.CancelledError:  # pragma: no cover
+            pass
+        except Exception:  # pragma: no cover
+            logger.exception(exception_formatter, *args)
+            if _is_ci:
+                sys.exit(-1)
+
+    aiotask.add_done_callback(_on_mars_aiotask_done)
```
chaokunyang, line 1786: Can we use `mars.oscar.debug.create_task_with_error_log` for print exception?

mars/services/lifecycle/supervisor/tracker.py
```diff
@@ -235,8 +240,18 @@ async def decref_tileables(
             list(decref_chunk_key_to_counts),
             counts=list(decref_chunk_key_to_counts.values()),
         )
-        # make _remove_chunks release actor lock
-        yield self._remove_chunks(to_remove_chunk_keys)
+        to_remove_tileable_keys = await asyncio.to_thread(
```
chaokunyang, line 243: Will there be many items in `tileable_keys`? If not, `asyncio.to_thread` is unnecessary

mars/services/task/supervisor/manager.py
```diff
@@ -174,12 +161,61 @@ async def submit_tileable_graph(
...
+        def _remove_task():
+            logger.debug("Remove task %s.", task_id)
+            self._task_id_to_processor_ref.pop(task_id, None)
+            if loop.is_running():
+                aiotask = loop.create_task(_destroy_actor())
+                add_aiotask_done_check_callback(
+                    aiotask, "Destroy processor ref %s failed.", processor_ref
+                )
+
+        weakref.finalize(ref_holder, _remove_task)
```
chaokunyang, line 217: Can we just check whether `self._reserved_finish_tasks.append(ref_holder)` pop item, and execute `_remove_task` body directly?

fyrestone: The `_remove_task` and `_destroy_actor` are always needed to be called. But, poping queue manually should handle two cases,
1, The queue size > 0, pop first, call `_remove_task`, then push last.
2, The queue size is 0, call `_remove_task` directly, not to push.

Use weakref is a bit simpler than pop queue manually.

mars/services/lifecycle/supervisor/tracker.py
fyrestone: Tileable keys are less than chunk keys, but I am not sure how many keys are here. I think `asyncio.to_thread` may be slower than iterate in asyncio loop directly for small data sets, but it can avoid blocking loop issues which is hard to findout.

mars/utils.py
fyrestone: I didn't notice the `mars.oscar.debug.create_task_with_error_log`. Yes, I can use it. Thanks.

zhongchun (approved):
LGTM.

chaokunyang (approved):
LGTM

PR #3233, opened 2022-08-22
[Metrics] Fix metrics and docs

zhongchun:
## What do these changes do?
In the previous design, metrics are allowed to be used without initialization. In this case, the default metrics, namely console metrics, are used, but this may lead to wrong use. For example, using metrics before initializing metrics, then using the default console metrics, but we may think that we have initialized metrics with prometheus or ray, then it should be the corresponding metrics. So this pr:

* Lazy initialization if not call `init_metrics`
* Fix ray storage metrics
* Fix mars usage demo

## Related issue number

Fixes #xxxx

## Check code requirements


mars/metrics/api.py
```diff
@@ -61,14 +61,24 @@ def init_metrics(backend="console", config: Dict[str, Any] = None):
...
 def shutdown_metrics():
     global _metric_backend
     _metric_backend = "console"
     global _init
     _init = False
+    logger.info("Shutdown metrics of backend: %s.", _metric_backend)
+
+
+def _check_metrics_valid():
+    if _init is False:
```
chaokunyang, line 76:
```suggestion
    if not _init:
```

mars/metrics/api.py
```diff
@@ -98,18 +108,46 @@ class Metrics:
 
     @staticmethod
     def counter(name, description: str = "", tag_keys: Optional[Tuple[str]] = None):
+        _check_metrics_valid()
+        logger.info(
+            "Initializing a counter whose name: %s, tag keys: %s, backend: %s",
```
chaokunyang, line 113:
```suggestion
            "Initializing a counter with name: %s, tag keys: %s, backend: %s",
```

zhongchun: Thanks.

mars/metrics/api.py
zhongchun: Thanks.

chaokunyang (approved):
LGTM

mars/core/operand/core.py
```diff
@@ -490,7 +488,18 @@ def execute(results: Dict[str, Any], op: OperandType):
             try:
                 result = executor(results, op)
                 succeeded = True
-                op_executed_number.record(1, {"op": op.__class__.__name__})
+                global op_executed_number
```
chaokunyang, line 491: This seems not easy to use compared to previsous API

chaokunyang: Can we update the metrics Counter/Guage instance when metris are init?

zhongchun: > Can we update the metrics Counter/Guage instance when metris are init?

Yes, good idea. I'll fix it.

chaokunyang (approved):
LGTM

fyrestone (approved):
LGTM

PR #3242, opened 2022-09-02
TypeDispatcher support one type multiple serializers

fyrestone:
## What do these changes do?

Mars serializers are not compatible with Ray serializers because Ray needs to dispatch `ActorHandle` and `ObjectRef` to Ray's serializer for reference counting. But, some `ActorHandle` and `ObjectRef` are nested in objects, then we needs to register object type to RaySerializer which is not compatible with Mars object type serializer.

This PR allows Mars to register multile serializers to one type, then Ray serializers are compatible with Mars serializers. We don't needs to `register_ray_serializers` and `unregister_ray_serializers`.

## Related issue number

Fixes #xxxx

## Check code requirements

### DIFF
--- a/mars/_utils.pxd
+++ b/mars/_utils.pxd
@@ -23,7 +23,6 @@ cdef class TypeDispatcher:
     cpdef void unregister(self, object type_)
     cdef _reload_lazy_handlers(self)
     cpdef get_handler(self, object type_)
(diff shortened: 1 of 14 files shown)


qinxuye (approved):
LGTM

chaokunyang (approved):
LGTM

PR #3246, opened 2022-09-08
Fix optimization rule memory leak

fyrestone:
## What do these changes do?

OptimizationRule has a cache that referenced optimized graphs. This PR removes the global cache.

## Related issue number

Fixes https://github.com/mars-project/mars/issues/3192

## Check code requirements

### DIFF
--- a/mars/optimization/logical/core.py
+++ b/mars/optimization/logical/core.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+import functools
 import weakref
 from abc import ABC, abstractmethod
 from collections import defaultdict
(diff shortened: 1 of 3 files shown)


mars/optimization/logical/core.py
```diff
@@ -106,17 +102,11 @@ def __init__(
...
-    ):
-        if (cls, graph, records) in cls._instances:
-            return cls._instances[cls, graph, records]
-        inst = cls._instances[cls, graph, records] = object.__new__(cls)
-        return inst
+        self._cached_rule = functools.lru_cache(maxsize=None)(
+            lambda _rule_type: _rule_type(
+                self._graph, self._records, self._optimizer_cls
+            )
+        )
```
zhongchun, line 109: It may cause memory keep growing if set `maxsize=None`, maybe we should set a fixed value.

fyrestone: This is a local cache, it will be deleted with rule object.

zhongchun (approved):
LGTM.

hekaisheng (approved):
LGTM

PR #3294, opened 2022-11-17
[Ray] Implement Ray executor subtask GC

fyrestone:
## What do these changes do?

Ray executor does not gc the input data and the intermediate data of subtask, then the memory usage of worker may be very high. This PR gc the chunk data after every operand finish.

- The input data are referenced by Ray core worker in _raylet.pyx, we can't gc them. 

## Related issue number

Fixes #xxxx

## Check code requirements


zhongchun (approved):
LGTM.

PR #3295, opened 2022-11-21
[Ray] Add metrics for Ray executor

fyrestone:
## What do these changes do?

- Add metrics for Ray executor:
  - submitted_subtask_number
  - started_subtask_number
  - completed_subtask_number
- Log ray task id
- Fallback to dump dot file if dot executable is not found
- Pin setuptools<64 before any installations

## Related issue number

Fixes #xxxx

## Check code requirements

### DIFF
--- a/mars/services/task/execution/ray/executor.py
+++ b/mars/services/task/execution/ray/executor.py
@@ -34,7 +34,7 @@ from .....core.operand import (
 from .....core.operand.fetch import FetchShuffle
 from .....lib.aio import alru_cache
 from .....lib.ordered_set import OrderedSet
-from .....metrics.api import init_metrics
+from .....metrics.api import init_metrics, Metrics
 from .....resource import Resource
 from .....serialization import serialize, deserialize
 from .....typing import BandType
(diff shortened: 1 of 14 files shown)


mars/services/task/execution/ray/executor.py
```diff
@@ -67,6 +67,24 @@
...
+)
+started_subtask_number = Metrics.counter(
+    "mars.band.started_subtask_number",
+    "The number of started subtask.",
+    ("subtask_id",),
+)
+completed_subtask_number = Metrics.counter(
+    "mars.band.completed_subtask_number",
+    "The number of completed subtask.",
+    ("subtask_id",),
```
zhongchun, line 84: Maybe we should add tags `task_id`, `stage_id` for `started_subtask_number` and `completed_subtask_number` so that we can easily calculate a stage, a task total number of these two metrics.
If not, we can also calculate these number by specifying the time range.

fyrestone: > Maybe we should add tags `task_id`, `stage_id` for `started_subtask_number` and `completed_subtask_number` so that we can easily calculate a stage, a task total number of these two metrics. If not, we can also calculate these number by specifying the time range.

Currently, the `execute_subtask` does not contains these info. I have to pass the `task_id` and `stage_id` to every Ray task for the tags.

zhongchun (approved):
LGTM.

zhongchun (dismissed):
LGTM.

chaokunyang (approved):
LGTM

zhongchun (approved):
LGTM.

Issue #3299, opened 2022-12-07
[BUG] Duplicate compute when rechunk after sort_values


fyrestone:
**Describe the bug**

```python
import numpy as np
import pandas as pd
import mars.dataframe as md

ns = np.random.RandomState(0)
df = pd.DataFrame(ns.rand(100, 10), columns=["a" + str(i) for i in range(10)])

# test one chunk
mdf = md.DataFrame(df, chunk_size=10)
result = mdf.sort_values("a0").rechunk(chunk_size=10).execute().fetch()
```

The graph:
[mars-3JNIaz5JsOzqzWcCmo825Wbt.pdf](https://github.com/mars-project/mars/files/10172647/mars-3JNIaz5JsOzqzWcCmo825Wbt.pdf)
![mars-3JNIaz5JsOzqzWcCmo825Wbt](https://user-images.githubusercontent.com/6308809/206084860-67bf466e-af1f-497a-a293-f839544300d2.png)

**To Reproduce**
To help us reproducing this bug, please provide information below:
1. Your Python version
2. The version of Mars you use Latest master
3. Versions of crucial packages, such as numpy, scipy and pandas
4. Full stack of the error.
5. Minimized code to reproduce the error.

**Expected behavior**

**Additional context**

PR #3301, opened 2022-12-08
Fix duplicate execution

fyrestone:
## What do these changes do?

Some chunks are copied when tiling (e.g. copied chunks with new chunk index), these copied chunks can't be treated as executed, so they will be regenerated. This PR use chunk key instead of chunk object to check whether it is executed.

## Related issue number

Fixes https://github.com/mars-project/mars/issues/3299

## Check code requirements


zhongchun (approved):
LGTM.

PR #3305, opened 2022-12-19
[Ray] Basic slow subtask detection

fyrestone:
## What do these changes do?

- Detect slow subtasks for Ray executor.
    - The unready subtasks will be checked by the historical durations in the same logic key. So, if a slow task has an unique logic key, then it can't be detected. 
- Fix scikit-learn 1.2.0 compatibility issue.

## Related issue number

Fixes #xxxx

## Check code requirements

### DIFF
--- a/mars/services/subtask/core.py
+++ b/mars/services/subtask/core.py
@@ -52,7 +52,7 @@ class SubtaskStatus(Enum):
 
 
 class Subtask(Serializable):
-    __slots__ = ("_repr", "_pure_depend_keys")
+    __slots__ = ("_repr", "_pure_depend_keys", "_runtime")
 
     subtask_id: str = StringField("subtask_id")
     subtask_name: str = StringField("subtask_name")
@@ -65,7 +65,6 @@ class Subtask(Serializable):
     virtual: bool = BoolField("virtual")
     retryable: bool = BoolField("retryable")
     priority: Tuple[int, int] = TupleField("priority", FieldTypes.int32)
-    rerun_time: float = Float64Field("rerun_time")
     extra_config: dict = DictField("extra_config")
     stage_id: str = StringField("stage_id")
     # chunks that need meta updated
@@ -95,7 +94,6 @@ class Subtask(Serializable):
         priority: Tuple[int, int] = None,
         virtual: bool = False,
         retryable: bool = True,
-        rerun_time: int = 0,
         extra_config: dict = None,
         stage_id: str = None,
         update_meta_chunks: List[ChunkType] = None,
@@ -116,7 +114,6 @@ class Subtask(Serializable):
             priority=priority,
             virtual=virtual,
             retryable=retryable,
-            rerun_time=rerun_time,
             extra_config=extra_config,
             stage_id=stage_id,
             update_meta_chunks=update_meta_chunks,
@@ -129,11 +126,13 @@ class Subtask(Serializable):
         )
         self._pure_depend_keys = None
         self._repr = None
+        self._runtime = None
 
     def __on_deserialize__(self):
         super(Subtask, self).__on_deserialize__()
         self._pure_depend_keys = None
         self._repr = None
+        self._runtime = None
 
     @property
     def expect_band(self):
@@ -153,6 +152,14 @@ class Subtask(Serializable):
             )
         self._pure_depend_keys = pure_dep_keys
         return pure_dep_keys
+
+    @property
+    def runtime(self):
+        return self._runtime
+
+    @runtime.setter
+    def runtime(self, r):
+        self._runtime = r
 
     def __repr__(self):
         if self._repr is not None:
--- a/mars/services/subtask/core.py
+++ b/mars/services/subtask/core.py
@@ -65,7 +65,7 @@ class Subtask(Serializable):
     virtual: bool = BoolField("virtual")
     retryable: bool = BoolField("retryable")
     priority: Tuple[int, int] = TupleField("priority", FieldTypes.int32)
-    rerun_time: int = Int32Field("rerun_time")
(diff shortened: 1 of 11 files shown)


mars/services/task/execution/ray/executor.py
```diff
@@ -368,6 +371,74 @@ class _RayMonitorContext:
     )
 
 
+@dataclass
+class _RaySubtaskRuntime:
+    start_time: float = 0.0
+
+
+class _RaySlowSubtaskChecker:
```
chaokunyang, line 379: Is it possible to unify this with SepculativeScheduler?

mars/services/task/execution/ray/executor.py
```diff
@@ -368,6 +371,74 @@ class _RayMonitorContext:
...
+    def __init__(
+        self,
+        total_subtask_count: int,
+        submitted_subtasks: OrderedSet,
+        completed_subtasks: OrderedSet,
+    ):
+        self._total_subtask_count = total_subtask_count
+        self._submitted_subtasks = submitted_subtasks
+        self._completed_subtasks = completed_subtasks
+        self._logic_key_to_subtask_costs = collections.defaultdict(list)
```
chaokunyang, line 394: homogeneous subtasks relies on symmetric subtask graph, would be better to be consistent with existing slow task detection logic

fyrestone: This is similar with existing slow task detection logic, the subtask durations are grouped by the logic key.

mars/services/task/execution/ray/executor.py
fyrestone: It's hard to unify these two checker. The subtask durations are updated by the checker in Ray executor.

zhongchun:
LGTM. But i have a questions:
It seems we just detect slow subtasks and do not trigger speculative execution. If so, we may add a switch to control whether enable `_RaySlowSubtaskChecker` or not.

mars/services/subtask/core.py
```diff
@@ -65,7 +65,6 @@ class Subtask(Serializable):
     virtual: bool = BoolField("virtual")
     retryable: bool = BoolField("retryable")
     priority: Tuple[int, int] = TupleField("priority", FieldTypes.int32)
-    rerun_time: int = Int32Field("rerun_time")
```
zhongchun, line 68: Why remove the `rerun_time`?

fyrestone:
> LGTM. But i have a questions: It seems we just detect slow subtasks and do not trigger speculative execution. If so, we may add a switch to control whether enable `_RaySlowSubtaskChecker` or not.

This overhead is small, so we can always check the slow subtasks.

mars/services/task/execution/ray/executor.py
```diff
@@ -368,6 +371,74 @@ class _RayMonitorContext:
...
+        if logic_key not in self._logic_key_to_subtask_costs:
+            # The subtask logic key has no costs.
+            return False
+        logic_parallelism = subtask.logic_parallelism
+        if not logic_parallelism:
+            # Invalid parallelism.
+            return False
+        subtask_costs = self._logic_key_to_subtask_costs[logic_key]
+        complete_count = len(subtask_costs)
+        if complete_count / logic_parallelism < 0.75:
```
zhongchun, line 425: Could we turn the `0.75` to a configurable parameter?

chaokunyang (approved):
LGTM

mars/services/task/execution/ray/executor.py
```diff
@@ -368,6 +371,74 @@ class _RayMonitorContext:
...
+            # Too few complete subtasks.
+            return False
+        check_info = self._logic_key_to_check_info.get(logic_key)
+        if check_info is None or check_info.count != complete_count:
+            arr = np.array(subtask_costs)
+            q1, q3 = np.quantile(arr, 0.25), np.quantile(arr, 0.75)
+            upper = q3 + 3 * (q3 - q1)
+            self._logic_key_to_check_info[
+                logic_key
+            ] = _RaySlowSubtaskChecker._CheckInfo(complete_count, upper)
```
zhongchun, line 435: There is an algorithm to calculate the upper value which is called `duration_threshold` in `SpeculativeScheduler`. Is it possible to keep them the same or extract a common detection algorithm?

mars/services/subtask/core.py
fyrestone: It is unused.

mars/services/task/execution/ray/executor.py
fyrestone: It could be a config, but if covers most cases, fewer config is better. If it is a config, users need to know the DAG to change the config.

mars/services/task/execution/ray/executor.py
fyrestone: Good suggestion. `duration_threshold` is a better name.

fyrestone: Fixed.

zhongchun (approved):
LGTM.

chaokunyang (approved):
LGTM

mars/services/task/execution/ray/config.py
```diff
@@ -77,6 +78,19 @@ def get_log_interval_seconds(self):
             "log_interval_seconds", DEFAULT_LOG_INTERVAL_SECONDS
         )
 
+    def get_check_slow_subtasks_interval_seconds(self) -> float:
+        return self._ray_execution_config.get(
```
chaokunyang, line 82: why no just reuse existing sppeculative config?

fyrestone: This is not the speculative execution. If we implemented the speculative execution in Ray executor, then both of them can share the config.

PR #3342, opened 2023-04-21
Optimize SubtaskGraph generation

### LINKED ISSUE #3341: [GRAPH][ENHANCEMENT] Optimize `SubtakGraph` generation
## Problem
In `gen_subtask_graph`, Mars always create new out chunks even if the out chunk already exists. It costs a lot of time if there are plenty of chunks. Actually, there is no need to create out chunks unless the `FetchChunk`.
I did a comparison, in which one creates new out chunks and the other does not. The result shows that the cost is reduced from 43.4s to 14.6s with about 66.4% reduction.

So we should optimize the `SubtaskGraph` generation and make it more efficient.

zhongchun:
## What do these changes do?
In gen_subtask_graph, Mars always create new out chunks even if the out chunk already exists. It costs a lot of time if there are plenty of chunks. 

## Related issue number

Fixes #3341

I did a comparison, in which one creates new out chunks and the other does not. The test scripts are:
```python
size = 50000
da1 = mt.random.random((size, 2), chunk_size=(1, 2))
df1 = md.DataFrame(da1, columns=list("AB"))
df2 = df1 + 10
df3 = df2.sum()
ret = df3.execute()
```
Cost time are: 122.92s, 56.63s.

## Check code requirements


mars/services/task/analyzer/analyzer.py
```diff
@@ -287,14 +289,23 @@ def _gen_subtask_info(
...
+                    for c in copied_op.new_chunks(
+                        inp_chunks, kws=[c.params.copy() for c in chunk.op.outputs]
+                    )
+                ]
+            else:
+                out_chunks = chunk.op.outputs
+                # Note: `dtypes`, `index_value`, and `columns_value` are lazily
+                # initialized, so we should call property `params` to initialize
+                # these fields.
+                [o.params for o in out_chunks]
```
fyrestone, line 307: It's wired, what would happen without these codes?

zhongchun: There will no `columns_value`, `index_value` which are used in `MainPool`.

fyrestone: These field are lazily initialized, but field `a` or `b` are lazily initialized by `params`.  Can you make `a` initialized by `a`, `b` initialized by `b`? Then we can lazily initialize them in Worker Main Pool.

PR #3354, opened 2023-07-01
Fix test fail with numpy 1.25

wjsi:
## What do these changes do?

Fix test fail with numpy 1.25.

## Related issue number

NA


zhongchun (approved):
LGTM.

